PR #62, opened 2026-06-04
feat(auth): add structured logging to diagnose the OAuth browser flow

dlecan:
## Context

On Claude Cowork Windows the OAuth login fails — the browser never opens
(works on Mac). See #60.

Root cause for the lack of a diagnosis (confirmed in code, not the failure
itself):
- `open(authUrl).catch(() => {})` **swallowed** any browser-launch error, so we
  could never see *why* it failed on Windows.
- The fallback authorization URL only went to **stderr**, invisible in most MCP
  client UIs.
- `config.logLevel` / `--log-level` was parsed but **wired to nothing**.

This PR is **iteration 1 = observability only**. It does not change the auth
UX or the 5-min timeout (those are tracked for later iterations).

## What changed

New `src/utils/logger.ts` — a small structured logger with two sinks, designed
to be useful on **every** MCP client (not just Cowork):

| Sink | Mechanism | Reach |
|------|-----------|-------|
| stderr (floor) | gated by `LOG_LEVEL` | every mainstream client captures it to a log file |
| MCP logging | `notifications/message` via the newly advertised `logging` capability | clients that support it; ignored gracefully otherwise |

- **A1** — the browser-open failure is now logged as `oauth_browser_open_failed`
  with the error message, `errorCode`, `platform`, `os.release()`, WSL
  detection, and the **presence** (booleans, never values) of `SYSTEMROOT` /
  `WINDIR` / `ComSpec` / `PATH` / `DISPLAY`.
- **A2** — step traces at `debug` (callback listening, authorize URL, callback
  received, token-exchange HTTP status, timeout) + `LOG_LEVEL` finally wired.
- **A3** — `{ capabilities: { logging: {} } }` declared; the SDK auto-handles
  `logging/setLevel` and client-side level filtering.
- **Redaction** — a key denylist guarantees no token / code / verifier / secret
  / env value can ever reach a sink (covered by a dedicated test).

## Out of scope (next iterations)

- **B1**: surface the auth URL in the tool response (visible to the user).
- **B2**: non-blocking / fail-fast timeout.
- No API-token workaround (Zendesk's token model is too broad).

## Tests & gate

- New `tests/unit/utils/logger.test.ts` (level filtering, fields, redaction,
  MCP forward, silent logger).
- `browser-oauth.test.ts`: asserts the failure is **logged with diagnostics**
  instead of swallowed.
- `core-scenarios.ts`: asserts the `logging` capability is advertised over the
  wire (transport-agnostic).
[...]

PR #73, opened 2026-06-06
fix(auth): escape reflected values in OAuth callback HTML responses

dlecan:
## Vulnerability (CodeQL `js/reflected-xss`)

The local OAuth callback HTTP server (`src/auth/browser-oauth.ts`) echoed **attacker-controllable** values straight into its HTML response without escaping:

- the `error_description` query parameter (`desc`) → interpolated into `<p>${desc}</p>`;
- the token-exchange error message (`err.message`) → interpolated into the failure page.

An attacker can lure a victim into visiting a URL such as
`http://localhost:3000/callback?error=x&error_description=<script>…</script>`
and run JavaScript in the context of the locally served page → **reflected XSS**.

## Fix

- Added an `escapeHtml` helper (escapes `& < > " '`).
- Applied it to **every** reflected value in the callback responses.
- Regression test: a `<script>alert(1)</script>` payload comes back escaped (`&lt;script&gt;`) and never appears verbatim.

`pnpm check`, `pnpm typecheck` and the test suite all pass (233 tests).

## Why this slipped through the original PR without blocking

The file was introduced in `65ca70f` and instrumented in #62. CodeQL runs here in **default setup** — there is no `codeql.yml` in `.github/workflows/` (only `ci.yml` and `release.yml`). Code scanning is therefore **not a required check** on `pull_request`: CI (lint, typecheck, tests, build) was green while the CodeQL analysis ran outside the merge gate, so the alert was raised after the fact without ever blocking the merge.

### Suggestion (separate from this fix)
To block this in the future: enable branch protection requiring the Code Scanning check, or add a CodeQL workflow triggered on `pull_request` and mark it as required.

https://claude.ai/code/session_01R9ezGJu7sfDgD9GBB6S7mT

---
_Generated by [Claude Code](https://claude.ai/code/session_01R9ezGJu7sfDgD9GBB6S7mT)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a security vulnerability in OAuth authentication callbacks where error messages and descriptions were not properly escaped before display, potentially allowing malicious script injection. Error responses are now HTML-escaped.

* **Tests**
  * Added unit tests to verify that authentication error messages are properly escaped in OAuth callbacks and protected from reflected XSS attacks.

### DIFF
--- a/src/auth/browser-oauth.ts
+++ b/src/auth/browser-oauth.ts
@@ -32,6 +32,20 @@ interface TokenResult {
   scope: string;
 }
 
+/**
+ * Escape a string for safe interpolation into HTML text/attribute context.
+ * The local callback server echoes attacker-controllable values (the OAuth
+ * `error_description` query param, token-exchange error bodies) back into the
+ * browser response; without escaping these are a reflected-XSS sink.
+ */
+const escapeHtml = (value: string): string =>
+  value
+    .replace(/&/g, '&amp;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;');
+
 const generateCodeVerifier = (): string => randomBytes(32).toString('base64url');
 
 const generateCodeChallenge = (verifier: string): string =>
@@ -75,7 +89,9 @@ export const authenticateViaBrowser = (
       if (error) {
         const desc = url.searchParams.get('error_description') ?? error;
         res.writeHead(400, { 'Content-Type': 'text/html' });
-        res.end(`<html><body><h1>Authentication failed</h1><p>${desc}</p></body></html>`);
+        res.end(
+          `<html><body><h1>Authentication failed</h1><p>${escapeHtml(desc)}</p></body></html>`,
+        );
         clearTimeout(authTimeout);
         callbackServer.close();
         reject(new Error(`OAuth error: ${desc}`));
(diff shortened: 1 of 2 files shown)


src/auth/browser-oauth.ts
```diff
@@ -32,6 +32,20 @@ interface TokenResult {
   scope: string;
 }
 
+/**
+ * Escape a string for safe interpolation into HTML text/attribute context.
+ * The local callback server echoes attacker-controllable values (the OAuth
+ * `error_description` query param, token-exchange error bodies) back into the
+ * browser response; without escaping these are a reflected-XSS sink.
+ */
+const escapeHtml = (value: string): string =>
```
dlecan, line 41: Question : does it exist standard functions for that?

dlecan: Good question. There's no HTML-escaping function in the Node.js standard library — `node:querystring`/`URLSearchParams` handle URL encoding, but nothing in core escapes HTML text/attribute context.

The usual options are a small dependency (`escape-html`, `he`, or `lodash.escape`), but pulling one in for these two interpolations felt heavier than the inline 5-line helper, and it keeps the dependency surface minimal (in line with the functional, low-dep style of the repo).

We do already depend on `hast-util-to-html`, which escapes internally, but it's a full HAST→HTML serializer meant for the markdown pipeline — not exposed as a general-purpose `escape(string)`, so using it here would be awkward.

Happy to switch to a dedicated dependency (e.g. `escape-html`) instead if you'd prefer a vetted, reusable helper over the inline one — just let me know.

---

PR #40, opened 2026-05-25
feat: remote MCP deployment with per-user OAuth 2.1 PKCE

dlecan:
Closes #1.

> **HTTP transport ships as experimental.** Stdio is the supported path and is the one the local gate exercises end-to-end. The HTTP transport works in unit tests, in the smoke test, and against a local `curl`, but its end-to-end behavior against a real Zendesk tenant from every supported MCP client has not yet been validated. Until that is done, expect possible rough edges around OAuth discovery behind reverse proxies, CORS with browser-based clients, and the 401 / refresh flow. The same notice is on the corresponding README section so operators see it before deploying.

## Summary

Adds an HTTP transport so the same binary can run as a local stdio MCP server (laptop, Claude Desktop / Code / VS Code) or as a private remote MCP server (Claude.ai web, ChatGPT, etc.). The per-user OAuth 2.1 PKCE security model is preserved in both modes; API token auth is refused in HTTP because a shared static credential reachable over the network defeats the per-user model.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor (no external behavior change)
- [ ] Documentation
- [ ] Tooling / CI

## Author checklist
- [x] `pnpm test` passes locally (326 tests, was 280 pre-merge)
- [x] `pnpm test:coverage` meets the thresholds (gate 94/77/98/95)
- [x] `pnpm check` is clean (lint + format, with `--error-on-warnings`)
- [x] `pnpm typecheck` passes
- [x] I have read the diff myself, line by line
- [x] I ran a Claude Code review on the diff and addressed its findings
- [x] Documentation is updated where needed (README split into local / remote quick starts, env var table, CLI reference, experimental banner)
- [x] If this PR adds an MCP tool: it is documented in `README.md` (n/a: no new tools, only a new transport)

## What is now possible

- Run a private remote MCP server for your Zendesk account. Point any modern MCP client (Claude Desktop, Claude Code, claude.ai web, ChatGPT, Gemini, Copilot, Cursor, VS Code, Windsurf, Zed, Perplexity, Mistral) at `https://your-server/mcp`. It discovers the OAuth flow against Zendesk automatically and runs every tool call with the end user's permissions.
- Stdio still works exactly as before. No regression for the laptop plus Claude Desktop scenario; the CI / headless API token escape hatch is preserved.
[...]

PR #77, opened 2026-06-09
fix(http): address code-review findings on the HTTP transport (PR #40)

dlecan:
Follow-up to #40: fixes the 9 findings from a code review of the HTTP transport branch. Targets `claude/allons-traiter-1-Sxf5O` so it folds into #40 before merge.

## Correctness / security

1. **Authorization is now validated on every `/mcp` request** (was: only at session init — a leaked `mcp-session-id` alone granted access to the session's Zendesk token, and a revoked bearer was never re-checked). The latest presented bearer is adopted per session, so mid-session token refresh works. Locked by new tests: sessionful request without bearer → 401; rotated bearer reaches the Zendesk call (asserted via MSW).
2. **Request bodies are capped at 4 MB** → `413` (was: unbounded buffering before `JSON.parse`, read before any auth check on the existing-session path).
3. **Idle sessions are evicted after 30 minutes** (sweep every 60 s, `unref`'d). The SDK only fires `onclose` on an explicit DELETE, so clients that vanished leaked a `McpServer` + bearer per session forever. Timeout is injectable for tests.
4. **Malformed JSON now returns `400` + JSON-RPC `-32700` Parse Error** (was: escaped to the outer catch → `500` / `-32603`).
5. **`--cors-origin` / `CORS_ORIGIN` values are normalized to their URL origin** in `loadConfig` — `https://app.example.com/` (trailing slash, natural when copy-pasting) previously passed validation but could never match the browser's `Origin` header.

## Cleanup

6. Single strict `parsePort` helper shared by `--port`, `PORT`, `--callback-port` and `ZENDESK_OAUTH_CALLBACK_PORT` (the callback ports previously used bare `Number()`).
7. The `PUBLIC_URL` warning goes through the structured logger (`public_url_unset`) instead of raw `console.error`.
8. One `sendJsonRpcError` helper instead of three hand-built JSON-RPC error envelopes.
9. Dropped the `isOriginAllowed` wrapper (tests now assert on `resolveAllowedOrigin` directly); sessions close in parallel on shutdown.

## Verification

- `pnpm typecheck` / `pnpm check` clean
- `pnpm test`: 338 tests pass (was 326; +12 new/updated)
- `pnpm test:coverage`: 96.28 / 82.38 / 99.09 / 97.03 (gate 94/77/98/95)
- `pnpm test:smoke`: both `stdio_transport_ready` and `http_transport_ready` markers observed
- README HTTP section documents the new operational behavior (per-request auth, 4 MB cap, 30-min idle eviction)

Note: the test "routes a follow-up request to its existing session" previously locked the *absence* of auth on follow-ups; it was updated deliberately as part of fix 1, with a new companion test locking the 401.

https://claude.ai/code/session_01CZ61R9DAyjxijJWLgLTvxw

---
_Generated by [Claude Code](https://claude.ai/code/session_01CZ61R9DAyjxijJWLgLTvxw)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,25 +28,7 @@ jobs:
 
       - run: pnpm typecheck
 
-      # TEMPORARY DIAGNOSTIC (revert before merge): test:coverage hangs
-      # intermittently on GitHub runners only, and only when vitest writes to
-      # the runner's stdout pipe (file-redirected runs are consistently
-      # green). Keep the pipe condition via `tee`, capture everything to a
-      # file, and kill the whole process group at the bound so the step always
-      # exits with readable logs (the last running test + open handles).
-      - run: |
-          set +e
-          setsid bash -o pipefail -c 'pnpm test:coverage --reporter=verbose --reporter=hanging-process 2>&1 | tee vitest-diag.log' &
-          pgid=$!
-          ( sleep 240 && kill -TERM -- "-$pgid" 2>/dev/null; sleep 15 && kill -KILL -- "-$pgid" 2>/dev/null ) &
-          watcher=$!
-          wait "$pgid"
-          code=$?
-          kill "$watcher" 2>/dev/null
-          echo "=== exit code: $code ==="
-          echo "=== last 120 lines of vitest output ==="
-          tail -n 120 vitest-diag.log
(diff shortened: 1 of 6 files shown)


dlecan:
@coderabbitai review

---

PR #83, opened 2026-06-12
feat(help-center): expose structural topology via instructions + resource

dlecan:
## What & why

First batch ("lot 1") of #44 — the **structural** dimension. It moves the auto-discoverable parts of a Help Center's context out of a hand-distributed Claude Code skill and into the MCP server itself, so any compliant client gets it natively. The goal: stop the LLM guessing locales or fuzzy-matching section names, and have it use real IDs.

Delivered through a **hybrid** of two MCP-native channels (both gated on the `help_center` namespace and a new `--no-topology` opt-out):

1. **`instructions`** on `initialize` — a short, **static, I/O-free** blob (deliberately no API call at init, so it never triggers the lazy OAuth/PKCE flow just to connect). It names the subdomain and points at the topology resource.
2. **`zendesk-hc://topology`** — a **pull-only resource** read on demand **with the caller's own token** (so auth timing and ACL are both correct; #44 forbids a shared privileged token). It aggregates:
   - active locales + default
   - category → section tree with IDs (flat: category → section → article)
   - visibility user segments + permission groups (id + name)
   - the current user's role + id

## Design decisions (validated with the author)

- **Cache is per-session, not per-user-global, on purpose.** In HTTP mode `createMcpServer` (hence the topology provider + its cache) is built per session, so the memoized-promise cache is naturally per-caller — no cross-tenant leakage. The redundancy is acceptable because the resource is pull-only and the payload is light metadata cached for a short TTL; the per-caller fetch is required to respect each user's read permissions. Failures are evicted so the next read retries.
- **Large Help Centers**: categories/sections are capped at one max-size page. If sections overflow (`has_more`), the resource switches to a **summary mode** (categories only + a pointer to `list_sections`) instead of rendering a misleading partial tree. `truncateIfNeeded` remains the final guard.
- **Opt-out**: `--no-topology` disables instructions + resource together, server-wide.
- **401 on read** routes through `onUnauthorized` (stdio OAuth only), mirroring the tool dispatch path.

## Note for reviewers

This adds a **resource**, not a tool — so the tool-count assertions in `tests/unit/routing/registry.test.ts` and `tests/unit/tools/*.test.ts` are intentionally **unchanged**.

## Tests

- Unit: `buildInstructions` gating (namespace + `--no-topology`), `fetchTopology` aggregation + `has_more`, `formatTopology` (full tree + summary mode), provider cache (coalescing, failure eviction, 401 → `onUnauthorized`), `--no-topology` parsing.
[...]

PR #88, opened 2026-06-13
feat(auth): refresh OAuth tokens proactively and in the background

### LINKED ISSUE #86: Proactively refresh OAuth access tokens before expiry (Zendesk now expires tokens)
## Context

Zendesk has **enabled OAuth token expiration** (rolled out 2026-06-12). Access tokens now live ~12h maximum, or expire after ~8h of inactivity. In practice, every user's access token will be **expired when they start work in the morning**.

## Problem

`src/auth/token-store.ts` refreshes **lazily and reactively**:

- `getToken()` only refreshes when the cached token is already past `expiresAt - EXPIRY_SKEW_MS`.
- A token that Zendesk rejects mid-life is only recovered **after** a 401: the tool/resource handler calls `onUnauthorized` → `invalidate()` → the *next* `getToken()` refreshes.

With expiring tokens, the typical first request of the day becomes:

1. send the stale access token,
2. get a `401` from Zendesk,
3. `invalidate()` + refresh,
4. succeed on retry.

So **every user hits `error → refresh → works`** on their first action each morning — a poor first impression and avoidable noise.

This surfaced concretely during the #83 validation: the first `zendesk-hc://topology` read returned `-32603 Authentication failed` while a plain retry then succeeded (the token had been minted **before** expiry was enabled).

Aggravating factor: when the stored token has **no `expiresAt`** (minted before expiry was enabled, or a refresh response that omits `expires_in`), `isExpired()` returns `false` forever, so the dead token keeps being handed out until a 401 forces recovery.

## Proposal — be proactive

- Refresh **ahead of use**, not on 401:
  - On startup / first `getToken()`, if the token is expired, near-expiry, **or has no known `expiresAt`**, refresh silently *before* issuing the request.
  - Optionally schedule a background refresh shortly before `expiresAt` so a long-lived stdio session never serves a stale token.
- Harden `isExpired()` for tokens without `expiresAt` (treat unknown expiry conservatively, e.g. probe-refresh once).
- Keep the existing `401 → invalidate` path as a backstop.

## Acceptance

- A session whose stored token expired overnight performs its first request **without** a user-visible 401 — the refresh happens transparently first.
- Regression test: a stored token past expiry (and one with no `expiresAt`) triggers a silent refresh on the next `getToken()` **before** any API call.

## References

- `src/auth/token-store.ts` — `getToken`, `isExpired`, `tryRefresh`, `EXPIRY_SKEW_MS`
- `src/server.ts` — the reactive `onUnauthorized` path
- Observed in PR #83 (Help Center topology resource validation)

dlecan:
Closes #86.

## Context

Zendesk enabled OAuth token expiration (rolled out 2026-06-12): access tokens now live ~12h max, or expire after ~8h of inactivity. In practice every user's stored token is **expired when they start work in the morning**.

The previous refresh in `src/auth/token-store.ts` was **lazy and reactive**:
- `getToken()` only refreshed once the cached token was already past `expiresAt - EXPIRY_SKEW_MS`.
- A token with **no `expiresAt`** (minted before expiry was enabled, or a refresh response without `expires_in`) was treated as never-expiring → served forever until a 401 forced recovery.

So the first request of the day became `request → 401 → invalidate → refresh → retry` — a visible auth failure every morning (observed during #83 validation: the first `zendesk-hc://topology` read returned `-32603 Authentication failed`, a plain retry then succeeded).

## Changes

- **Refresh ahead of use** — `needsRefresh()` (replacing `isExpired()`) refreshes silently *before* the request when the token is expired, near expiry, or of **unknown age**. Unknown-expiry tokens are **probe-refreshed once**, then trusted, so we don't refresh on every call when `expires_in` keeps being omitted.
- **Exclusive refresh (single-flight)** — `getToken()` now awaits any in-flight refresh before acting, so concurrent callers and the background refresh share one network round-trip. This matters because **Zendesk rotates the refresh token on every use** — two concurrent refreshes would invalidate each other. The JS idiom here is promise-coalescing (the shared `refreshing` promise *is* the lock); no `async-mutex` dependency needed.
- **Background refresh every 4h** — an `unref()`'d interval (stdio-only; HTTP carries a per-session bearer and doesn't use this store) so a long-lived, idle session never serves a stale token, comfortably under Zendesk's ~8h inactivity window. Exposed `dispose()` to stop it.
- **401 → invalidate** path kept as a backstop.

## Tests

`tests/unit/auth/token-store.test.ts`:
- stored token expired overnight → silent refresh, no browser flow;
- stored token with **no `expiresAt`** → proactive refresh before any API call;
- unknown-expiry token whose refresh also omits `expires_in` → refreshed **once**, then trusted;
- 3 concurrent `getToken()` → a **single** `refreshAccessToken` call;
- background timer fires at 4h and refreshes without any `getToken()`;
- `dispose()` stops the timer.

All 358 tests pass; Biome `check`, `tsc --noEmit`, and `build` are green. README auth prose updated to describe the proactive + periodic refresh.

https://claude.ai/code/session_0113zrvJQkDmaceCFf5t26d3

---
_Generated by [Claude Code](https://claude.ai/code/session_0113zrvJQkDmaceCFf5t26d3)_

## Summary by CodeRabbit

* **New Features**
  * Added periodic background token refresh to keep sessions active during idle periods.
  * Added token store cleanup via a `dispose` method to stop background refresh when no longer needed.
* **Documentation**
  * Expanded OAuth token renewal guidance to clarify proactive refresh timing and when browser sign-in is required.
* **Tests**
  * Added unit tests covering refresh coalescing, handling tokens with unknown expiry, resilient behavior on transient refresh failures, and `dispose` behavior.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -3,6 +3,9 @@
     "commit": "",
     "pr": ""
   },
+  "permissions": {
+    "allow": ["mcp__zendesk-local"]
+  },
   "hooks": {
     "PreToolUse": [
       {
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -3,6 +3,7 @@
     "commit": "",
     "pr": ""
   },
+  "includeCoAuthoredBy": false,
   "permissions": {
     "allow": ["mcp__zendesk-local"]
   },
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -4,9 +4,7 @@
     "pr": ""
   },
   "includeCoAuthoredBy": false,
-  "permissions": {
-    "allow": ["mcp__zendesk-local"]
-  },
+  "enabledMcpjsonServers": ["zendesk-local"],
   "hooks": {
     "PreToolUse": [
       {
--- a/.claude/skills/functional-validation-plan/SKILL.md
+++ b/.claude/skills/functional-validation-plan/SKILL.md
@@ -1,0 +1,69 @@
+---
+name: functional-validation-plan
+description: Produce a functional validation plan for a feature/bugfix, to be executed by an INDEPENDENT validator (another agent or a human), not the implementer. Use whenever you write an implementation plan, or land a feature/bugfix that warrants end-to-end validation. The plan lives in the PR description; the validator posts their report as a PR comment.
+---
+
+# Functional validation plan — author protocol
+
+A code change is not "done" when the unit tests pass. Someone who did **not**
+write the code must be able to exercise the feature end-to-end and confirm it
+behaves as claimed. This skill produces the brief that lets them do that.
+
+## When to apply
+
+- Whenever you produce an **implementation plan**: the plan must include a
+  functional validation plan as a first-class section, not an afterthought.
+- Whenever you land a feature or bugfix whose behaviour is observable at runtime
+  (auth flows, transports, tool surface, persistence, timing).
+
+Skip only for changes with no runtime-observable behaviour (pure docs, comments,
+formatting) — say so explicitly rather than silently omitting it.
+
+## Hard rules
+
+- **Independent validator.** Write the plan FOR someone other than the
+  implementer — another agent (e.g. the local executor) or a human. Never assume
+  shared context: spell out setup, env vars, and how to observe results.
+- **Lives in the PR description.** Put the plan in the PR body under a
+  `## Functional validation plan` heading so it travels with the PR. Keep it in
+  sync if the change evolves.
+- **Report goes to a PR comment.** The validator posts their findings as a PR
+  comment (English — `AGENTS.md` rule), referencing each scenario id with an
+  OK/FAIL verdict and the observed evidence (logs, file state, tool output).
+- **No waiting on real time.** Prefer levers that simulate state (edit a
+  persisted timestamp, lower an interval constant, point at an unroutable host)
+  over waiting out a real timeout. Call out which behaviours are already covered
+  by fake-timer unit tests so the validator doesn't re-prove them needlessly.
+- **Distinct from `/functional-testing`.** That harness drives the committed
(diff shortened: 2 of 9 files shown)


.claude/skills/functional-validation-plan/SKILL.md
```diff
@@ -0,0 +1,97 @@
+---
+name: functional-validation-plan
+description: Produce a functional validation plan for a feature/bugfix, to be executed by an INDEPENDENT validator (another agent or a human), not the implementer. Use whenever you write an implementation plan, or land a feature/bugfix that warrants end-to-end validation. The plan lives in the PR description; the validator posts their report as a PR comment.
```
dlecan: Nitpick : as usual the "how" and the "when" are inverted
The description is the "when" to launch the skill

The how must the body of the skill, never in the description

.claude/settings.json
```diff
@@ -3,6 +3,10 @@
     "commit": "",
     "pr": ""
   },
+  "includeCoAuthoredBy": false,
+  "permissions": {
+    "allow": ["mcp__zendesk-local"]
```
dlecan: Suggestion: this MCP should be permanently allowed in the settings, not just the tools

.claude/skills/run-validation-plan/SKILL.md
```diff
@@ -0,0 +1,68 @@
+---
+name: run-validation-plan
+description: Execute the functional validation plan from a PR description as the INDEPENDENT validator, then post the report as a PR comment. Use when asked to validate/QA a feature on a branch, or when handed a "Functional validation plan". Counterpart to the `functional-validation-plan` skill (the author side); this is the executor side.
```
dlecan: Nitpick: same remark as for the other skill
You should use the skill creator to better understand 
https://github.com/anthropics/skills/blob/main/skills/skill-creator/SKILL.md

PR #110, opened 2026-07-01
feat(tickets): attach files to comments and notes via the Uploads API

### LINKED ISSUE #15: Support markdown-inline images in comment bodies for get_ticket_attachments
## Context

Follow-up from #13 (merged). During manual testing, @florentleveque identified an edge case not covered by `get_ticket_attachments`: images referenced **inline in the comment body** as markdown `alt` are not enrolled in the comment's `attachments` array returned by the Zendesk API, so they are invisible to the new tool.

## Current behavior

`get_ticket_attachments` collects images via `comment.attachments[]`. Images that authors embed directly in the message body (drag-and-drop screenshot turned into a markdown image link, or pasted image URLs) are skipped silently.

## Desired behavior

Detect and surface markdown-inline image URLs from `comment.body` alongside the regular attachments. Open questions to settle when picking this up:

- Same caps as `get_ticket_attachments` (5 MB / image, 10 embedded / call) ?
- Auth header scoping: these URLs may point to Zendesk CDN hosts or to external hosts — the same host-matching rule as `fetchZendeskBinary` should apply (no token forwarded to non-`${subdomain}.zendesk.com` hosts).
- Format: returned as additional `image` content blocks in `get_ticket_attachments`, or as a separate tool? Probably the former for consistency.

## Priority

Minor. The common case (drag-and-drop into the Zendesk composer) does populate `attachments`. This covers manually-pasted markdown image links.

Refs #13.

### LINKED ISSUE #106: Support attaching files to ticket comments and notes (Zendesk Uploads API)
## Problem

Through the MCP, there is no way to attach a file to a ticket comment or
internal note. `add_private_note` and `add_public_comment` accept text only, and
`update_ticket` carries no file either. The only upload-capable tool today,
`create_article_attachment`, targets Help Center articles, not tickets.

As a result, any workflow that needs to attach a file to a ticket (logs,
exports, documents, screenshots, etc.) is impossible through the server and has
to fall back to the Zendesk web UI.

## How Zendesk actually models this (verified against the API docs)

Files are never attached to a ticket directly. The flow is always **Uploads API
→ token → comment**:

1. **Upload the file.** `POST /api/v2/uploads.json?filename={name}` with the raw
   file bytes as the body and a correct `Content-Type` MIME header. The response
   is a `201` carrying an `upload` object with a `token`, an `expires_at`, the
   `attachment`, and the cumulative `attachments` list.
2. **Attach to a comment.** When creating/updating a ticket, set the tokens on
   the **comment** object via `ticket.comment.uploads: ["<token>", ...]`. There
   is no way to attach to the ticket itself - only to a comment added in the same
   create/update call.

Key constraints to respect:

- **Field name is `uploads`** on the comment object (write side). `attachments`
  is the read-side list of attachment objects returned by the API - not an input.
- **Token is valid 60 minutes** (`expires_at`); upload and attach within that
  window.
- **Multiple files under one token**: upload the first file to get a token, then
  pass that token back on each subsequent upload via the query param
  `?filename={name}&token={token}`. The token is *not* single-use; it aggregates
  files until consumed by a comment.

API reference:
[Ticket attachments / Uploads](https://developer.zendesk.com/api-reference/ticketing/tickets/ticket-attachments/)
and
[Adding ticket attachments with the API](https://developer.zendesk.com/documentation/ticketing/managing-tickets/adding-ticket-attachments-with-the-api/).

## Proposed solution

- A dedicated tool (e.g. `upload_file`) that calls
  `POST /api/v2/uploads.json?filename=...`, optionally chaining `&token=...` for
  additional files, and returns the upload token.
- An `uploads: ["<token>"]` parameter on `add_private_note`,
  `add_public_comment` and `update_ticket`, mapped to `comment.uploads` so the
  created comment carries the file(s). On `update_ticket` the tokens attach to
  the comment, never to the bare ticket.
- Optional convenience: also accept a local file path (and/or base64 content) as
  an `attachments`-style tool parameter and handle upload + token aggregation +
  association internally, so the caller does not have to juggle tokens or
  expiry. (This `attachments` parameter is tool sugar, distinct from Zendesk's
  read-side `attachments` field.)

## Alternatives considered

- Attaching manually in the Zendesk UI: breaks an otherwise assistant-driven
  workflow, especially when the file is generated by the assistant.
- Inlining the file content in the comment body: not a real attachment, and
  unusable for binary or large files.

## Scope

- [ ] Read-only (no data is modified in Zendesk)
- [x] Write operation (creates an upload and attaches it to a ticket comment)

## Additional context

This is a general limitation of the server, independent of any specific use
case: there is currently no path to attach a file when commenting on a ticket.

florentleveque:
## Summary

Add ticket attachment support to the comment tools. `add_public_comment` and
`add_private_note` gain an optional `attachments` parameter (base64-encoded
files); the handler uploads each file through the Zendesk Uploads API and carries
the resulting token on the comment via `comment.uploads`, so the created
comment/note ships the file(s).

This PR also fixes the read side (issue #15): `get_ticket_attachments` and
`get_ticket(include_comments)` now pass `include_inline_images=true` when listing
comments, so images embedded inline in a comment body (drag-and-drop /
paste-from-clipboard) are surfaced as attachments. Without the flag Zendesk omits
them from the comment attachments array, making them invisible. The two topics
are bundled here at the maintainer-author's request given the one-line nature of
the inline fix.

Closes #106. Closes #15.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tooling / CI

## Author checklist
- [x] `pnpm test` passes locally (391 tests)
- [x] `pnpm typecheck` passes
- [x] Documentation is updated where needed (README tool table)
- [x] If this PR adds an MCP tool: N/A — no new tool, only a new optional param on two existing tools

Design choices, all verified against the Zendesk Uploads API docs:

- **All-in-one, not token-juggling.** Rather than a separate `upload_file` tool
  returning a token plus an `uploads: [token]` param, the caller passes files
  directly and the handler does upload + aggregation + association internally.
  Simplest caller UX; no 60-minute token expiry leaked to the LLM.
- **Base64 only** (no local file path), consistent with `create_article_attachment`
  and required for the HTTP transport where the server can't read a client-side path.
- **`update_ticket` intentionally not extended.** Zendesk has no direct ticket
  attachment — a file always rides a comment (`comment.uploads`). `update_ticket`
  posts no comment, so attachments live on the two comment tools only.
- **Multi-file aggregation** under a single upload token via the `?token=` query
  param on subsequent uploads (`POST /uploads?filename=...&token=...`).
- **New `zendeskUpload` client helper** for a raw-binary POST: `executeRequest`
  always JSON-encodes, so the Uploads API (raw bytes + file MIME `Content-Type`)
  needs a direct `fetch`, mirroring the existing `helpCenterUpload`.

No new behavior was invented: field names (`comment.uploads`), the 60-minute token
validity, and the aggregation mechanism are all from the Uploads API reference.
Live-validated end-to-end on a real ticket (upload → `comment.uploads` →
`get_ticket_attachments` confirms the file), in addition to the unit + integration
tests.

**Author-side code review (high effort) — findings and disposition:**

- *Fixed* — `content_type` accepted an explicit empty string (the schema default
  only fills a missing field), which would have sent an empty `Content-Type`
  header. Added `.min(1)`, matching `file_name` / `file_base64`, plus two schema
  tests. (commit `fix(tickets): reject empty content_type ...`)
- *Deferred (out of scope)* — the two comment handlers share ~90% of their logic,
  and `attachmentSchema` overlaps with `create_article_attachment`'s schema. Both
  are pre-existing shapes; folding them into a `public: boolean` factory or a
  shared exported schema is a cross-cutting refactor beyond this feature's scope
  (repo's "scope discipline" rule). Happy to open a follow-up if desired.
- *Not a defect* — `uploadAttachments` uses `return uploadToken as string`; both
  callers are guarded by `attachments?.length`, so the empty-array path is
  unreachable (verifier verdict: PLAUSIBLE, not confirmed). The assertion is
  documented in a code comment.

## Functional validation plan

**Context.** The feature adds an optional `attachments` array to two tools:
`mcp__zendesk-local__add_public_comment` and `mcp__zendesk-local__add_private_note`.
Each entry is `{ file_name, file_base64, content_type }`. The handler uploads the
file(s) via the Zendesk Uploads API and attaches them to the created comment/note.
Observe the result with `mcp__zendesk-local__get_ticket_attachments`. No other tool
exercises this code; `update_ticket` deliberately does not.

**Setup & prerequisites.** The environment is already on the PR branch with the
`zendesk-local` MCP server running and authenticated (real Zendesk round-trips
work). Record the tested commit: `git rev-parse HEAD`. Create a throwaway ticket to
avoid polluting real data:
`mcp__zendesk-local__create_ticket { "subject": "attach-validation", "description": "scratch" }`
and note its id as `TID`. Ready-to-use base64 payloads (plain text): `"hello"` =
`aGVsbG8=` (5 bytes), `"world!"` = `d29ybGQh` (6 bytes). A credential/egress error is
a `BLOCKED` finding, not a setup task.

**Scenarios.**

| id | Validates | Manipulation / call | Expected observable |
| S1 | Private note with one attachment | `add_private_note { ticket_id: TID, body: "note+file", attachments: [{ file_name: "s1.txt", file_base64: "aGVsbG8=", content_type: "text/plain" }] }` | Returns text `Private note added to ticket #TID with 1 attachment(s).` |
| S2 | Public comment with one attachment | `add_public_comment { ticket_id: TID, body: "reply+file", attachments: [{ file_name: "s2.txt", file_base64: "aGVsbG8=", content_type: "text/plain" }] }` | Returns text `Public comment added to ticket #TID with 1 attachment(s).` |
| S3 | Attachment actually lands on the ticket | `get_ticket_attachments { ticket_id: TID }` | Lists `s1.txt` and `s2.txt`, each `text/plain`, `5 bytes`, with a `fruggr.zendesk.com/attachments/token/.../?name=...` URL |
| S4 | Multi-file aggregation under one token | `add_public_comment { ticket_id: TID, body: "two files", attachments: [{ file_name: "a.txt", file_base64: "aGVsbG8=", content_type: "text/plain" }, { file_name: "b.txt", file_base64: "d29ybGQh", content_type: "text/plain" }] }` then `get_ticket_attachments { ticket_id: TID }` | Return text says `with 2 attachment(s).`; the two new files `a.txt` (5 bytes) and `b.txt` (6 bytes) both appear on the ticket |
| S5 | No-attachment path unchanged | `add_private_note { ticket_id: TID, body: "plain note" }` | Returns `Private note added to ticket #TID.` with **no** `with N attachment(s)` suffix |
| S6 | Strict input validation | `add_public_comment { ticket_id: TID, body: "x", attachments: [{ file_name: "c.txt" }] }` (missing `file_base64`) | Tool error naming the missing required field; nothing posted |
| S7 | Inline comment images surfaced (#15) | On a ticket whose comment body has an image pasted/dragged inline (create one in the Zendesk UI, or reuse an existing ticket, and note its id `IID`): `get_ticket_attachments { ticket_id: IID }` | The inline image appears among the returned attachments/images (it was previously omitted). `get_ticket { ticket_id: IID, include_comments: true }` also lists it under the comment's attachments |

**Long-running behaviours.** None. The feature is synchronous (upload then PUT in
the same call); the 60-minute token validity is never waited on. Multi-file
aggregation and the no-attachment path are also covered by unit tests in
`tests/unit/tools/tickets.test.ts` and the wire round-trip by
`tests/integration/core-scenarios.ts`.

**Priorities.** S1, S2, S3 are the real user-facing paths — do them first. S4 covers
aggregation, S5 guards against regression, S6 is defensive.

**Reporting.** Post a PR comment (English): per-id `OK`/`FAIL`/`BLOCKED` with the
observed evidence (returned tool text, the `get_ticket_attachments` listing), the
tested commit SHA, and an overall green/failing summary.

## Summary by CodeRabbit

* **New Features**
  * You can now add file attachments when creating private notes and public comments on tickets.
  * Ticket comments can now include multiple attachments, with upload status reflected in the response.

* **Bug Fixes**
  * Ticket comment views now include inline images when comments are fetched.
  * Attachment inputs are now validated more strictly, including file type and base64 content.

* **Documentation**
  * Updated the tool documentation to mention optional attachments.

### DIFF
--- a/src/tools/tickets.ts
+++ b/src/tools/tickets.ts
@@ -5,6 +5,7 @@ import {
   zendeskGet,
   zendeskPost,
   zendeskPut,
+  zendeskUpload,
 } from '../client/zendesk-api';
 import {
   DEFAULT_PAGE_SIZE,
@@ -20,6 +21,7 @@ import type {
   ZendeskSlaSideloadEntry,
   ZendeskTicket,
   ZendeskTicketAttachment,
+  ZendeskUpload,
 } from '../types';
 import {
   formatComment,
@@ -165,6 +167,35 @@ const fetchTicketSla = async (
 
 export const createTicketTools = (ctx: ToolContext): ToolDefinition[] => {
   const { subdomain, getToken } = ctx;
+
+  const attachmentSchema = z.object({
+    file_name: z.string().min(1).describe('File name, e.g. "app.log" or "screenshot.png".'),
+    file_base64: z.string().min(1).describe('File content encoded as base64.'),
+    content_type: z
+      .string()
+      .default('application/octet-stream')
+      .describe('MIME type, e.g. "text/plain", "image/png", "application/pdf".'),
+  });
+  type AttachmentInput = z.infer<typeof attachmentSchema>;
+
+  // Upload each file via the Zendesk Uploads API, aggregating them under a single
+  // upload token (the token from the first upload is passed to the next), and
+  // return that token for use in a comment's `uploads` array.
+  const uploadAttachments = async (token: string, files: AttachmentInput[]): Promise<string> => {
+    let uploadToken: string | undefined;
+    for (const file of files) {
+      const { upload } = await zendeskUpload<{ upload: ZendeskUpload }>(
+        subdomain,
+        token,
+        file.file_name,
+        Buffer.from(file.file_base64, 'base64'),
+        file.content_type,
+        uploadToken,
+      );
+      uploadToken = upload.token;
+    }
+    return uploadToken as string;
+  };
 
   return [
     {
@@ -458,10 +489,15 @@ export const createTicketTools = (ctx: ToolContext): ToolDefinition[] => {
       namespace: 'tickets',
       readOnly: false,
       title: 'Add Private Note',
-      description: 'Add an internal note (not visible to requester) to a ticket.',
+      description:
+        'Add an internal note (not visible to requester) to a ticket, optionally with file attachments (uploaded via the Zendesk Uploads API and carried on the note).',
       inputSchema: z.object({
         ticket_id: z.number().int().describe('Ticket ID'),
         body: z.string().min(1).describe('Note content'),
+        attachments: z
+          .array(attachmentSchema)
+          .optional()
+          .describe('Files to attach to this note (base64-encoded content).'),
       }),
       annotations: {
         readOnlyHint: false,
@@ -470,12 +506,22 @@ export const createTicketTools = (ctx: ToolContext): ToolDefinition[] => {
         openWorldHint: true,
       },
       handler: async (params) => {
-        const { ticket_id, body } = params as { ticket_id: number; body: string };
+        const { ticket_id, body, attachments } = params as {
+          ticket_id: number;
+          body: string;
+          attachments?: AttachmentInput[];
+        };
         const token = await getToken();
+        const uploads = attachments?.length
+          ? [await uploadAttachments(token, attachments)]
+          : undefined;
         await zendeskPut(subdomain, token, `/tickets/${ticket_id}`, {
-          ticket: { comment: { body, public: false } },
+          ticket: { comment: { body, public: false, ...(uploads && { uploads }) } },
         });
-        return { content: [{ type: 'text', text: `Private note added to ticket #${ticket_id}.` }] };
+        const suffix = attachments?.length ? ` with ${attachments.length} attachment(s)` : '';
+        return {
+          content: [{ type: 'text', text: `Private note added to ticket #${ticket_id}${suffix}.` }],
+        };
       },
     },
     {
@@ -483,10 +529,15 @@ export const createTicketTools = (ctx: ToolContext): ToolDefinition[] => {
       namespace: 'tickets',
       readOnly: false,
       title: 'Add Public Comment',
-      description: 'Add a public comment (visible to requester) to a ticket.',
+      description:
+        'Add a public comment (visible to requester) to a ticket, optionally with file attachments (uploaded via the Zendesk Uploads API and carried on the comment).',
       inputSchema: z.object({
         ticket_id: z.number().int().describe('Ticket ID'),
         body: z.string().min(1).describe('Comment content'),
+        attachments: z
+          .array(attachmentSchema)
+          .optional()
+          .describe('Files to attach to this comment (base64-encoded content).'),
       }),
       annotations: {
         readOnlyHint: false,
@@ -495,13 +546,23 @@ export const createTicketTools = (ctx: ToolContext): ToolDefinition[] => {
         openWorldHint: true,
       },
       handler: async (params) => {
-        const { ticket_id, body } = params as { ticket_id: number; body: string };
+        const { ticket_id, body, attachments } = params as {
+          ticket_id: number;
+          body: string;
+          attachments?: AttachmentInput[];
+        };
         const token = await getToken();
+        const uploads = attachments?.length
+          ? [await uploadAttachments(token, attachments)]
+          : undefined;
         await zendeskPut(subdomain, token, `/tickets/${ticket_id}`, {
-          ticket: { comment: { body, public: true } },
+          ticket: { comment: { body, public: true, ...(uploads && { uploads }) } },
         });
+        const suffix = attachments?.length ? ` with ${attachments.length} attachment(s)` : '';
         return {
-          content: [{ type: 'text', text: `Public comment added to ticket #${ticket_id}.` }],
+          content: [
+            { type: 'text', text: `Public comment added to ticket #${ticket_id}${suffix}.` },
+          ],
         };
       },
     },
--- a/src/tools/tickets.ts
+++ b/src/tools/tickets.ts
@@ -173,6 +173,7 @@ export const createTicketTools = (ctx: ToolContext): ToolDefinition[] => {
     file_base64: z.string().min(1).describe('File content encoded as base64.'),
     content_type: z
       .string()
+      .min(1)
       .default('application/octet-stream')
       .describe('MIME type, e.g. "text/plain", "image/png", "application/pdf".'),
   });
--- a/src/tools/tickets.ts
+++ b/src/tools/tickets.ts
@@ -173,7 +173,7 @@ export const createTicketTools = (ctx: ToolContext): ToolDefinition[] => {
 
   const attachmentSchema = z.object({
     file_name: z.string().min(1).describe('File name, e.g. "app.log" or "screenshot.png".'),
-    file_base64: z.string().min(1).describe('File content encoded as base64.'),
+    file_base64: z.string().min(1).base64().describe('File content encoded as base64.'),
     content_type: z
       .string()
       .min(1)
--- a/src/tools/tickets.ts
+++ b/src/tools/tickets.ts
@@ -70,7 +70,10 @@ const fetchAllTicketComments = async (
     const response = await zendeskGet<{
       comments: ZendeskComment[];
       meta?: { has_more: boolean; after_cursor: string };
-    }>(subdomain, token, `/tickets/${ticketId}/comments`, buildCursorParams(MAX_PAGE_SIZE, cursor));
+    }>(subdomain, token, `/tickets/${ticketId}/comments`, {
+      ...buildCursorParams(MAX_PAGE_SIZE, cursor),
+      include_inline_images: 'true',
+    });
     all.push(...response.comments);
     pages += 1;
     if (!response.meta?.has_more || !response.meta?.after_cursor) break;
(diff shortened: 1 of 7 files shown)


florentleveque:
@dlecan this one is ready for your review whenever you have a moment.

It bundles two topics (kept together given how small the second one is):
- **#106** — attach files to ticket comments/notes via the Uploads API (new `attachments` param on `add_public_comment` / `add_private_note`).
- **#15** — the inline-images fix you had marked "Planned": `include_inline_images=true` is now passed when listing comments, so images embedded inline in a comment body surface as attachments on both `get_ticket_attachments` and `get_ticket(include_comments)`.

State:
- Local gate green: `pnpm check` / `typecheck` / `test` / `test:coverage` / `build`.
- Author-side AI review + a `/simplify` pass applied; findings either fixed or documented in the description.
- Live-validated end to end on real tickets (attachment upload round-trip, and inline images now surfaced).

A functional validation plan (S1-S7) for an independent validator is in the PR description. Happy to split #15 into its own PR if you'd rather review them separately.

src/tools/tickets.ts
```diff
@@ -166,6 +171,39 @@ const fetchTicketSla = async (
 export const createTicketTools = (ctx: ToolContext): ToolDefinition[] => {
   const { subdomain, getToken } = ctx;
 
+  const attachmentSchema = z.object({
+    file_name: z.string().min(1).describe('File name, e.g. "app.log" or "screenshot.png".'),
+    file_base64: z.string().min(1).describe('File content encoded as base64.'),
```
dlecan: `file_base64` only checks `.min(1)`, so any non-empty string is accepted. `Buffer.from(x, 'base64')` is lenient and silently drops invalid characters instead of throwing, so a malformed payload is uploaded as corrupt bytes rather than rejected. Add `.base64()` so bad input fails validation up front (clear error, nothing posted).

I verified the wire schema does not regress: with `zod/v4-mini` → draft-07 (the exact path the MCP SDK 1.29 uses), the `description` stays byte-for-byte identical and the schema only *gains* `format: "base64"` + a `pattern` — an enrichment, standard and multi-agent-safe.

```suggestion
    file_base64: z.string().min(1).base64().describe('File content encoded as base64.'),
```

---

dlecan (changes requested):
Solid PR — design is sound and matches the Uploads API (token chaining across `?token=`, `comment.uploads`, `include_inline_images=true` on List Comments is a real, documented param and inline images are omitted by default so the #15 fix is justified). CI is green and the tests are well targeted. One change requested before merge.

### Requested change — validate `file_base64` as base64
See the inline suggestion on `src/tools/tickets.ts`. `.min(1)` lets any non-empty string through, and `Buffer.from(x, 'base64')` silently mangles invalid input instead of rejecting it, so a malformed payload uploads as corrupt bytes. Adding `.base64()` fails it at the schema boundary.

Please also add a test to lock the behavior (mirrors the existing empty-`content_type` test), in `tests/unit/tools/tickets.test.ts`:

```ts
it('rejects an attachment whose file_base64 is not valid base64', () => {
  const tool = findTool('add_public_comment');
  const result = tool.inputSchema.safeParse({
    ticket_id: 1,
    body: 'x',
    attachments: [
      { file_name: 'a.txt', file_base64: 'not base64!!', content_type: 'text/plain' },
    ],
  });
  expect(result.success).toBe(false);
});
```

I applied both locally on the PR head and the full gate stays green: `typecheck` / `check` / `test` (396) / `test:coverage` (statements 96.3, branches 82.2, functions 98.5, lines 97.2 — all above threshold).

### Tool-description quality (no regression)
Checked deliberately since the server is indexed on Glama. Reproducing the exact SDK 1.29 conversion (`zod/v4-mini` → draft-07), the `file_base64` `description` stays byte-for-byte identical and the emitted schema only gains `format: "base64"` + `pattern` — an enrichment, not a regression, and standard/multi-agent-safe. The two tool descriptions this PR rewrites (`add_private_note` / `add_public_comment`) also remain single standalone sentences, so the first-sentence proxy descriptions stay correct. 👍

### Intentionally not requested
Noted and fine as-is for this PR: the ~90% overlap between the two handlers (author already deferred it — agreed, it's a cross-cutting refactor), the `return uploadToken as string` cast (unreachable empty path, documented), and `get_ticket(include_comments)` only reading the first comments page (pre-existing; the inline-images fix is correct for the comments it returns). No action needed on those here.

---

dlecan:
Thanks a lot for this contribution, @florentleveque — really clean work: the design matches the Uploads API, the tests are thorough, and bundling the small #15 inline-images fix here makes sense. Just the one base64-validation change requested above (suggestion + test provided) and it's good to go. Ping me once you've pushed and I'll re-check and merge.

---

dlecan (approved):
The requested base64 validation is in (commit `f2cff4d`) — `.base64()` on `file_base64` plus the rejection test, exactly as suggested. Re-verified locally against this head: `typecheck` / `check` / `test` (396) all green, and CI is green here too. Descriptions/schema confirmed non-regressing (the `file_base64` description is unchanged; the schema only gains `format: "base64"`). Approving — thanks for the quick turnaround, @florentleveque.

---

PR #154, opened 2026-07-12
chore: allow in-session tool hot-reload with --dev mode

dlecan:
## Summary
Adds a stdio-only `--dev` mode that exposes one tool, **`reload_tools`**. Calling it re-imports the tool modules from source and re-registers the toolset **in place** on the live session (via `notifications/tools/list_changed`) — so tool code you just edited takes effect **without restarting the process or reconnecting the client**.

The point is a tight **in-session edit → reload → test loop** for MCP tools: an agent (or a developer with an MCP client) edits a tool, calls `reload_tools`, and immediately exercises the updated tool over the same connection. No process bounce, no session teardown, no client reconnect between iterations. This is a genuinely different way to develop and test the server from the inside — the reload is explicit and developer-driven (not a file watcher firing on every save), so it stays predictable and produces one notification burst per cycle.

## Linked issue
None.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tooling / CI

## What changed
- **config**: `--dev` (CLI-only, off by default; documented as dev-only).
- **server**: split `createMcpServer` into `createServerShell` + `registerToolset`. `registerToolset` returns a disposable handle plus the tool count and is **atomic** — if a `registerTool`/`registerResource` throws partway (e.g. an edit introduced a duplicate tool name) it rolls back the handles already registered. `createMcpServer`'s external behaviour is unchanged; the product path gets the atomicity for free.
- **dev/reload**: `createReloadableServer` (server shell + a `reload()` that swaps in freshly re-imported tool code and returns the new count, restoring the last-good generation if re-registration fails), `registerReloadTool` (the persistent `reload_tools` meta-tool), and `startDevServer`. Reload re-imports the **leaf** tool modules with a per-call cache-busting nonce and recomposes them.
- **index**: route stdio startup through dev mode when `--dev` is set; warn-and-ignore under HTTP (per-request servers have nothing long-lived to hot-swap).
- **tests**: `--dev` config parsing; `registerToolset` atomicity/rollback; and an integration suite over an in-memory client covering a successful reload, `list_changed` on swap, `reload_tools` reporting a failed reload as a tool error (previous tools stay live), and restoration of the previous generation when re-registration collides.
- **docs**: "Dev mode" section in `docs/configuration.md`.

## How to use it
Point an MCP client's server command at the source and add the flag:
```jsonc
// .mcp.json
{ "command": "pnpm", "args": ["exec", "tsx", "src/index.ts", "--mode", "all", "--dev"], "env": { "ZENDESK_SUBDOMAIN": "acme" } }
```
Then, per edit cycle: edit a `src/tools/*.ts` file → call `reload_tools` → test the updated tool over the same session.

## Design notes / limits (by design)
- **On-demand, not a watcher.** The reload is an explicit tool call, so it stays predictable and produces one `list_changed` burst per cycle rather than one per file save.
- **Why leaf modules?** An ESM cache-bust does not cascade to a module's own imports — busting `tools/index.ts` keeps serving stale `tickets.ts` etc. (verified empirically). So reload busts the leaves where handler code lives. Edits to shared infra below them (HTTP client, `definitions.ts`, guidance, server wiring) still need a full restart. `TOOL_MODULES` must be kept in sync with `createAllTools` (drift is caught by the integration test, which compares the reload surface against `createAllTools`).
- **Resilient reload.** A reload that throws (syntax error mid-edit) returns a tool error and leaves the previous generation live.
- **`reload_tools` is dev-only** — never exposed without `--dev`, and not part of the product tool surface, so it stays out of `docs/mcp-tools-reference.md`.
- **stdio only** (HTTP builds a per-request server); **dev-only** (the published package ships compiled JS with no sources to reload).

## Follow-up (out of scope here)
For the in-session loop to *exercise* tools against Zendesk in a headless environment (CI, remote agent), the server still needs a token without the interactive browser PKCE flow. That token-injection story (seed a credential via an environment secret → `ZENDESK_TOKEN_FILE`, mindful that Zendesk rotates refresh tokens on every use) is being designed separately and is not part of this PR.

### Context
The feature is the `--dev` startup mode for the **stdio** transport, which adds the `reload_tools` tool. It is observable only when the server is launched with `--dev` from source: after editing a `src/tools/*.ts` file, calling `reload_tools` must make the edit visible in `tools/list` without the session dropping. The HTTP transport does **not** exercise this (it ignores the flag). Note: listing/calling `reload_tools` requires **no Zendesk credentials** — it makes no Zendesk API call and the server serves `tools/list` before any auth — so this plan needs no token or egress.

If the MCP server already running in your environment was started with `--dev` and `--mode all`, you have `mcp__<server>__reload_tools` available and can drive V1–V2 through it directly. If it was not started with `--dev` (the likely default), use the self-contained repro script below, which launches a throwaway `--dev` server from the working tree. Either way, capture the SHA under test: `git rev-parse HEAD`.

### Setup & prerequisites
Beyond the ready checkout, nothing. Any file a scenario edits under `src/` must be **restored** afterward (the repro script does this in a `finally`; confirm `git status` is clean when done).

Save this repro helper as `scratch-dev-verify.mjs` at the repo root (self-contained; delete it when done):

```js
import { readFileSync, writeFileSync } from 'node:fs';
import { setTimeout as sleep } from 'node:timers/promises';
import { Client } from '@modelcontextprotocol/sdk/client/index.js';
import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js';

const FILE = 'src/tools/tickets.ts';
const MARKER = 'HOTRELOAD_MARKER_TITLE';
const original = readFileSync(FILE, 'utf8');
const transport = new StdioClientTransport({
  command: 'node',
  args: ['--import', 'tsx', 'src/index.ts', 'fruggr', '--dev', '--mode', 'all', '--log-level', 'error'],
  env: { ...process.env, ZENDESK_SUBDOMAIN: 'fruggr' },
  stderr: 'inherit',
});
const client = new Client({ name: 'verify', version: '0.0.0' });
const titleOf = async (n) => (await client.listTools()).tools.find((t) => t.name === n)?.title;
try {
  await client.connect(transport);
  console.log('reload_tools present =', (await client.listTools()).tools.some((t) => t.name === 'reload_tools'));
  console.log('BEFORE =', JSON.stringify(await titleOf('get_ticket')));
  writeFileSync(FILE, original.replace("title: 'Get Zendesk Ticket',", `title: '${MARKER}',`));
  const res = await client.callTool({ name: 'reload_tools', arguments: {} });
  console.log('reload result =', JSON.stringify(res.content?.[0]?.text), 'isError =', res.isError ?? false);
  await sleep(200);
  console.log('AFTER  =', JSON.stringify(await titleOf('get_ticket')));
} finally {
  writeFileSync(FILE, original);
  await client.close();
  process.exit(0);
}
```

Run with `node scratch-dev-verify.mjs`.

### Scenarios
| id | validates | manipulation | expected observable |
|----|-----------|--------------|---------------------|
| V1 | Edited tool code goes live on demand | run the repro helper (or, on a `--dev` live server, edit the file then call `reload_tools`) | `reload_tools present = true`; `BEFORE = "Get Zendesk Ticket"`; reload result `Reloaded N tool(s)...`, `isError = false`; `AFTER = "HOTRELOAD_MARKER_TITLE"` — same connected session |
| V2 | Dev tool survives its own reload | same run | `reload_tools` still present in `tools/list` after the call |
| V3 | Reload is resilient to a broken edit | edit `src/tools/users.ts` to contain a syntax error, call `reload_tools`, then restore and call again | first call returns `isError: true` with a message, previous tools stay live; after restore, the call succeeds. **Restore the file.** |
| V4 | HTTP ignores `--dev` | start `node --import tsx src/index.ts fruggr --dev --transport http --port 0 --log-level warn` briefly | stderr logs `dev_mode_ignored_http`; server starts normally; no `reload_tools` on the HTTP surface; kill it |
| V5 | Non-dev startup unaffected (regression) | `pnpm test` and `pnpm test:coverage` | all green; coverage thresholds met; default (non-`--dev`) tool surface unchanged (no `reload_tools`) |

### Long-running behaviours
None. The reconcile/reload pipeline, rollback, and the `reload_tools` tool are covered without a transport by `tests/integration/dev-reload.test.ts` and `tests/unit/server.test.ts`, so V1–V3 mainly confirm the tool works over a real stdio process.

### Priorities
V1 and V2 are the core path — do first. V3 (resilience) next. V4/V5 are guard rails.

### Reporting
Post a PR comment (English): per-id `OK`/`FAIL`/`BLOCKED` with evidence (console output for V1–V3, the relevant stderr line + a `tools/list` check for V4, the test summary for V5), plus the tested commit SHA and an overall green/failing summary.

- Flag name: `--dev` (a general dev-affordances switch; today it adds `reload_tools`). Open to `--dev-reload` if you'd rather name the single behaviour.
- The `TOOL_MODULES` list duplicates the composition in `createAllTools`; a shared manifest would fight ESM/TS resolution (static import symbols vs. an absolute `.ts` path for cache-busting) and would force `createAllTools` async, rippling into `createMcpServer`/`index.ts`. Kept as-is with the drift guarded by the integration test — but happy to revisit.
- This started as a spike and is now proposed as a first-class dev/test capability; the token-injection follow-up (above) is what makes the in-session loop able to hit Zendesk headless.

## Author checklist
- [x] I checked no open or merged PR already addresses this
- [x] N/A — no linked issue
- [x] `pnpm typecheck` passes
- [x] N/A — `reload_tools` is a dev-only tool, deliberately not in `docs/mcp-tools-reference.md`
- [ ] Functional validation plan present (above); independent validation pending

https://claude.ai/code/session_01D9RqKKFbEG47JUSAiBKnoj

## Summary by CodeRabbit

- **New Features**
  - Added stdio dev mode via the `--dev` flag.
  - Added `reload_tools` to hot-reload tools without restarting the server or reconnecting clients.
  - `reload_tools` supports in-place tool refresh with safe rollback on failure.

- **Documentation**
  - Expanded dev-mode documentation, including reload scope/limits and reload error behavior.

- **Bug Fixes**
  - Improved toolset registration to be atomic, preventing partial tool exposure after reload/registration issues.
  - Dev mode is now explicitly warned as ignored for non-stdio (HTTP) setups.

### DIFF
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -11,14 +11,14 @@ fi
 cd "$CLAUDE_PROJECT_DIR"
 
 # pnpm is pinned via package.json "packageManager"; Corepack selects that exact
-# version rather than trusting whatever pnpm happens to be on PATH.
+# version rather than trusting whatever pnpm is on PATH.
 #
-# Plain install (not --frozen-lockfile): this is a dev bootstrap, so if a session
-# adds or bumps a dependency the install should reconcile the lockfile rather
-# than fail the way a frozen/CI install would. --ignore-scripts installs
-# dependencies only, never running lifecycle scripts — it deliberately skips the
-# repo's own `prepare` (tsdown build) and any dependency install scripts, so a
-# session start never executes branch or package code (the dev/test loop runs
-# from source via tsx, and CI builds separately).
+# Plain install (not --frozen-lockfile): this is a dev bootstrap, so a session
+# that adds or bumps a dependency should reconcile the lockfile rather than fail
+# the way a frozen/CI install would. Dependency lifecycle scripts are already
+# governed by pnpm-workspace.yaml (the `allowBuilds` whitelist + the
+# `minimumReleaseAge` cooldown), so we do NOT pass --ignore-scripts — that would
+# skip whitelisted builds that packages need to work (e.g. esbuild's native
+# binary, which tsx/tsdown depend on).
 corepack enable >/dev/null 2>&1 || true
-corepack pnpm install --ignore-scripts
+corepack pnpm install
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -13,10 +13,12 @@ cd "$CLAUDE_PROJECT_DIR"
 # pnpm is pinned via package.json "packageManager"; Corepack selects that exact
 # version rather than trusting whatever pnpm happens to be on PATH.
 #
-# --ignore-scripts: install dependencies only, never run lifecycle scripts. This
-# deliberately skips the repo's own `prepare` (which builds via tsdown) and any
-# dependency install scripts, so a session start never executes branch or
-# package code — the dev/test loop runs from source via tsx, and CI builds
-# separately. It also keeps the hook cheap when node_modules is already cached.
+# Plain install (not --frozen-lockfile): this is a dev bootstrap, so if a session
+# adds or bumps a dependency the install should reconcile the lockfile rather
+# than fail the way a frozen/CI install would. --ignore-scripts installs
+# dependencies only, never running lifecycle scripts — it deliberately skips the
+# repo's own `prepare` (tsdown build) and any dependency install scripts, so a
+# session start never executes branch or package code (the dev/test loop runs
+# from source via tsx, and CI builds separately).
 corepack enable >/dev/null 2>&1 || true
-corepack pnpm install --frozen-lockfile --ignore-scripts
+corepack pnpm install --ignore-scripts
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+set -euo pipefail
+
+# Install dependencies at the start of a Claude Code on the web session so tests,
+# linters and the MCP server work out of the box. Web-only: local sessions manage
+# their own dependencies (and would re-run this on every start with no benefit).
+if [ "${CLAUDE_CODE_REMOTE:-}" != "true" ]; then
+  exit 0
+fi
+
+cd "$CLAUDE_PROJECT_DIR"
+
+# pnpm ships via Corepack (bundled with Node); enable it if pnpm isn't on PATH.
+if ! command -v pnpm >/dev/null 2>&1; then
+  corepack enable >/dev/null 2>&1 || true
+fi
+
+# Idempotent and fast when node_modules is already cached; --frozen-lockfile
+# keeps the install faithful to the committed pnpm-lock.yaml.
+pnpm install --frozen-lockfile
--- a/.claude/hooks/session-start.sh
+++ b/.claude/hooks/session-start.sh
@@ -10,11 +10,13 @@ fi
 
 cd "$CLAUDE_PROJECT_DIR"
 
-# pnpm ships via Corepack (bundled with Node); enable it if pnpm isn't on PATH.
-if ! command -v pnpm >/dev/null 2>&1; then
-  corepack enable >/dev/null 2>&1 || true
-fi
-
-# Idempotent and fast when node_modules is already cached; --frozen-lockfile
-# keeps the install faithful to the committed pnpm-lock.yaml.
-pnpm install --frozen-lockfile
+# pnpm is pinned via package.json "packageManager"; Corepack selects that exact
+# version rather than trusting whatever pnpm happens to be on PATH.
+#
+# --ignore-scripts: install dependencies only, never run lifecycle scripts. This
+# deliberately skips the repo's own `prepare` (which builds via tsdown) and any
+# dependency install scripts, so a session start never executes branch or
+# package code — the dev/test loop runs from source via tsx, and CI builds
+# separately. It also keeps the hook cheap when node_modules is already cached.
+corepack enable >/dev/null 2>&1 || true
+corepack pnpm install --frozen-lockfile --ignore-scripts
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,6 +15,8 @@ documentation.
 
 Toolchain (Node 24 + pnpm 11) is the dev floor; the published package still runs on Node 20+ (a CI job exercises the packed tarball on Node 20).
 
+On **Claude Code on the web**, dependencies are installed automatically at session start by the `SessionStart` hook (`.claude/hooks/session-start.sh`, web-only) — you can run tests/linters right away. Locally the hook is a no-op; run `pnpm install` yourself.
+
 ## Claim before you build — no duplicate work
 
 Parallel duplicate work has happened here. Before coding an issue: re-read it
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -15,7 +15,9 @@ documentation.
 
 Toolchain (Node 24 + pnpm 11) is the dev floor; the published package still runs on Node 20+ (a CI job exercises the packed tarball on Node 20).
 
-On **Claude Code on the web**, dependencies are installed automatically at session start by the `SessionStart` hook (`.claude/hooks/session-start.sh`, web-only) — you can run tests/linters right away. Locally the hook is a no-op; run `pnpm install` yourself.
+On **Claude Code on the web**, the Node environment is set up automatically at session start (dependencies installed), so tests and linters work right away.
+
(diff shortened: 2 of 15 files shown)


AGENTS.md
```diff
@@ -15,6 +15,8 @@ documentation.
 
 Toolchain (Node 24 + pnpm 11) is the dev floor; the published package still runs on Node 20+ (a CI job exercises the packed tarball on Node 20).
 
+On **Claude Code on the web**, dependencies are installed automatically at session start by the `SessionStart` hook (`.claude/hooks/session-start.sh`, web-only) — you can run tests/linters right away. Locally the hook is a no-op; run `pnpm install` yourself.
```
dlecan: Suggestion: not really useful to explain the how. Just explain the node environment is set up at start, that's enough 
However explanations about the "reload Dev tool to reload all tools" is missing and how to use it

.claude/hooks/session-start.sh
```diff
@@ -0,0 +1,20 @@
...
+cd "$CLAUDE_PROJECT_DIR"
+
+# pnpm ships via Corepack (bundled with Node); enable it if pnpm isn't on PATH.
+if ! command -v pnpm >/dev/null 2>&1; then
+  corepack enable >/dev/null 2>&1 || true
+fi
+
+# Idempotent and fast when node_modules is already cached; --frozen-lockfile
+# keeps the install faithful to the committed pnpm-lock.yaml.
+pnpm install --frozen-lockfile
```
dlecan: Question: why frozen deps in your own environment?

dlecan:
Addressed the review in `525844e`:

- **Session-start hook / lifecycle scripts (security).** Kept the hook committed in the repo on purpose — the goal is that every future web session (and agent run) installs deps automatically, which a user/org-level config would not guarantee. To remove the code-execution vector the finding is really about, the install now runs with `--ignore-scripts`, so no lifecycle scripts execute at session start — notably the repo's own `prepare` (tsdown build) and any dependency install scripts. Nothing in the dev/test loop needs them: the server and tests run from source via `tsx`, and CI builds separately. Sessions run on the author's own branches, so this is defense-in-depth rather than a fork-PR boundary.
- **Pinned pnpm.** The hook now selects the `package.json` `packageManager` pin (pnpm@11.10.0) via `corepack pnpm …` instead of trusting whatever `pnpm` is on `PATH`.
- **Reload rollback double-failure.** The rollback `registerToolset` is now wrapped in its own try/catch; a double failure is logged (`tools_rollback_failed`) and the original reload error is still surfaced to the caller.

If you'd prefer the hook not live in the repo at all (env setup script / opt-in instead), happy to move it — flag it here.

---

.claude/hooks/session-start.sh
```diff
@@ -0,0 +1,24 @@
...
+#
+# Plain install (not --frozen-lockfile): this is a dev bootstrap, so if a session
+# adds or bumps a dependency the install should reconcile the lockfile rather
+# than fail the way a frozen/CI install would. --ignore-scripts installs
+# dependencies only, never running lifecycle scripts — it deliberately skips the
+# repo's own `prepare` (tsdown build) and any dependency install scripts, so a
+# session start never executes branch or package code (the dev/test loop runs
+# from source via tsx, and CI builds separately).
+corepack enable >/dev/null 2>&1 || true
+corepack pnpm install --ignore-scripts
```
dlecan: Question: I understand the security point but scripts executions are whitelisted+ new deps have a 7-day execution delay
What happens if a dependency needs an execution script for it's regular behaviour?

dlecan:
Correction to my earlier note about `--ignore-scripts` (`6ba0c86`): @dlecan is right, and I've dropped it.

`pnpm-workspace.yaml` already governs dependency lifecycle scripts — the `allowBuilds` whitelist (`esbuild`, `msw`) plus the `minimumReleaseAge` 7-day cooldown. `--ignore-scripts` would override that and skip whitelisted builds a package genuinely needs — notably `esbuild`'s native binary, which `tsx`/`tsdown` depend on — breaking the toolchain on a cold install (it only looked fine in my validation because node_modules was already warm). So the hook is now a plain `corepack pnpm install`, relying on the repo's existing, more precise supply-chain gating rather than a blunt flag.

---

docs/configuration.md
```diff
@@ -54,6 +57,42 @@ zendesk-mcp-server acme --transport http --port 8080 \
...
+file watcher, the reload happens only when *you* call `reload_tools` — at the
+end of an edit cycle, right before testing — which keeps the reload explicit and
+the notification noise to one burst per cycle. Point an MCP client's server
+command at the source and add the flag:
+
+```jsonc
+// .mcp.json
+{
+  "command": "pnpm",
+  "args": ["exec", "tsx", "src/index.ts", "--mode", "all", "--dev"],
```
dlecan, line 77: Question : according to this line, the current .MCP.json should have been updated with --dev flag, shouldn't it?

Issue #158, opened 2026-07-14
Study: which tools could benefit from MCP elicitation (without degrading multi-agent compatibility)


dlecan:
## Context

Spun off from the design discussion on #134 (section reordering). That feature needs a guard when an operation would reposition a large number of articles. We decided to ship #134 with a plain `confirm: boolean` guard (the existing `archive_article` pattern), and to study MCP **elicitation** separately rather than fold it into #134.

This issue tracks that study.

## Why not just use elicitation everywhere

Elicitation (`elicitation/create`) is a **server→client request**, negotiated at `initialize`: the client must advertise the `elicitation` capability. Several constraints follow, and they are exactly what makes this a study rather than a quick add:

1. **Multi-agent compatibility is an absolute rule** (see `CLAUDE.md`). Many mainstream MCP clients do not implement elicitation. It can therefore only ever be an *enhancement, capability-gated, with a graceful fallback* — never the sole mechanism. Any tool using it must still be fully operable when the client cannot elicit.
2. **Our primary caller is an LLM, not a human.** Elicitation opens a human-facing form. In an autonomous agent loop — and especially in headless contexts (cron, HTTP with no human attached) — an elicitation request may hang or be auto-declined. A model-settable `confirm` param is often both more compatible *and* semantically more appropriate.
3. **Transport dependency.** Elicitation needs a server→client channel. It works over stdio and StreamableHTTP, but our HTTP transport builds a per-session server; behaviour and client support differ across transports and must be verified on both.
4. **Surface coherence.** If we adopt elicitation, the guard pattern should stay uniform across destructive tools (`archive_article`, the #134 reorder guard, `update_ticket`/macro applies, etc.) rather than be one-off per tool.

## Goal

Produce a short design note (in `docs/`, or a decision recorded on this issue) answering:

- **Inventory** — which tools/operations have a UX where a human confirmation or a small structured input genuinely adds value (destructive/large-blast-radius writes, ambiguous disambiguation, missing-required-field prompts, …).
- **Fit test per candidate** — for each, is elicitation actually better than a model-settable param, given that the caller is usually the model? Reject candidates where a `confirm`/param is equal or better.
- **Compatibility contract** — the exact fallback behaviour when the client does not advertise `elicitation`: detect via `getClientCapabilities()?.elicitation`, fall back to the `confirm`/param path, never block. Define this once, reusably.
[...]

PR #159, opened 2026-07-14
feat(help_center): add reorder_article tool

### LINKED ISSUE #134: No section reordering helper: moving an article to the top requires manual, ambiguous reindexing
### Problem

Ordering articles within a section is only possible by setting each article's `position` one by one via the metadata update. There is no "move to top / bottom" or bulk-reorder helper. Two things compound this:

1. **Duplicate positions.** Newly created articles often land at `position: 0`, so several articles share `0`. Setting a target article to `0` then does **not** guarantee it appears first — ties are resolved in an undefined way. Achieving a deterministic order means manually reindexing the head of the section (`0, 1, 2, …`), which touches several otherwise-unrelated articles.
2. **Sort mode is invisible.** Whether a section is manually sorted or auto-sorted (by date, alphabetical, …) is not exposed. If it is auto-sorted, `position` is silently ignored and there is no way to tell from the API why a reorder "didn't take".

### Why it matters

"Put this article at the top of its section" is a common, simple editorial request. Today it is a multi-call, fragile operation with a real chance of no visible effect. We just had to reindex three articles by hand to move one to the top, precisely because two of them were tied at `position: 0`.

### Proposed solution

- Add a reorder / move helper — e.g. *move article to position N*, or *reorder a section from an ordered list of article IDs*, normalizing positions to be contiguous; **and/or**
- Expose the section's **sort mode** (manual vs automatic) so callers can detect up front when `position` will be ignored.

dlecan:
## Summary
Adds a `reorder_article` Help Center tool that moves an article within its section relative to its siblings — `top`, `bottom`, or `before`/`after` another article — and reports whether the new order actually took effect. This replaces the fragile, multi-call "set each `position` by hand" workaround: it writes the minimal set of positions needed to make the order deterministic even when several articles are tied at `position: 0`.

## Linked issue
Closes #134

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tooling / CI

## Author checklist
- [x] `pnpm test` passes locally (518 tests)
- [x] `pnpm typecheck` passes
- [x] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md`
- [x] If the change has runtime-observable behaviour: this PR carries a `## Functional validation plan` (authored with `/functional-validation-plan`), and it has been executed by an independent validator (`/run-validation-plan`) whose report is posted as a PR comment

Design decisions worth validating:
- **Gap-aware placement (default).** The tool sets an article's absolute `position` on a free integer between its two new neighbours, so `bottom` and roomy `before`/`after` cost **one write**; `top` is one write when slot 0 is free. It degrades to a bounded right-cascade only when neighbours leave no integer room, and to O(N) only on a section saturated with ties (the #134 case — a genuine lower bound). Trade-off: positions drift and leave gaps (invisible to end users, and cheap for later moves). `normalize: true` renumbers the section 0..N contiguously on demand.
- **No sort-mode field exists in Zendesk's API.** Auto-sorted sections (by date / alphabetical) silently ignore `position` writes (they return 200). Detection is best-effort: an up-front inversion probe avoids mass-writing a large auto-sorted section, and a definitive post-write re-read of the effective order catches every case, returning guidance to switch the section to Manual in Guide (no fabricated admin deep-link — none is stable across Guide versions).
- **Blast-radius guard** `ZENDESK_REORDER_CONFIRM_THRESHOLD` (default 20): refuses without `confirm: true` and reports the count.
- **Idempotency:** positions are absolute, so a re-run after a partial failure recomputes against current state and resumes without double-applying.
- Pure ordering logic is isolated in `src/utils/article-order.ts` (unit-tested directly); the handler in `src/tools/help-center.ts` does only I/O. Namespace/tool counts and doc tables were bumped/synced.
- Follow-up: MCP elicitation for the confirm guard is tracked separately in #158 (kept out of scope to preserve multi-agent compatibility).

### Context
The feature is the `reorder_article` MCP tool (namespace `help_center`). It is exercised **only** by calling `mcp__zendesk-local__reorder_article`; it reads a section's effective order via the articles list and writes `position` via the article-update endpoint. No other tool or transport exercises this code. `update_article`'s `section_id` (moving an article *between* sections) is a different path and out of scope here.

Everything numeric and every branch (gap-aware math, normalize, tie-breaking, threshold refusal, auto-sort detection, partial-failure) is already covered by unit tests (`tests/unit/utils/article-order.test.ts`, the `reorder_article` block in `tests/unit/tools/help-center.test.ts`) and an over-the-wire integration round-trip (`tests/integration/core-scenarios.ts`). This plan validates the behaviour against the **real Zendesk Help Center**, which the mocks cannot prove.

### Setup & prerequisites
The environment is already on the branch with the authenticated server running; drive it via the `mcp__zendesk-local__*` tools. Record the commit under test with `git rev-parse HEAD`.

**Pick a throwaway, manually-sorted section** (ideally a small test/sandbox section, ≥ 4 articles). Use `mcp__zendesk-local__list_sections` and `mcp__zendesk-local__list_articles` to choose one.

**Snapshot first, restore last (mandatory).** Before any scenario, capture the section's current order and each article's `position`:
`mcp__zendesk-local__list_articles { section_id: <S>, sort_by: "position", sort_order: "asc" }` — save the id→position map. After all scenarios, restore it (reorder or `update_article` per article, or one `reorder_article … normalize:true` pass followed by re-applying the saved order). Confirm the restored order matches the snapshot.

**Bounding real writes.** Prefer short-distance moves and a small section so each scenario is a handful of PUTs. Treat a hundreds-of-article section only as an optional documented stress check (measure the reported "N repositioned" count and the threshold refusal), not the default.

### Scenarios
| id | Validates | Manipulation & call | Expected observable |
| V1 | Move to top (real display) | On section S with a known non-first article A: `reorder_article { article_id: A, target: "top" }` | Success text `moved to top … section #S (N … repositioned)`. Re-list without sort → A is index 0. |
| V2 | Move to bottom, single write | `reorder_article { article_id: A, target: "bottom" }` | Success text says **1 article repositioned**; re-list → A is last. |
| V3 | before / after a reference | `reorder_article { article_id: A, target: "before", reference_article_id: B }` then `"after"` | Re-list → A immediately precedes / follows B. |
| V4 | Tie-break (#134) | Create 2–3 articles (they land at `position 0`) via `create_article`; move one to top | Success; re-list → the moved one is uniquely first; `list_articles sort_by=position` shows distinct positions (ties broken). |
| V5 | normalize | `reorder_article { article_id: A, target: "top", normalize: true }` | Success; `list_articles sort_by=position` shows contiguous 0,1,2,… |
| V6 | Threshold guard | On a section with > 20 articles (or lower `ZENDESK_REORDER_CONFIRM_THRESHOLD` is NOT available to the validator — use a large section), move a deep article to top **without** confirm | Refusal text `above the safety threshold of 20`; re-list → order **unchanged**. Then repeat with `confirm: true` → success. |
| V7 | Auto-sorted section (⚠) | In Guide, set a scratch section's "Order articles by" to **Creation date**; call `reorder_article … target:"top"` | Returns guidance containing `sorted automatically` and `Order articles by`; re-list → order unchanged. (Requires Guide UI access; mark BLOCKED if unavailable.) |
| V8 | Validation errors | `target:"before"` with no `reference_article_id`; `target:"top"` with a reference; reference == article; reference in another section | Each returns a clear error and makes **no** writes. |
| V9 | Idempotent re-run | Immediately re-run a V1/V3 call | Reports `already positioned … No changes made` (0 writes). |

### Long-running behaviours
None — the tool has no time-based behaviour. All timing-independent branches are covered by the unit suite; the validator only needs to confirm the real round-trip.

### Priorities
Do **V1, V2, V4, V6** first (the core user-facing paths: first/last, tie-break, the guard). V7 depends on Guide UI access. V3/V5/V8/V9 round out coverage.

### Reporting
Post a PR comment (English) with, per scenario id, an **OK / FAIL / BLOCKED** verdict and the observed evidence (the tool's returned text and the re-listed order). Confirm the tested commit SHA and that the section was restored to its snapshot. End with a green/failing summary.

---
https://claude.ai/code/session_01F82kn9YT3MaV2cVH7Ushyd

---
_Generated by [Claude Code](https://claude.ai/code/session_01F82kn9YT3MaV2cVH7Ushyd)_

## Summary by CodeRabbit

* **New Features**
  * Added a Help Center `reorder_article` tool to move articles (top/bottom/before/after) within a section, with deterministic tie-handling, optional normalization, no-op detection, and auto-sort awareness (with verification after writes).
  * Refuses large write sets unless confirmation is provided.
* **Documentation**
  * Documented `ZENDESK_REORDER_CONFIRM_THRESHOLD` (optional, default `20`) and updated the tools reference to include `reorder_article`.
* **Tests**
  * Expanded unit and integration tests for ordering calculations, validation, auto-sorted behavior, and mid-write failure recovery.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -160,6 +160,11 @@ Maximum number of images embedded as native image content in a single tool call.
 
 Hard cap on the number of comment pages fetched when collecting a ticket's attachments (raise it for tickets with very long comment threads).
 
+### `ZENDESK_REORDER_CONFIRM_THRESHOLD`
+**Required:** no · **Default:** `20`
+
+Safety threshold for `reorder_article`. When moving an article would rewrite more than this many article positions (for example moving an article to the top of a large or heavily tied section), the tool refuses and reports the count until the call is retried with `confirm: true`. Lower it to be prompted sooner, raise it to reorder large sections without confirmation.
+
 ---
 
 ← Back to the [README](../README.md).
--- a/docs/mcp-tools-reference.md
+++ b/docs/mcp-tools-reference.md
@@ -55,6 +55,7 @@ out every `write` tool before the proxies are built.
 | `compare_translations` | Section-level diff between two locales of an article | read |
 | `create_article` | Create a new article in a section | write |
 | `update_article` | Update article metadata (draft, labels, tags, visibility, section, sort position) | write |
+| `reorder_article` | Move an article within its section (top/bottom/before/after), breaking position ties deterministically | write |
 | `archive_article` | Archive (soft-delete) an article; recoverable only via the Guide admin UI | write |
 | `create_article_translation` | Create a translation for an article | write |
 | `update_article_translation` | Update an article's translation (full body) | write |
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -42,6 +42,13 @@ export const MAX_EMBEDDED_IMAGE_COUNT = positiveIntEnv('ZENDESK_MAX_EMBEDDED_IMA
 // Overridable via ZENDESK_MAX_COMMENT_PAGES for tickets with many comments.
 export const MAX_COMMENT_PAGES = positiveIntEnv('ZENDESK_MAX_COMMENT_PAGES', 10);
 
+// Blast-radius guard for reorder_article. Reordering an article can require
+// rewriting the `position` of several neighbours (to break ties or shift a run);
+// if that write set exceeds this many articles the tool refuses unless the caller
+// passes confirm:true, so a single "move to top" can't silently rewrite hundreds
+// of articles. Override via ZENDESK_REORDER_CONFIRM_THRESHOLD.
+export const REORDER_CONFIRM_THRESHOLD = positiveIntEnv('ZENDESK_REORDER_CONFIRM_THRESHOLD', 20);
+
 // Thresholds used to nudge callers toward section-scoped article tools
 // (get_article_outline / get_article_section / update_article_section)
 // instead of fetching/rewriting the full body.
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -6,12 +6,14 @@ export const TOKEN_CACHE_TTL_MS = 5 * 60 * 1000;
 // Read a positive-integer override from the environment, falling back to a safe
 // default. Unchecked Number() coercion is unsafe here: an empty string yields 0
 // and a typo yields NaN, either of which would silently break the guardrail that
-// relies on the value. Missing/empty/non-finite/non-positive values fall back.
+// relies on the value. Missing/empty/non-positive values and anything that is not
+// a positive safe integer (fractions like "1.5", values beyond 2^53) fall back —
+// these constants are all counts/sizes, so a fractional or unsafe value is a typo.
 const positiveIntEnv = (name: string, fallback: number): number => {
   const raw = process.env[name];
   if (raw === undefined || raw.trim() === '') return fallback;
   const parsed = Number(raw);
-  return Number.isFinite(parsed) && parsed > 0 ? parsed : fallback;
+  return Number.isSafeInteger(parsed) && parsed > 0 ? parsed : fallback;
 };
 
 // TTL for the per-session Help Center topology cache (zendesk-hc://topology).
--- a/src/tools/help-center.ts
+++ b/src/tools/help-center.ts
@@ -944,11 +944,12 @@ export const createHelpCenterTools = (ctx: ToolContext): ToolDefinition[] => {
           };
         }
 
-        // If the section provably ignores position (its effective order is not
-        // sorted by position) and the reorder would be large, refuse up front
-        // rather than fire many writes that will be silently ignored.
-        const autoSorted = hasPositionInversion(effective);
-        if (autoSorted && writes.length > REORDER_CONFIRM_THRESHOLD) {
+        // A strict inversion in the effective order is definitive proof the
+        // section ignores `position` (it is auto-sorted). Short-circuit before any
+        // write, at any size — writing would be silently ignored regardless. The
+        // post-write verification below still backstops the cases an inversion
+        // cannot reveal up front (e.g. an auto order that happens to be ascending).
+        if (hasPositionInversion(effective)) {
           return { content: [{ type: 'text', text: autoSortNotice(sectionId) }] };
         }
 
--- a/src/tools/help-center.ts
+++ b/src/tools/help-center.ts
@@ -104,14 +104,25 @@ export const createHelpCenterTools = (ctx: ToolContext): ToolDefinition[] => {
   // Fetch a section's articles in their EFFECTIVE display order (no sort_by), the
   // order an end user sees. Fully paginated so reorder decisions and verification
   // see every article, not just the first page. Used by reorder_article.
-  const fetchSectionOrder = async (sectionId: number, token: string): Promise<OrderedArticle[]> => {
+  //
+  // The listing MUST be locale-scoped: without sort_by the endpoint falls back to
+  // the section's configured "Order articles by", and a locale-dependent mode
+  // (title / recent activity / edited_at) makes the non-locale endpoint reject the
+  // request with HTTP 400 ("must specify a locale in order to sort by title…").
+  // Scoping by the article's locale both satisfies that requirement and still
+  // surfaces the auto-sorted order so the inversion probe can detect it.
+  const fetchSectionOrder = async (
+    sectionId: number,
+    locale: string,
+    token: string,
+  ): Promise<OrderedArticle[]> => {
     const order: OrderedArticle[] = [];
     let cursor: string | undefined;
     do {
       const response = await helpCenterGet<ZendeskListResponse<ZendeskArticle>>(
         subdomain,
         token,
-        `/sections/${sectionId}/articles`,
+        `/${locale}/sections/${sectionId}/articles`,
         buildCursorParams(MAX_PAGE_SIZE, cursor),
       );
       const articles = response.articles ?? [];
@@ -906,8 +917,12 @@ export const createHelpCenterTools = (ctx: ToolContext): ToolDefinition[] => {
           `/articles/${article_id}`,
         );
         const sectionId = article.section_id;
+        // Scope every section listing to the article's locale — the endpoint
+        // rejects a locale-less request when the section's default sort is
+        // locale-dependent (see fetchSectionOrder).
+        const locale = article.source_locale;
 
-        const effective = await fetchSectionOrder(sectionId, token);
+        const effective = await fetchSectionOrder(sectionId, locale, token);
 
         // For before/after, the reference must be in the same section. Disambiguate
         // not-found vs wrong-section so the caller gets an actionable message.
@@ -984,7 +999,7 @@ export const createHelpCenterTools = (ctx: ToolContext): ToolDefinition[] => {
         }
 
         // Verify the move took effect (the definitive auto-sort check).
-        const after = await fetchSectionOrder(sectionId, token);
+        const after = await fetchSectionOrder(sectionId, locale, token);
         if (!isPlacedAsRequested(after, article_id, target, reference_article_id)) {
           return { content: [{ type: 'text', text: autoSortNotice(sectionId, applied) }] };
         }
--- a/src/tools/help-center.ts
+++ b/src/tools/help-center.ts
@@ -13,6 +13,7 @@ import {
   LARGE_ARTICLE_BODY_CHARS,
   LARGE_ARTICLE_SECTION_COUNT,
   MAX_PAGE_SIZE,
+  REORDER_CONFIRM_THRESHOLD,
 } from '../constants';
 import type {
   ZendeskArticle,
@@ -26,6 +27,14 @@ import type {
   ZendeskTranslation,
   ZendeskUserSegment,
 } from '../types';
+import {
+  arrangeDesiredOrder,
+  computePositionWrites,
+  hasPositionInversion,
+  isPlacedAsRequested,
+  type OrderedArticle,
+  type ReorderTarget,
+} from '../utils/article-order';
 import {
   htmlToMarkdown,
   markdownToHtml,
@@ -76,8 +85,44 @@ const largeArticleHint = (body: string, sectionCount: number): string | null => 
   ].join('\n');
 };
 
+// Message returned when a reorder's position writes are (or would be) silently
+// ignored because the section is sorted automatically rather than manually. There
+// is no API field exposing the sort mode, so we name the section and the exact UI
+// steps rather than fabricate an admin deep-link (none is stable across Guide
+// versions). `applied` is set only after writes were attempted (post-verify path).
+const autoSortNotice = (sectionId: number, applied?: number): string =>
+  [
+    applied === undefined
+      ? `Section #${sectionId} looks like it is sorted automatically, so a manual reorder would have no visible effect.`
+      : `Wrote ${applied} article position(s), but the display order of section #${sectionId} did not change — the section is sorted automatically, so positions are ignored.`,
+    'To order its articles manually: in Guide, open the section, choose "Edit section", set "Order articles by" to Manual, then re-run this tool.',
+  ].join(' ');
+
 export const createHelpCenterTools = (ctx: ToolContext): ToolDefinition[] => {
   const { subdomain, getToken } = ctx;
+
+  // Fetch a section's articles in their EFFECTIVE display order (no sort_by), the
+  // order an end user sees. Fully paginated so reorder decisions and verification
+  // see every article, not just the first page. Used by reorder_article.
+  const fetchSectionOrder = async (sectionId: number, token: string): Promise<OrderedArticle[]> => {
+    const order: OrderedArticle[] = [];
+    let cursor: string | undefined;
+    do {
+      const response = await helpCenterGet<ZendeskListResponse<ZendeskArticle>>(
+        subdomain,
+        token,
+        `/sections/${sectionId}/articles`,
+        buildCursorParams(MAX_PAGE_SIZE, cursor),
+      );
+      const articles = response.articles ?? [];
+      for (const article of articles) {
+        order.push({ id: article.id, position: article.position });
+      }
+      const meta = extractPaginationMeta(response, articles.length);
+      cursor = meta.has_more && meta.after_cursor ? meta.after_cursor : undefined;
+    } while (cursor);
+    return order;
+  };
 
   return [
     {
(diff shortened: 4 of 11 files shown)


dlecan:
Self-review of the diff (high-effort pass) done before requesting review. Two correctness issues were found and fixed in `c5d794c`:

1. **`bottom` idempotency** — an already-last article was being bumped to `max+1` on every call; now anchored above the *other* articles' max, so it's a true no-op when already last.
2. **`before`/`after` verification** — was checking strict index adjacency, which a co-tied reference's arbitrary display tie-break could fail, yielding a spurious "section auto-sorted" warning on the exact tie case this tool targets. Now verified by *side* (moved is before/after the reference), which is the property we control.

Two low-severity items consciously left as-is (noted for the reviewer):
- The "already positioned" no-op path returns without a post-write re-read, so an auto-sorted section whose article already holds the target position value could report a false no-op. Adding a verification read to every no-op didn't seem worth the extra call on the common path.
- A draft or otherwise-unlisted article surfaces as "…is not in the section". Whether `/sections/{id}/articles` returns drafts is one of the empirical assumptions the functional validation plan (V-series) will confirm; the message can be refined if needed then.

---

dlecan:
Thanks @coderabbitai. Addressed in `cbc98ed`:

**Applied**
- **Fractional/unsafe threshold** (`src/constants.ts`): `positiveIntEnv` now requires `Number.isSafeInteger`, so `ZENDESK_REORDER_CONFIRM_THRESHOLD=1.5` (or an oversized value) falls back to the default instead of silently changing the guardrail. This hardens all four env-tunable constants, which are all counts/sizes. Added coverage in `tests/unit/constants.test.ts` (fractional, non-numeric, non-positive).
- **Untested `missing` path**: added a `reorder_article` test exercising the "reference article #N was not found" (404) branch, which left `seedSection`'s `missing` option uncovered.

**Skipped, with reasons**
- **Retry/backoff on the write loop + final read** (`help-center.ts`): intentionally not adding it here. There is no shared retry/backoff layer in the client (`executeRequest` surfaces 429s directly for every tool), so retry logic only in this one tool would be a special-case bandaid rather than a real fix. Read-after-write consistency is explicitly an assumption to confirm in the PR's functional-validation plan (item D) — if validation shows lag or throttling in practice, the right move is a shared retry in the API client, tracked separately.
- **Integration assertion** (`core-scenarios.ts`): the per-PUT write sequences and every outcome (moved / no-op / threshold / auto-sort / partial-failure) are asserted exhaustively in the unit tests via a stateful MSW section. The integration case deliberately mirrors the `archive_article` round-trip: a single over-the-wire dispatch check, not a behavioral re-test.

---

dlecan:
## Functional validation report — cbc98eddd23993df38f660a0cd2c2582983a665c

Independent execution of the `## Functional validation plan` for `reorder_article` against the **live fruggr Zendesk Help Center**. Real values (titles, section/category names) are redacted; numeric IDs are kept for reproduction only, never paired with the content they belong to.

### Headline
**The feature is non-functional against the real tenant.** Every reorder path — all four `target` values (`top`, `bottom`, `before`, `after`) — fails during the **read/listing phase, before any position write**, with the same Zendesk error:

```
Zendesk API error 400: Bad Request. {"error":{"title":"Invalid attribute","message":"You passed an invalid value for the sort_by attribute. The endpoint must specify a locale in order to sort by title, recent user activity, or edited_at"}}
```

The tool's internal article-listing call sends a `sort_by` (or uses an endpoint) that Zendesk requires a **locale** for on this multi-locale Help Center, and supplies none (the tool exposes no `locale` parameter). For contrast, a plain `list_articles { section_id: <S>, sort_by: "position", sort_order: "asc" }` **succeeds on the very same section without any locale** — so the working list shape exists; `reorder_article` isn't using it. This is exactly the class of defect the MSW mocks cannot surface (they don't enforce Zendesk's locale-for-sort rule), which is why the unit suite + integration round-trip pass while the live feature fails.

### Environment notes (context, not findings against this PR)
- Authenticated account: role `agent` (non-admin). Guide-management resources return Zendesk 403 `Permission denied`: the `zendesk-hc://topology` resource, `list_permission_groups`, `list_user_segments`. The maintainer is tracking this account/permission discrepancy as a **separate issue** — it is out of scope here, but it does block scenarios that need `create_article` (which requires a `permission_group_id`).
- The Claude Code auto-mode classifier blocks external writes to live articles not created this session. Where the maintainer explicitly authorized a named operation, the classifier cleared and the tool still failed at the listing 400 before any write — so **no write ever reached Zendesk**.
- **No state was mutated:** no article was created and no `position` changed. Verified by re-listing the most-probed section (25809226850973) after all scenarios — positions still `0, 13, 14, 19`, identical to the pre-run snapshot. Nothing to restore.

### Scenarios
| ID | Verdict | Evidence |
| -- | ------- | -------- |
| V1 | FAIL | `reorder_article { article_id: <already-first, pos 0>, target: "top" }` → Zendesk 400 `sort_by`/`locale` (see Headline). A no-op (article already at position 0) fails at the listing phase before any write. |
| V2 | FAIL | `reorder_article { article_id: <already-last, pos 19>, target: "bottom" }` → same Zendesk 400. A no-op (already last) fails identically at listing. |
| V3 | FAIL | `target: "after"` and `target: "before"` with a **valid same-section** `reference_article_id` (both articles in section 25809226850973) → same Zendesk 400 (reproduced on both). Fails at neighbour-listing before any write. |
| V4 | BLOCKED | Needs `create_article` (requires a `permission_group_id`; `list_permission_groups` → Zendesk 403) **and** a working reorder (which 400s). Not executable in this environment. |
| V5 | BLOCKED | Write path (auto-mode) and would hit the same listing 400 regardless. Not independently executable. |
| V6 | BLOCKED | Needs >20 repositions, i.e. 21+ tied articles — cannot create articles (permission 403). The threshold count is computed *after* the section listing, which 400s, so even the refusal path is unreachable here. |
| V7 | BLOCKED | Requires Guide UI to set a section to auto-sort; no Guide UI access in this session. |
| V8 | PARTIAL (3 OK / 1 FAIL) | Input-level validation (no API call) is clean: `before` w/o reference → `target "before" requires reference_article_id ...`; `top` with a reference → `reference_article_id must be omitted when target is "top" ...`; `reference_article_id == article_id` → `reference_article_id must differ from article_id.` **FAIL:** `reference_article_id` in *another* section leaks the same Zendesk 400 (`sort_by`/`locale`) instead of a clean "reference must be in the same section" error — the section listing 400s before the cross-section check runs. |
| V9 | FAIL | The "already positioned" (idempotent) path was exercised by the V1/V2 no-ops (article already first / already last). Expected `already positioned … No changes made`; observed the Zendesk 400 instead. Idempotency detection can't run because it lists the current order first. |

### Summary
**FAIL.** Core user-facing paths V1, V2, V3 fail; V9 (idempotency) fails; V8 is partial (the cross-section validation case leaks the raw 400). V4/V5/V6/V7 are BLOCKED by account permissions and lack of Guide UI. All failures trace to a **single reproducible defect**: `reorder_article`'s section-listing call sends a `sort_by` that Zendesk rejects without a `locale` on a multi-locale Help Center. A plain `list_articles` by `position` works on the same section, so the fix is likely to list the way the existing tool does (position sort) and/or pass the section's locale, then add a functional/live test exercising a real multi-locale section. Re-validate after the fix — none of the write scenarios could be genuinely exercised until the read phase succeeds.

dlecan:
Thanks for the thorough live validation — this caught exactly what the mocks can't. You're right on the root cause, and it's fixed.

**Root cause & fix (`aa1357c`).** `reorder_article` listed the section with `GET /sections/{id}/articles` **without a locale**. With no `sort_by`, Zendesk applies the section's configured *Order articles by*, and a locale-dependent mode (title / recent activity / edited_at) makes the locale-less endpoint 400 — which is why every read-dependent path failed before any write. The fix scopes both the effective-order read and the post-write verification to the article's `source_locale` (`GET /{locale}/sections/{id}/articles`). This satisfies Zendesk's locale requirement **and** still returns the section's effective (possibly auto-sorted) order, so the inversion probe and post-write auto-sort detection keep working.

This single fix should clear the read-phase failures:
- **V1, V2, V3, V9** — all traced to the listing 400; now unblocked.
- **V8** — the cross-section reference case now reaches the "must be in the same section" check instead of leaking the raw 400 (the listing no longer errors first).

Tests: the MSW handlers gained the locale-scoped listing path and a regression test now asserts the listing carries the locale, so a drop-the-locale regression fails CI.

**Re-validation request.** Could you re-run the plan against the current head (`e807709`)? V4–V7 stayed BLOCKED by the test account (no Guide permissions → can't `create_article`; no Guide UI for the auto-sort toggle) — if a Guide-permissioned account is available, V4/V6/V7 would round out the coverage; otherwise those remain environment-blocked, not code failures.

---

dlecan:
## Functional validation report — `19f13dc2d07fdce83aea15e5664ccda10bf818db`

**Tool under test:** `mcp__zendesk-local__reorder_article` (namespace `help_center`), driven directly against the live Zendesk tenant. All real article values (titles, bodies, client/org names) are **redacted**; only numeric article/section IDs are shown, per the validator redaction rule.

**Sections used** (internal, low-blast-radius sections; each snapshotted before and restored after):
- `#37276911319837` — **manually** sorted, 6 real articles → V1, V2, V3, V4, V5, V8, V9 (and V6 after temporary inflation)
- `#9816536711069` — 27 articles, **auto-sorted** (discovered during the run) → V6 pre-check + V7 (up-front detection)
- `#13064764773917` — 13 articles, **auto-sorted** (discovered during the run) → V7 (post-write detection)

**Environment note:** the Claude Code auto-mode permission classifier initially blocked writes to pre-existing (non-session-created) articles. The repo owner explicitly authorized those writes mid-run; all subsequent scenarios and the full restore proceeded under that authorization.

| ID | Verdict | Evidence |
| -- | ------- | -------- |
| V1 | OK | Manual section `#37276911319837`, `reorder {article:37284313803037, target:top}` → `Article #37284313803037 moved to top in section #37276911319837 (1 article repositioned).` Re-list (no sort) → moved article at index 0. Single write (slot 0 was free). |
| V2 | OK | `reorder {37284313803037, bottom}` → `... moved to bottom ... (1 article repositioned).` Re-list → moved article last. Exactly **one** write, as the plan expects. |
| V3 | OK | `after`: `reorder {37259296197789, after, ref 37260001893917}` → `... moved to after article #37260001893917 ... (2 articles repositioned).` `before`: `reorder {37259296197789, before, ref 37260001400989}` → `... moved to before article #37260001400989 ... (1 article repositioned).` Re-list confirmed the moved article's SIDE relative to each reference. |
| V4 | OK | Created 2 articles via `create_article` → each landed at `position 0` (3-way tie with the section's existing position-0 article — the #134 condition). `reorder {<created>, top}` → `... moved to top ... (7 articles repositioned).` Re-list `sort_by=position` → moved article **uniquely first**, positions now **distinct & contiguous** `0..7` (ties broken). |
| V5 | OK | `reorder {37259296197789, top, normalize:true}` → `... moved to top ... (5 articles repositioned).` Re-list `sort_by=position` → contiguous `0,1,2,3,4,5` (no gaps, no ties). |
| V6 | OK‡ | On a MANUAL section inflated to 24 articles with throwaway drafts. Without confirm: `reorder {<deep>, top}` → `Reordering article #... to top would reposition 23 articles in section #..., above the safety threshold of 20. Re-run with confirm: true to proceed.` (**no writes**, order unchanged). With `confirm:true` → `... moved to top ... (23 articles repositioned).` — count matches the refusal. |
| V7 | OK‡ | Validated over-the-wire on two **pre-existing** auto-sorted sections (no MCP tool exposes the Guide sort-mode, so the planned Guide-UI change was not possible). **Up-front path** (`#9816536711069`): `Section #9816536711069 looks like it is sorted automatically, so a manual reorder would have no visible effect. To order its articles manually: in Guide ... set "Order articles by" to Manual ...` (0 writes). **Post-write path** (`#13064764773917`): `Wrote 15 article position(s), but the display order of section #13064764773917 did not change — the section is sorted automatically, so positions are ignored. To order its articles manually: in Guide ... "Order articles by" to Manual ...`. Both contain `sorted automatically` + `Order articles by`; display order unchanged. |
| V8 | OK | All four return a clear error and make **no** writes: (1) `before` w/o ref → `target "before" requires reference_article_id (the article to move before).` (2) `top` w/ ref → `reference_article_id must be omitted when target is "top" (it only applies to "before"/"after").` (3) ref==article → `reference_article_id must differ from article_id.` (4) ref in another section → `Reference article #9879944233629 is in section #9816536711069, not section #37276911319837. It must be in the same section ...`. |
| V9 | OK | Immediately re-ran the V1 top move → `Article #37284313803037 is already positioned top in section #37276911319837. No changes made.` (0 writes — idempotent). |

### ‡ Deviations from the written plan (forced by the tenant's data)
Both large real sections turned out to be **auto-sorted**, and `list_articles` defaults to `sort_by=position` (so it does *not* reveal a section's configured sort — only the tool's own `fetchSectionOrder` does). Consequences:
- **V6:** an auto-sorted section short-circuits with the auto-sort notice *before* the threshold check runs, so a real >20 section here cannot exercise the guard. To test the guard on a genuinely manual section I temporarily created **18 throwaway draft articles** in manual section `#37276911319837` to push a to-top move above 20 writes, ran the refusal + confirm, then archived all 18 and restored the 6 real articles.
- **V7:** the plan called for switching a scratch section's "Order articles by" to Creation date in the Guide UI. No MCP tool exposes the section sort mode, so that manipulation was not performed. Instead the two pre-existing auto-sorted sections validated **both** detection code paths (up-front inversion probe *and* post-write verification) — broader coverage than the single planned case.

### Restoration (verified by re-list)
- `#37276911319837` → snapshot restored: positions `0,1,2,3,4,5` in original order.
- `#13064764773917` → snapshot restored: 6 articles at `position 0`, then `1..7`.
- `#9816536711069` → untouched (V6/V7 attempt made 0 writes; positions identical to snapshot).
- All **23** session-created articles (2 tie-break + 3 from an initial V4 attempt in the auto-sorted section + 18 threshold fillers) **archived** (soft-deleted). NB: archived articles remain in the Guide trash; permanent deletion requires the Guide admin UI (no API/tool exposes it).

## Summary
All 9 scenarios **OK** (V6 and V7 via the documented, data-forced deviations above). `reorder_article` behaves correctly over the wire against the real Help Center: single-write top / bottom / before / after on manual sections, #134 tie-breaking, `normalize`, the >20 blast-radius guard (refusal + `confirm` bypass with matching counts), auto-sort detection on both code paths with Guide guidance, no-op idempotency, and input validation. Every touched section was snapshotted and fully restored; all created test articles were archived.

dlecan:
Re-validation confirmed green against the tested head (`19f13dc`) — thanks for the thorough second pass. All 9 scenarios OK, and the report specifically exercises the two fixes the first round surfaced:
- the locale-scoped listing (V1–V9 all read successfully; V8's cross-section case now returns the clean "must be in the same section" error instead of the raw 400);
- the side-based before/after verification (V3).

The two deviations are sound and, if anything, widen coverage: V6 on a manual section inflated with throwaway drafts (the tenant's large sections are auto-sorted), and V7 on two pre-existing auto-sorted sections validating **both** detection paths (up-front inversion probe + post-write verification). Sections snapshotted/restored and test articles archived — no residue beyond the Guide trash (permanent deletion is UI-only).

From my side this closes the functional-validation loop for the feature. The PR is green on CI, clean to merge, and CodeRabbit had no remaining concerns — ready for human review / merge.

---

PR #173, opened 2026-07-20
feat: make the Help Center resource URI scheme configurable (param --hc-resource-scheme)

### LINKED ISSUE #169: Make the MCP resource URI scheme configurable (default `zendesk-hc://`)
## Problem

The resource URI scheme is hardcoded as `zendesk-hc://` — a single constant `TOPOLOGY_RESOURCE_URI = 'zendesk-hc://topology'` (`src/guidance/instructions.ts:4`), also baked into the `instructions` blob and the `README`. A deployer can't align it with their own Help Center.

The mismatch surfaces in practice: for the fruggr HC (public address `wiki.fruggr.io`), you'd intuitively expect a scheme like `wiki://` or `<brand>-wiki://`, not the generic `zendesk-hc://`.

Two things to keep straight, because they pull in opposite directions:

- **This is a general-purpose published package.** `@fruggr/zendesk-mcp-server` runs against *any* Zendesk subdomain (the `@fruggr` scope is misleading). So a hardcoded `fruggr-wiki://` is out — it wouldn't generalize and would break for every other deployer.
- **A URI scheme is a type/protocol identifier, not the public web address.** `wiki.fruggr.io` is where humans read articles; the MCP scheme is an opaque namespace for the MCP client, like `file://` or `git://` name a *kind* of resource rather than an instance. `zendesk-hc://` correctly says "this is a Zendesk Help Center resource" (RFC 3986: the scheme names the protocol, not the content). So the default is defensible — but being unable to brand it at all is the real gap.

## Proposed solution

Make the scheme **configurable**, defaulting to the current `zendesk-hc://` (no behavior change unless opted in):

- A CLI flag / env var, e.g. `--resource-scheme` / `ZENDESK_RESOURCE_SCHEME`, validated as an RFC 3986 scheme (lowercase, `[a-z][a-z0-9+.-]*`).
- The value flows into: the resource URI(s), the `instructions` blob that cites the URI literally, and stays in sync with the three touch-points (`instructions.ts`, `server.ts` registration, `README`).
- Keep it a single source of truth so any future resource (e.g. a per-article resource) inherits the same scheme.

This lets a deployer brand it (`wiki`, `<brand>-hc`, …) without imposing fruggr branding on a public package.

## Alternatives considered

- **Derive the scheme from the subdomain** (e.g. `fruggr-hc://` from subdomain `fruggr`). Gives per-instance uniqueness with zero config, but still doesn't match the public `wiki.fruggr.io` domain the expectation comes from, and it breaks fixed examples in the docs (the scheme now varies per deployment).
- **Keep `zendesk-hc://` and only document the rationale.** Zero new surface, closes the expectation gap in prose — but gives deployers no way to brand the scheme, which is the actual ask.

## Scope

- [x] Read-only (no data is modified in Zendesk)

## Additional context

- Touch-points today: `src/guidance/instructions.ts:4` (`TOPOLOGY_RESOURCE_URI`), `src/server.ts:303` (registration + `instructions` citation), `README.md` ("Help Center context (instructions + resources)").
- Guardrails: reject invalid schemes at config parse time (a bad scheme would surface as a broken resource URI); a too-generic scheme like `wiki://` risks collisions in multi-server clients, so document that trade-off rather than blocking it.
- Docs to sync if implemented: `docs/configuration.md` (new flag/env) and the `README` resource section.
- Follow-up context: split out of the discussion on #126.

dlecan:
## Summary
The Help Center topology resource URI scheme was hardcoded as `zendesk-hc://`. This PR adds a `--hc-resource-scheme` CLI flag / `HC_RESOURCE_SCHEME` env var (CLI wins) so a deployer can brand it (e.g. `wiki` → `wiki://topology`), defaulting to the current `zendesk-hc` — no behavior change unless opted in. `TOPOLOGY_RESOURCE_URI` becomes `topologyResourceUri(config)`, the single source of truth shared by the resource registration and the `instructions` blob.

## Linked issue
Closes #169

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tooling / CI

## Author checklist
- [x] `pnpm typecheck` passes
- [ ] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md` (n/a — no tool change)
- [x] If the change has **MCP-runtime-observable** behaviour (tool surface, transports, auth, resources, prompts, persistence, timing): this PR carries a `## Functional validation plan` (authored with `/functional-validation-plan`), executed by an independent validator (`/run-validation-plan`) whose report is posted as a PR comment. Pure tooling / dev-tooling is exempt — the standard gates cover it.

- **Strict validation, no normalization** (agreed in the issue discussion): the value must be a bare RFC 3986 scheme (`[a-z][a-z0-9+.-]*`). `wiki://`, `Wiki`, `-wiki`, `1wiki` are rejected at config parse time with an ASCII message that does not echo the value (same policy as `parsePort`). Note the issue's `-wiki` example is unattainable by design — RFC 3986 schemes must start with a letter.
- **WHATWG-special schemes are rejected** (`http`, `https`, `ws`, `wss`, `ftp`, `file`): their URL serialization appends a trailing slash, and the MCP SDK's read handler normalizes the requested URI through WHATWG `URL` before its exact-string registry lookup — the resource would be listed but never readable. The check validates the actual URI round-trips through `URL`, so it covers the whole class rather than a hand-kept denylist. Found by the Claude Code review pass, reproduced end-to-end.
- **No denylist for "generic" schemes** like `wiki`: MCP resource URIs are scoped per server, so cross-server collisions are not a functional concern; the trade-off is documented in `docs/configuration.md` instead.
- Empty env (`HC_RESOURCE_SCHEME=`) means unset (default applies), matching the `parsePortEnv` convention.
- `scripts/mcp-live.ts` `VALUE_FLAGS` was completed with the six value-taking flags it was already missing (pre-existing gap; its comment mandates sync with `parseCliArgs`).
- Known pre-existing parser trait, unchanged: a value-taking flag given as the last argument (or with an empty value) is silently ignored by the `&& next` guard — this applies to every flag (`--mode`, `--host`, …), not just the new one. Fixing it is a cross-cutting parser change out of scope here.

## Functional validation plan

**Context.** This PR makes the Help Center resource URI scheme configurable (`--hc-resource-scheme` / `HC_RESOURCE_SCHEME`, default `zendesk-hc`). Observable surface: the `instructions` blob sent on `initialize`, `resources/list`, `resources/read`. Tool calls do NOT exercise this change. Default behaviour must be identical to `main`.

**Setup.** The running server (default config) covers the default-scheme scenarios. A startup flag cannot be observed on an already-running instance: for custom-scheme scenarios use the sanctioned no-restart lever `pnpm mcp:live … -- --hc-resource-scheme wiki`, which spawns a short-lived instance (its `list` path is credential-free; `read` reuses `ZENDESK_OAUTH_TOKEN` if present — a missing token is a BLOCKED finding for the read step only, `list` still validates the surface). Capture `git rev-parse HEAD` in the report.

**Scenarios.**

| id | validates | manipulation | expected observable |
|----|-----------|--------------|---------------------|
| V1 | default unchanged — resource | none (running server) | `resources/list` contains `zendesk-hc://topology`; reading it returns the topology Markdown (locales, category tree, role) |
| V2 | default unchanged — instructions | none | server `instructions` cite exactly `zendesk-hc://topology` |
| V3 | custom scheme end-to-end | `pnpm mcp:live list -- --hc-resource-scheme wiki` (and `read wiki://topology` if a token is available) | listing shows `wiki://topology` and no `zendesk-hc://` URI; read (if run) returns the same topology content |
| V4 | env var + CLI precedence | `HC_RESOURCE_SCHEME=docs pnpm mcp:live list`; then same env + `-- --hc-resource-scheme wiki` | env alone → `docs://topology` listed; with the flag → `wiki://topology` (CLI wins) |
| V5 | strict rejection at startup | `pnpm mcp:live list -- --hc-resource-scheme 'wiki://'` (also try `Wiki`) | process exits at config parse with an error naming `HC_RESOURCE_SCHEME` / `--hc-resource-scheme`; no server starts |
| V6 | WHATWG-special scheme rejection | `pnpm mcp:live list -- --hc-resource-scheme http` | startup rejected with the WHATWG-special message (resource would be unreadable) |

**Long-running behaviours.** None (no timing involved). The full validation matrix (regex classes, env precedence, empty env) is covered by unit tests in `tests/unit/config.test.ts`; one rejected value per class is enough live.

**Priorities.** V1/V2 first (non-regression is the contract), then V3 (the feature), V4–V6 last.

**Reporting.** Post a PR comment (English): per-id OK/FAIL/BLOCKED verdict with the observed evidence (tool output / stderr excerpt) and the tested commit SHA.

## Summary by CodeRabbit

* **New Features**
  * Added support for customizing the Help Center resource URI scheme via `--hc-resource-scheme` or `HC_RESOURCE_SCHEME` (defaulting to `zendesk-hc`).
  * Help Center topology resources and generated instructions now use the configured scheme.
* **Documentation**
  * Documented the new configuration option, precedence rules (CLI over env), accepted bare RFC 3986 scheme formats, and examples.
* **Bug Fixes**
  * Improved CLI argument handling in the live MCP script to avoid misinterpreting additional flags.
* **Tests**
  * Expanded unit and integration coverage for defaults, overrides, env precedence, and invalid-scheme validation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,9 @@ disabled together with `--no-topology`):
 
 Clients that don't consume `instructions` or `resources` simply ignore them —
 the feature degrades silently. Use `--no-topology` to turn both off server-wide.
+The `zendesk-hc://` URI scheme is the default; a deployer can brand it with
+[`--hc-resource-scheme` / `HC_RESOURCE_SCHEME`](docs/configuration.md#hc_resource_scheme)
+(e.g. `wiki` → `wiki://topology`).
 
 ## Prerequisites
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -23,6 +23,10 @@ Options:
   --read-only             Only expose read operations
   --no-topology           Disable the Help Center structural context
                           (instructions + zendesk-hc://topology resource)
+  --hc-resource-scheme <scheme>
+                          URI scheme of the Help Center resources
+                          (default: zendesk-hc, i.e. zendesk-hc://topology);
+                          bare RFC 3986 scheme, e.g. "wiki" — no "://"
   --log-level <level>     debug | info (default) | warn | error
   --transport <t>         stdio (default) | http
   --host <host>           HTTP bind host (default: 0.0.0.0)
@@ -139,6 +143,19 @@ Public URL advertised in OAuth discovery metadata. See [Public URL](http-deploym
 **Required:** no · **Default:** —
 
 Comma-separated browser origins added to the default CORS allowlist.
+
+### `HC_RESOURCE_SCHEME`
+**Required:** no · **Default:** `zendesk-hc`
+
+URI scheme of the Help Center MCP resources (also `--hc-resource-scheme`, which
+takes precedence). With the default, the topology resource is
+`zendesk-hc://topology`; set e.g. `wiki` to expose it as `wiki://topology` and
+have the `instructions` blob cite that URI. Strictly a **bare RFC 3986 scheme**
+— a lowercase letter followed by lowercase letters, digits, `+`, `-` or `.`;
+anything else (`wiki://`, `Wiki`, `-wiki`, `1wiki`) is rejected at startup. A
+generic scheme like `wiki` is safe collision-wise (MCP resource URIs are scoped
+per server), it is just less self-descriptive in a client connected to several
+servers.
 
 ### `LOG_LEVEL`
 **Required:** no · **Default:** `info`
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -152,7 +152,10 @@ takes precedence). With the default, the topology resource is
 `zendesk-hc://topology`; set e.g. `wiki` to expose it as `wiki://topology` and
 have the `instructions` blob cite that URI. Strictly a **bare RFC 3986 scheme**
 — a lowercase letter followed by lowercase letters, digits, `+`, `-` or `.`;
-anything else (`wiki://`, `Wiki`, `-wiki`, `1wiki`) is rejected at startup. A
+anything else (`wiki://`, `Wiki`, `-wiki`, `1wiki`) is rejected at startup. The
+WHATWG-special schemes (`http`, `https`, `ws`, `wss`, `ftp`, `file`) are also
+rejected: URL normalization appends a trailing slash to them, which would make
+the advertised resource URI unreadable by MCP clients. A
 generic scheme like `wiki` is safe collision-wise (MCP resource URIs are scoped
 per server), it is just less self-descriptive in a client connected to several
 servers.
--- a/scripts/mcp-live.ts
+++ b/scripts/mcp-live.ts
@@ -54,6 +54,12 @@ const VALUE_FLAGS = new Set([
   '--tool',
   '--log-level',
   '--hc-resource-scheme',
+  '--transport',
+  '--host',
+  '--port',
+  '--public-url',
+  '--cors-origin',
+  '--callback-port',
 ]);
 const hasPositionalSubdomain = configArgs.some((arg, i) => {
   if (arg.startsWith('-')) return false;
--- a/scripts/mcp-live.ts
+++ b/scripts/mcp-live.ts
@@ -48,7 +48,13 @@ const fail = (message: string): never => {
 // Skip values that follow a value-taking flag (e.g. `all` in `--mode all`),
 // otherwise the placeholder is wrongly skipped. Keep in sync with the
 // value-taking flags in `src/config.ts` `parseCliArgs`.
-const VALUE_FLAGS = new Set(['--mode', '--namespace', '--tool', '--log-level']);
+const VALUE_FLAGS = new Set([
+  '--mode',
+  '--namespace',
+  '--tool',
+  '--log-level',
+  '--hc-resource-scheme',
+]);
 const hasPositionalSubdomain = configArgs.some((arg, i) => {
   if (arg.startsWith('-')) return false;
   const prev = configArgs[i - 1];
--- a/src/config.ts
+++ b/src/config.ts
@@ -44,6 +44,25 @@ export const ConfigSchema = z.object({
       message:
         'Invalid HC_RESOURCE_SCHEME / --hc-resource-scheme value. Expected a bare RFC 3986 scheme: a lowercase letter followed by lowercase letters, digits, "+", "-" or "." (no "://").',
     })
+    // WHATWG "special" schemes (http, https, ws, wss, ftp, file) serialize
+    // with a trailing slash (`new URL('http://x').toString()` === 'http://x/'),
+    // so the SDK's read handler — which normalizes the requested URI through
+    // `URL` before its exact-string registry lookup — would list the resource
+    // but never find it on read. Require the actual URI to round-trip.
+    .refine(
+      (scheme) => {
+        const uri = `${scheme}://topology`;
+        try {
+          return new URL(uri).toString() === uri;
+        } catch {
+          return false;
+        }
+      },
+      {
+        message:
+          'Invalid HC_RESOURCE_SCHEME / --hc-resource-scheme value. WHATWG-special schemes (http, https, ws, wss, ftp, file) do not survive URL normalization and would make the resource unreadable; pick a custom scheme such as "wiki".',
+      },
+    )
     .default('zendesk-hc'),
   /**
    * Dev-only (stdio): expose the `reload_tools` tool, which re-imports the tool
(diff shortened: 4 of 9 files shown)


dlecan:
@coderabbit review

dlecan:
## Functional validation report — 513b1ef4ded5f9874932cd7d83b710df554c440c

Independent validator run of the plan in the PR description. All tenant values below are redacted: the server under test points at a production Help Center.

| ID | Verdict | Evidence |
| -- | ------- | -------- |
| V1 | OK | `resources/list` on the running server returns exactly one resource: uri `zendesk-hc://topology`, name `help-center-topology`, mimeType `text/markdown`. `resources/read` on that uri returns the topology Markdown with the expected structure — `**Your access**: <name> (id <int>), role "<role>"`, `## Locales` (`Default: <locale>`, `Active: <locale list>`), `## Categories → sections` as `**<category>** (<id>)` with indented `**<section>** (<id>) — Category: <id>` children, `## Visibility (user segments)`, `## Permission groups`. |
| V2 | OK | The `instructions` blob this session received at `initialize` cites `zendesk-hc://topology`, and it is the only resource URI in the blob. |
| V3 | OK (content read BLOCKED) | `pnpm mcp:live list -- --hc-resource-scheme wiki` lists `wiki://topology` and no `zendesk-hc://` uri; tool list unchanged (3 proxies, `namespace` mode). Reading the content was not possible: `ZENDESK_OAUTH_TOKEN` is unset in this environment and `scripts/mcp-live.ts` reads only that variable (no token-store fallback), so the step is blocked, not failed. Routing was still verified without credentials: `read wiki://topology` under `--hc-resource-scheme wiki` fails with `-32603` carrying the script's own "Live calls need a Zendesk OAuth access token" message, i.e. the registered handler was reached and only the Zendesk call was missing; the control `read zendesk-hc://topology` under the same config fails with `-32602 Resource zendesk-hc://topology not found`. The new uri is registered and resolvable, the old one is gone. |
| V4 | OK | `HC_RESOURCE_SCHEME=docs pnpm mcp:live list` → `docs://topology`. Same env plus `-- --hc-resource-scheme wiki` → `wiki://topology`. CLI wins over env. |
| V5 | OK | `--hc-resource-scheme 'wiki://'` and `--hc-resource-scheme Wiki` both exit 1 with a `ZodError` raised from `loadConfig` (`src/config.ts:236`, called at `scripts/mcp-live.ts:73`, before `createMcpServer`), so no server starts. Both carry an `invalid_format` issue: "Invalid HC_RESOURCE_SCHEME / --hc-resource-scheme value. Expected a bare RFC 3986 scheme: a lowercase letter followed by lowercase letters, digits, \"+\", \"-\" or \".\" (no \"://\")." Neither message echoes the rejected value. |
| V6 | OK | `--hc-resource-scheme http` exits 1 at the same point with a single `custom` issue: "Invalid HC_RESOURCE_SCHEME / --hc-resource-scheme value. WHATWG-special schemes (http, https, ws, wss, ftp, file) do not survive URL normalization and would make the resource unreadable; pick a custom scheme such as \"wiki\"." |

## Summary

Green. Default behaviour is unchanged (V1, V2), the custom scheme works end to end through both `resources/list` and resource resolution (V3), CLI beats env (V4), and both rejection classes fail at config parse before the server boots (V5, V6).

One step could not be run: reading the topology *content* under a custom scheme, because no OAuth token is exposed to `scripts/mcp-live.ts` here. The remaining risk is small — V1 already proves the content path on the default scheme, and the content does not depend on the scheme.

One non-blocking observation: for values rejected on format (`wiki://`, `Wiki`), the `ZodError` reports both issues, so the WHATWG-special message shows up alongside the format one even though the value is not a special scheme. The refinement itself is right (`Wiki://topology` does not round-trip through `URL`), only the wording is misleading if a reader picks the last message. Cosmetic.

dlecan:
Validation loop closed. I re-checked the report against the plan's expected observables: V1-V6 all match, tested at the right SHA (`513b1ef`). The V3 content-read BLOCKED is per plan (no `ZENDESK_OAUTH_TOKEN` in the validator env) and the routing evidence actually exceeds the plan's ask — `read wiki://topology` reaching the handler (`-32603` token error) while `read zendesk-hc://topology` returns `-32602 not found` proves registration/resolution of the new URI and removal of the old one; the content path itself is covered by V1 and is scheme-independent.

The non-blocking observation is addressed in ca4c20a: string checks now abort on the RFC 3986 regex failure, so a format-rejected value (`Wiki`, `wiki://`) reports only the format message and the WHATWG-special message only appears for values that are actually special schemes (unit-tested). Message texts are unchanged, so no validation re-run is needed — only the issue count for V5-class values changes.

---

dlecan:
### Follow-up on the validation report — same commit `513b1ef4ded5f9874932cd7d83b710df554c440c`

The report above left two gaps. Both are now closed, so nothing in it needs re-running.

Setup for this pass: `--hc-resource-scheme wiki` added to the `zendesk-local` entry in `.mcp.json`, then the client session restarted. That exercises the real startup path (`src/index.ts` over stdio), not the `mcp:live` in-memory harness, and the session carries a real token — which is what unblocked the read.

**V3 read, previously BLOCKED.** `resources/list` returned `wiki://topology` as the only resource, and reading it returned the topology Markdown with the same sections and entries as the default-scheme read in V1 (`## Locales`, `## Categories → sections`, `## Visibility (user segments)`, `## Permission groups`, the access line). Content is unaffected by the scheme. Verdict: OK.

**Instructions under a custom scheme, previously an inference from the code.** Now observed. Filtering the `instructions` blob to lines containing a URI, via an MCP client against `createMcpServer`:

```
--- args: [] ---
When creating or editing Help Center content, the resource zendesk-hc://topology is useful context:
--- args: ["--hc-resource-scheme","wiki"] ---
When creating or editing Help Center content, the resource wiki://topology is useful context:
```

One URI-bearing line in the blob, and it follows the config. The resource registration and the instructions stay in sync, which was the point of routing both through `topologyResourceUri(config)`.

No new findings. `.mcp.json` has been restored to its committed state.

PR #186, opened 2026-07-26
chore(lint): lint on edit, format at pre-commit

dlecan:
## Summary

The `PostToolUse` hook ran `biome check --write` on every edit. The cause is not the formatter and not Biome: two `types`-domain rules in `biome.json` (`useArrayFind`, `useArraySortCompare`) force a project-wide type-inference pass on *every* invocation, single file included — 2 rules out of 221 active, costing more than the other 219 combined. This splits the two roles by stage so the hot path can skip that domain while pre-commit and CI still enforce it.

This started as an evaluation of the Oxc toolchain (oxlint + oxfmt) as a replacement. **Oxc is not adopted** — the study is archived as a decision record.

## Linked issue

None

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Tooling / CI

## What changes

| Stage | Command | Median |
| --- | --- | ---: |
| `PostToolUse` (per edit) | `biome lint --write --skip=types <file>` | **137.0 ms** |
| pre-commit (staged) | `biome check --write --error-on-warnings` via lefthook | 503.5 ms |
| CI | `pnpm check` — **unchanged** | 807.2 ms |

Measured with `scripts/bench-lint-tooling.mjs` (added here), 13 runs, 2 warm-ups discarded, x86-64 container, on the **Biome 2.5.5** this branch now merges:

```text
1 edit  biome check --write  (the old hook)   458.5ms
1 edit  biome lint           (no skip)        433.9ms
1 edit  biome lint --write --skip=types       137.0ms   <- new hook
ci      biome check --error-on-warnings       807.2ms
ci      biome check --skip=types              404.8ms   <- cost of the 2 type rules
```

**3.3× on the hook.** Note that lint-only *on its own* buys ~5 %; the speedup is `--skip=types`, and the stage split is what makes skipping it free.

**Nothing is lost.** The two rules are skipped on the hot path by a flag, not removed — pre-commit and CI still enforce them. Both currently report zero findings anyway, including at the repo's two `.sort()` sites, where Biome's inference cannot resolve the receiver type.

## Notes for the reviewer

**Why lefthook.** It replaces a first pass built on husky + lint-staged: husky's last release is 2025-01-11 (18 months), and lefthook covers both jobs in one dependency instead of two. Its `postinstall` stays **blocked** in `allowBuilds` — the Go binary resolves from its platform optional dependency, verified working with `ERR_PNPM_IGNORED_BUILDS` reported.

**Two traps worth a second look in the diff.**
- `{src,tests,scripts}/**/*.{ts,…}` alone silently skips top-level files — lefthook's matcher needs at least one directory level for `**/`, so `src/index.ts`, `src/config.ts` and `tests/setup.ts` went unchecked. `lefthook.yml` carries a second single-level pattern for that reason. Caught by testing, not by reading.
- `pnpm add -D lefthook` writes a `lefthook: set this to true or false` placeholder into `allowBuilds` that breaks every later `pnpm` command until resolved.

**Partial staging is safe** — I documented the opposite first, then measured it. lefthook's docs describe `stage_fixed` as a plain `git add` and list no stash option, but it does hide unstaged hunks: a staged edit is committed formatted while a separate unstaged edit survives untouched. Parity with lint-staged on the property that justified lint-staged.

**The Biome version mattered, and has since resolved itself.** While this PR was in review the repo was pinned to 2.5.4, where the change made the hook **12×** faster (1348.6 → 112.3 ms). Renovate has since shipped 2.5.5 (#190) and this branch merged it; 2.5.5 is ~3.8× faster on the type-inference path on its own, so the remaining gain from this PR is 3.3× rather than 12×. The figures above and in §1/§6/§7 of the record were re-measured on 2.5.5. The per-edit hook is unaffected by the version either way — `--skip=types` already sidesteps that path.

**Why not Oxc.** Against `biome lint --skip=types`, oxlint measured in the same range on the same machine. Adopting it would buy a second toolchain, a second rule config and two drifting rule sets for no speed gain and 8 lost rules; a parity oxlint config produced 83 diagnostics on a tree Biome calls clean, 8 surviving full tuning. Rule coverage was derived mechanically — 709 isolated `biome migrate eslint` runs cross-referenced against oxlint's 738-rule inventory, then verified by replaying each Biome rule's own `expect_diagnostic` example through both linters (124/140 applicable rules have a full equivalent, 88.6 %). oxfmt matches Biome's output byte-for-byte on 68 of 70 TypeScript files but is slower at every scope and cannot sort imports. All of it is in `docs/decisions/lint-tooling.md`, a new build-doc directory.

## Author checklist
- [x] If this PR resolves an issue, its description above links it with a closing keyword (`Closes #`) so it auto-closes on merge to `main` — n/a, no issue
- [x] `pnpm test` passes locally (31 files, 546 tests)
- [ ] `pnpm test:coverage` meets the thresholds — not re-run; no `src/` change in this PR
- [x] `pnpm typecheck` passes
- [x] Documentation is updated where needed (`AGENTS.md`, `CONTRIBUTING.md`, `docs/decisions/`)
- [x] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md` — n/a
- [x] **Pure dev-tooling change, exempt from the functional validation plan.** Nothing MCP-runtime-observable changes: no tool surface, transport, auth, resource, prompt, persistence or timing difference for a client. The standard gates cover it.


AGENTS.md
```diff
@@ -108,6 +108,15 @@ test: if a running server behaves no differently for a client, it's tooling.
   from the *scan*, not the *lint* — diagnostics here are unchanged. Revisit if a
   type-aware rule needs dependency types. `biome.json` rejects comments, hence
   this note.
+- Keep `--skip=types` on the `PostToolUse` hook. The two `types`-domain rules in
```
dlecan: Nitpick: keep the agent context small so shorten this! The "why" is already explained in the dedicated document

CONTRIBUTING.md
```diff
@@ -52,6 +52,34 @@ pnpm dev -- <your-subdomain> --transport http --port 3000 --public-url http://lo
 pnpm build && pnpm typecheck && pnpm check && pnpm test
 ```
 
+### Lint and format: which stage does what
```
dlecan: Nitpick: too long

docs/decisions/README.md
```diff
@@ -0,0 +1,16 @@
+# Decision records
```
dlecan: Nitpick: remove this
This is an internal decision

CONTRIBUTING.md
```diff
@@ -52,6 +52,22 @@ pnpm dev -- <your-subdomain> --transport http --port 3000 --public-url http://lo
...
```
dlecan: Nitpick: this file is about HOW to CONTRIBUTE  (process) and NOT a documentation about HOW the project and scripts work
Please keep this file about the contributing process only

PR #188, opened 2026-07-27
chore(lint): widen the Biome rule set, and clear what stood in the way

dlecan:
## Summary

`recommended` left roughly half of Biome's JS/TS rules off. Every inactive rule was measured individually against `src/ tests/ scripts/`; the ones the tree was already clean for went on, the ones worth having but blocked by real debt got the debt paid first, and every remaining rule got a written reason. **85 rules are now enabled explicitly**, up from 8.

> **Scope note:** this started as the config-only change the title used to describe (43 zero-violation rules). It grew, at the reviewer's request, into paying down the debt that blocked the rest — so it now also carries the refactors and a decision record. The commits are ordered so each step is reviewable on its own.

## Linked issue
None

## Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Documentation
- [x] Tooling / CI

## Author checklist
- [x] `pnpm typecheck` passes
- [ ] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md`
- [x] If the change has **MCP-runtime-observable** behaviour: pure tooling and behaviour-preserving refactors — the standard gates cover it.

## What changed, in commit order

**1 — Rules the tree was already clean for.** No product code touched. The ones that carry weight: `noUndeclaredDependencies` (phantom deps before they reach the published tarball), `noImportCycles`, `noPrivateImports`, `noGlobalDirnameFilename` and `noCommonJs` (the package is ESM-only), `noConstantBinaryExpressions`, `noEvolvingTypes`, `noUnusedExpressions`, `useErrorMessage` / `useThrowOnlyError`, and `noSkippedTests` / `noExportsInTest` so a `.skip` can't land unnoticed.

**2 — Tests first, alone.** `tests/unit/client/zendesk-api.test.ts` asserted on rejections inside `catch` blocks. Those guard the error *message*, but not the throwing: make a 401 resolve instead of reject and the old suite stayed green at 11/11. Rewritten to `rejects.toThrow(…)` and a captured rejection. Also an `expect()` in a helper outside `it()`, and a dynamic `import()` shadowing a static one. **No `src/` file in that commit** — the suite still measured the same code it did before.

**3 — Small real fixes**, then the mechanical consistency pass.

**4 — The complexity debt.** Eleven functions over the cognitive-complexity threshold, worst at 48, all brought under it with the tests untouched as the safety net:

| Function | Was | How |
| --- | ---: | --- |
| `parseCliArgs` | 48 | 15-arm if/else chain → two lookup tables |
| `reorder_article` | 38 | validation, reference check, write loop extracted |
| `compare_translations` | 28 | four pure functions, one per report signal |
| `formatMacroPreviewDiff` | 28 | the two diff passes split |
| `requestListener` | 26 | static-route table + error exit |
| `handleMcpRequest`, OAuth callback | 21 | session dispatch / decision-vs-plumbing |
| 5 others | 16–18 | targeted extractions |

**5 — The decision record.** `docs/decisions/biome-rules.md`: the selection rule, what is on, and **every not-enabled JS/TS rule with a one-line reason and its measured count**. `AGENTS.md` keeps four lines and a pointer.

## Two things worth a reviewer's attention

**A regression this PR introduced and then fixed.** The table-driven CLI parser indexed plain object literals with raw `argv`, so inherited keys resolved — `zendesk-mcp-server toString` hit `Object.prototype.toString` and was swallowed instead of read as the subdomain. Now `Map`s. The regression test fails against the object-literal version.

**Rules tried and rejected on evidence, not taste.** `useExplicitLengthCheck`'s autofix rewrites `x?.length &&` to `x?.length > 0 &&`, which does not type-check, and the hand-written form loses the narrowing — 11 type errors for a cosmetic gain. `noExcessiveLinesPerFunction` would fight the documented `ToolDefinition[]` architecture, where line count measures how many tools exist. `noUnresolvedImports` is 31 false positives (Biome can't read the export maps of `msw` and the MCP SDK) — down from 110 on 2.5.4, so worth re-checking on future bumps.

- **Method.** Each rule enabled *alone*, diagnostics counted. Two measurement traps worth knowing: a config placed outside the repo silently breaks every multi-file rule (it made `noUndeclaredDependencies` report 85 phantom hits), and `domains: { project: "all" }` enables the whole domain, not just the rules you list — naming them explicitly works and is what `biome.json` does.
- **Every enabled rule was verified to actually engage** with deliberate-violation probes, rather than being accepted and silently ignored.
- **The tool surface is untouched.** Mechanically verified across the range: no `describe(`, `description:`, `title:` or `inputSchema` line added or removed under `src/tools/`. The multi-agent superset rule holds.
- **`types`-domain rules stay capped at the two already paid for.** Biome 2.5's inference misses even the documented examples of `noFloatingPromises`, and `noUnnecessaryConditions` false-positives on optional chaining. Zero diagnostics there means "not analysed", not "clean".
- Gates on the merged head: `pnpm check` clean, `pnpm typecheck` clean, 593 tests pass.


AGENTS.md
```diff
@@ -108,6 +108,16 @@ test: if a running server behaves no differently for a client, it's tooling.
   from the *scan*, not the *lint* — diagnostics here are unchanged. Revisit if a
   type-aware rule needs dependency types. `biome.json` rejects comments, hence
   this note.
+- The rule set beyond `recommended` is a **zero-violation ratchet**: a rule is
```
dlecan: Nitpick: unnecessary here so remove the paragraph

AGENTS.md
```diff
@@ -116,6 +116,14 @@ test: if a running server behaves no differently for a client, it's tooling.
   automated callers go through `scripts/biome.mjs`, never
   `node_modules/.bin/biome`, which has no binary to run on Android/Termux.
   Why, and what the shim does: `docs/decisions/biome-on-android.md`.
+- `nursery` rules stay off: they change semantics between Biome minors and
```
dlecan: Nitpick: all this new chapter is not necessary in the agents.md file

biome.json
dlecan, line 1: Suggestion: add also all rules not enabled, explaining why they are disabled (no more than 1 one line for each)

biome.json
```diff
@@ -10,20 +10,98 @@
...
+        "useGuardForIn": "error",
+        "useNumberToFixedDigitsArgument": "error",
+        "useStaticResponseMethods": "error"
       },
       "style": {
         "useErrorCause": "error",
-        "useGlobalThis": "error"
+        "useGlobalThis": "error",
+        "useGroupedAccessorPairs": "error",
+        "noCommonJs": "error",
```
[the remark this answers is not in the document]
dlecan: Not reproducible — no override needed.

`noCommonJs` does not fire on `scripts/biome.mjs`. Verified against the current branch:

```
$ node scripts/biome.mjs lint --only=style/noCommonJs src/ tests/ scripts/
Checked 81 files. No fixes applied.
```

The reason is the binding, not the file extension. `scripts/biome.mjs:39` declares `const require = createRequire(import.meta.url)`, so every `require(...)` in that file resolves to a local ESM-created function. `noCommonJs` targets the CommonJS *globals*; a locally-bound identifier shadows them, so the rule correctly stays quiet. The `.cjs`/`.cts` auto-disable the analysis cites is a separate mechanism and isn't what applies here.

`pnpm check` covers `scripts/` (`biome check --error-on-warnings src/ tests/ scripts/`) and is green on this branch, which is the same signal from the other direction.

Switching to dynamic `import()` isn't an option regardless: the launcher load at line 176 is synchronous by design — see `docs/decisions/biome-on-android.md`.

---

biome.json
dlecan: Done, but in `docs/decisions/biome-rules.md` rather than in `biome.json` — flagging the deviation since you asked for it here.

`biome.json` rejects comments, so a one-line reason per rule can't live in the config itself. Putting it in a decision record also settles your other comment: the rationale leaves `AGENTS.md`, which now keeps four lines and a pointer. That's what `AGENTS.md` asks for anyway — *"when a section grows into exhaustive detail, it's a deep dive (→ `docs/`): link it, don't inline it"* — and `docs/decisions/` already holds the two other lint decisions.

The record covers every not-enabled JS/TS rule:

- **Off by policy** — the whole `nursery` group, and the `types` domain beyond the two rules the hook already pays the inference pass for.
- **Not applicable** — the 68 framework rules as one line (no JSX, no DOM, no ORM), then a line each for the a11y/browser/deny-list-driven ones that can never fire here.
- **Evaluated and rejected** — one line each, with the measured diagnostic count: `useNamingConvention` (992, snake_case Zendesk fields), `noDelete` (13, the alternative writes the string `"undefined"`), `noAwaitInLoops` (12, the pagination loops are sequential by necessity), and so on.
- **Nursery rules the tree is already clean for** — listed so a future promotion can be picked up with no code change.

Counts are measured, not estimated: each rule enabled alone against `src/ tests/ scripts/`. I re-ran the whole sweep after merging `main` so the numbers match the current branch.

---

PR #191, opened 2026-07-29
chore(lint): run Biome through a shim so it works on Android/Termux

dlecan:
## Problem

Biome does not run in native Termux, where `process.platform` is `android`. Two
things follow from that. Biome's launcher maps the platform onto one of its
`@biomejs/cli-*` packages and has no android entry, so it exits 1 with "doesn't
ship with prebuilt binaries for your platform yet". And there would be nothing
to point it at anyway: pnpm skips all eight of those packages, because each
carries its own `os`/`cpu`/`libc` gate.

So `pnpm check` failed on the spot, and the `PostToolUse` hook, which ends in
`|| true`, failed silently on every edit.

The glibc `linux-arm64` build cannot run there either (`cannot execute: required
file not found`). The `linux-*-musl` builds are statically linked and run as-is
on bionic.

## What this does

`scripts/biome.mjs` hands off to Biome's own launcher on every platform except
android, in-process, so no second Node start-up. On android it fetches the musl
build matching the installed Biome version, caches it per version under
`XDG_CACHE_HOME`, and execs it. The version is read from the manifest, so a
Renovate bump fetches its own binary with no coupling to maintain.

Four callers go through the shim: the `check` and `check:fix` scripts, the
`PostToolUse` hook, the pre-commit job. `scripts/bench-lint-tooling.mjs` asks it
for the binary path (`--print-binary`) instead, so its samples exclude the
wrapper's start-up. That script targets Android/Termux and could not run there
at all before this.

Other platforms are untouched: no lockfile change, no extra download, same
launcher behaviour.

## Still broken on Termux, deliberately

A direct `node_modules/.bin/biome` from outside the repo (an IDE extension, an
ad-hoc `pnpm exec biome`) still fails. The escape hatch is
`export BIOME_BINARY=$(node scripts/biome.mjs --print-binary)`. And lefthook has
no android build either, so the pre-commit hook is not installed on that device;
repointing `lefthook.yml` serves the other platforms and will take effect here
if lefthook becomes runnable.

## Validation

Tooling only, so the standard gates are the validation: nothing a client
observes at runtime changes.

| check | result |
| --- | --- |
| `pnpm check` | 81 files, 13s (failed outright before) |
| exit-code fidelity | a lint violation exits 1 on both branches |
| `pnpm check:fix` | rewrites a mis-formatted file |
| `PostToolUse` command as configured | passes, 0.31s |
| pre-commit command as configured | passes on staged paths |
| cold fetch, throwaway cache | one log line, staging cleaned up |
| fetch failure (npm unreachable) | actionable message, exit 1, no leftovers |
| `pnpm typecheck` | passes (TS 6 fallback, expected on android) |
| `pnpm test` | 546 tests pass |

What I cannot verify on this device is the non-android branch. CI's lint job is
the proof for that one.

Rejected alternatives, and what each would cost everyone else:
[`docs/decisions/biome-on-android.md`](docs/decisions/biome-on-android.md).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Biome linting and formatting support on Android and Termux environments.
  * Updated automated checks, pre-commit hooks, and tooling to use a platform-compatible Biome launcher.
  * Biome binaries are now retrieved and cached automatically when required.

* **Documentation**
  * Added guidance for running project checks and documented the Android/Termux Biome decision and operational considerations.


dlecan:
@coderabbitai review

PR #196, opened 2026-08-02
chore(build): trim build dependencies and drop the empty d.ts

dlecan:
## Summary

Resource-trimming chores on the build toolchain. The build stops producing two artefacts nobody consumes — an empty declaration file and a source map nothing can read — and the hand-built yuku android-arm64 bindings grafted in #189 are replaced by the ones upstream now publishes, which also removes the last exotic subdependencies in the tree.

| | before | after |
| --- | --- | --- |
| `pnpm build` | ~1518 ms cold / ~445 ms warm | **~78 ms** |
| `dist/` | 633 kB, 3 files | **202 kB, 1 file** |
| npm tarball | 173.6 kB | **64.8 kB** |

## Linked issue

None

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Tooling / CI

## 1. Stop emitting an empty declaration file

`src/index.ts` exports nothing — it imports, defines `main()`, calls it. So `dist/index.d.ts` was:

```
export {}
```

Ten bytes typing an API that does not exist. The package is a binary (`bin.zendesk-mcp-server`; the smoke test spawns `node dist/index.js`), never imported as a library, and `"types"` in `package.json` advertised that non-existent API to consumers. Both are gone.

Emitting it was not free: it ran tsgo on every build, which is the only reason `rolldown-plugin-dts` loaded yuku, and the only reason `tsdown.config.ts` carried a `process.platform !== 'android'` special case — Microsoft publishes no `@typescript/typescript-android-arm64`. **That gate is now removed outright**, so the Android build stops being a special case rather than staying a documented exception.

Measured on x86, 3 runs each, tsdown's own timer: **~445 ms with dts, ~120 ms without.**

## 2. Stop generating the source map

A map earns its cost when something debugs the bundle. Nothing here does:

- `pnpm dev` runs tsx on `src/` and vitest runs against `src/` — routine work never touches `dist/`;
- in the published binary no code path reads `.stack`; all 14 error sites format `err.message`;
- the one stack that can print, from the fatal handler in `src/index.ts`, would need `--enable-source-maps`, which nothing sets — not the package, not its scripts, not the `#!/usr/bin/env node` shebang.

Measured, 3 runs each: **~143 ms with the map, ~67 ms without** — over half the remaining build time, for 431 kB on disk against a 202 kB bundle.

Turning it off also drops the trailing `//# sourceMappingURL=index.js.map` from the bundle, so there is no dangling reference for npm consumers to trip on. Reverting is one line — `sourcemap: true` gives a debuggable bundle back for a 143 ms build.

## 3. tsconfig cleanup

`declarationMap` and `sourceMap` both leave `tsconfig.json`. The typecheck runs `--noEmit`, so neither ever wrote anything, and what `dist/` contained came from tsdown's own flags — verified by rebuilding without them.

**`declaration` stays**, deliberately: under `--noEmit` it still makes the compiler report declaration-emit diagnostics (TS4094 and friends — verified on a synthetic case), guarding type portability independently of whether files are written.

## 4. Use upstream yuku android-arm64 bindings

```
package.json  tsdown ^0.22.0
  └─ tsdown 0.22.14        →  rolldown-plugin-dts ^0.27.13
       └─ rolldown-plugin-dts 0.27.13  →  yuku-{ast,parser,codegen} ^0.7.0
            └─ locked at 0.7.4  →  no android-arm64 binding
```

yuku 0.8.3 is the first release declaring `@yuku-{codegen,parser}/binding-android-arm64` in its `optionalDependencies`. yuku is `0.x`, so `^0.7.0` cannot resolve `0.8.x` — an override on yuku alone would contradict the declared range, and the plugin has to move. `rolldown-plugin-dts` 0.27.14 (2026-07-24) is the first version depending on yuku `^0.8.0`, and it falls inside tsdown 0.22.14's `^0.27.13`, with identical `engines` and `peerDependencies`. So this is a lockfile refresh with **no `package.json` dependency edit**. (0.28.0 would pull tsdown 0.23.0-beta.2 — not needed.)

Lockfile moves: `rolldown-plugin-dts` 0.27.13 → 0.27.14, `yuku-{ast,parser,codegen}` and `@yuku-toolchain/types` 0.7.4 → 0.8.3, bindings resolved from npm with registry integrity behind the standard `os`/`cpu` gate.

**The graft was already inert** — keyed on `yuku-codegen@0.6.5` / `yuku-parser@0.6.5`, it stopped matching when the tree moved to 0.7.4. Removing it drops the only exotic subdependencies in the tree, so **`blockExoticSubdeps` returns to its default**.

Scope note, since #189's comment overstated it: the missing binding did not by itself break `pnpm run build` on Android, because `dts` was already off there and tsdown only reaches `await import("rolldown-plugin-dts")` inside its `if (dts)` branch (the statically imported `/internal` subpath does not pull in yuku). A commit here corrects that comment rather than leaving the claim in the tree. With section 1, yuku is now permanently off the build's execution path — though it stays in the lockfile, since `rolldown-plugin-dts` is a hard dependency of tsdown.

## 5. The temporary release-age exemption

pnpm 11 revalidates the lockfile against `minimumReleaseAge` on *every* command (`runDepsStatusCheck`), not just install. yuku 0.8.3 was published 2026-08-02, so without an exemption the locked version fails `pnpm test`, `pnpm check` and CI for everyone until the 7-day gate clears on **2026-08-09**.

The selectors are **pinned to 0.8.3**, not family globs. That matters: `renovate.json` sets `minimumReleaseAge: null` for `lockFileMaintenance` explicitly because pnpm enforces the delay natively, and that job runs with `automerge: true` on **Friday 2026-08-07 — inside this exemption's window**. An unversioned `yuku-*` would have let a yuku published days earlier land with no delay from either side. Verified the pin bites: with the list saying `yuku-ast@0.8.2`, `pnpm install` rejects the locked 0.8.3.

Two consequences of pinning worth noting:

- The binding selectors stay unversioned, because pnpm rejects `pattern@version` (`Name patterns are not allowed with version unions`). Enumerating all 24 bindings at `@0.8.3` was the alternative; not taken, since each binding is an exact-version dependency of a now-pinned parent, so no other version can enter the tree.
- If the block is ever forgotten past 2026-08-09 it degrades to a no-op rather than a hole — by then 0.8.3 clears the gate on its own, and no other version is exempt.

For reference, Renovate would not reach 0.8.3 on its own before 2026-08-14: the Friday 2026-08-07 `lockFileMaintenance` run gets to yuku 0.8.1 at best, and 0.8.2 does not reference the android binding anyway.

## Author checklist
- [x] `pnpm typecheck` passes
- [ ] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md`
- [x] If the change has **MCP-runtime-observable** behaviour: pure tooling / dev-tooling is exempt — the standard gates cover it.

Verified on linux-x64: `pnpm build`, `pnpm typecheck` (native TS 7), `pnpm check`, `pnpm test:smoke` (both stdio and HTTP transports reach ready), 593 tests across 32 files, coverage 96.79% statements / 83.34% branches / 98.53% functions / 97.47% lines. `npm pack --dry-run` ships `dist/index.js` and `glama.json` — no declaration file, no map, and no `sourceMappingURL` left in the bundle.

**Not verified on Android** — the yuku bindings sit behind an `os: android` / `cpu: arm64` gate, so they are not downloaded on CI or on the machines used here.

`docs/decisions/biome-on-android.md` cited the yuku graft as a live precedent in its rejected-options table; that row is now phrased in the past tense.

The exemption in section 5 is the one judgement call. The alternative is holding this branch until 2026-08-09 and merging with no config change at all — the lockfile is identical either way, but CI stays red for a week.

Section 2 went through an intermediate state in this branch's history: the map was kept but excluded from the tarball via `"!dist/**/*.map"`. Once generation stopped, that exclusion became dead config and was removed with it — the diff against `main` shows only the end state.

No runtime behaviour changes — build toolchain only. Nothing a client observes at runtime differs, so no functional validation plan.


tsdown.config.ts
```diff
@@ -1,21 +1,20 @@
...
-
 export default defineConfig({
   entry: ['src/index.ts'],
   format: ['esm'],
   target: 'node20',
   outDir: 'dist',
   clean: true,
   sourcemap: true,
-  dts: dtsAvailable,
+  // This package is a binary, not a library: `src/index.ts` exports nothing, so
```
dlecan: Nitpick: one line comment only!!

pnpm-workspace.yaml
```diff
@@ -12,30 +12,21 @@ overrides:
...
+# TEMPORARY — remove on or after 2026-08-09, then re-run `pnpm install`.
 #
-# blockExoticSubdeps must be off for pnpm to accept these URL-resolved
-# subdependencies — a deliberate trade-off: the two grafts below are the only
-# exotic subdeps in the tree (the lockfile pins their URL + integrity, so any
-# new one shows up in review), accepted until either the bindings are
-# published on the npm registry under a personal scope (registry deps are not
-# exotic — lets this flag come back on) or upstream ships android-arm64
-# bindings, which deletes this whole block.
+# yuku 0.8.3 (published 2026-08-02) is the first release to declare
```
dlecan: Suggestion: sum up this comment

pnpm-workspace.yaml
```diff
@@ -12,30 +12,13 @@ overrides:
...
+# TEMPORARY — delete on or after 2026-08-09, then re-run `pnpm install`.
+# pnpm re-checks minimumReleaseAge against the lockfile on every command, not
+# just install, so the just-published yuku 0.8.3 (first release with an upstream
+# android-arm64 binding) would fail CI for everyone until its 7-day gate clears.
+# yuku only; every other package keeps the full gate.
+minimumReleaseAgeExclude:
+  - yuku-*
+  - '@yuku-codegen/*'
+  - '@yuku-parser/*'
+  - '@yuku-toolchain/*'
```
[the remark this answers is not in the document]
dlecan: Adopted the version-selector suggestion in 4c7ab98, with two adjustments.

**Why it mattered more than the glob breadth.** `renovate.json` sets `minimumReleaseAge: null` for `lockFileMaintenance`, explicitly because pnpm enforces the delay natively. That job runs with `automerge: true` on Friday 2026-08-07 — inside this exemption's window. With an unversioned `yuku-*`, a yuku published days earlier would have landed with no delay from either side. Version pinning closes it; verified the pin bites by setting the list to `yuku-ast@0.8.2`, which makes `pnpm install` reject the locked 0.8.3.

**No bare `yuku` package exists.** The tree has `yuku-ast`, `yuku-codegen`, `yuku-parser`, `@yuku-toolchain/types` and the two binding scopes — no `yuku`. "yuku 0.8.3" in the comment is shorthand for the family, so there is nothing to add a `yuku@0.8.3` selector for.

**The binding selectors stay unversioned**, because pnpm rejects the combination:

```
[ERR_PNPM_INVALID_MINIMUM_RELEASE_AGE_EXCLUDE] Invalid value in
minimumReleaseAgeExclude: Name patterns are not allowed with version unions.
Found: "@yuku-codegen/*@0.8.3"
```

Enumerating all 24 bindings at `@0.8.3` was the alternative. Not taken: each binding is an exact-version dependency of a now-pinned parent, so no other version can enter the tree anyway, and the whole block is deleted on 2026-08-09.

---

PR #197, opened 2026-08-02
test(utils): add mutation testing, gate PRs on it, and close the assertion gaps it found

dlecan:
## Summary

Adds mutation testing (StrykerJS), uses it to close the assertion gaps it found in `src/utils`, and gates pull requests on the mutants in their changed lines. Line coverage in `src/utils` is 98.58 % and its ratchet is near saturation — the metric can no longer distinguish a test that *asserts* something from one that merely executes the line.

`src/utils` mutation score **71.59 % → 77.18 %**, 232 → 187 surviving mutants, 593 → 659 tests. **No production code is touched.**

## Linked issue

None. #198 was opened from this work and has since been fixed on `main` by #199 — which this branch merges, and whose own gaps the new gate then caught (see below).

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Tooling / CI
- [x] Tests

## What the baseline showed

982 mutants injected into `src/utils`, **71.59 %** killed:

| | count | |
| --- | ---: | --- |
| Killed / Timeout | 703 | detected |
| **Survived** | **232 (23.6 %)** | **executed by the tests, undetected** |
| NoCoverage | 47 (4.8 %) | never reached — the part coverage already reports |

So the gap between 98.58 % line coverage and 71.59 % is almost entirely assertion quality, not unreached code. `logger.ts` was the clearest case: **100 % branch coverage, 66.28 % mutation score**, because nearly every assertion was a `toContain` on a fragment of the output.

## Result

| File | Before | After | Δ |
| --- | ---: | ---: | ---: |
| `pagination.ts` | 69.77 | **100.00** | +30.23 |
| `validation.ts` | 73.08 | 88.46 | +15.38 |
| `logger.ts` | 66.28 | 84.47 | +18.19 |
| `article-sections.ts` | 73.58 | 83.96 | +10.38 |
| `article-order.ts` | 91.30 | 91.30 | +0.00 |
| `package-info.ts` | 65.71 | 80.00 | +14.29 |
| `formatting.ts` | 66.48 | 67.43 | +0.95 |
| **`src/utils`** | **71.59** | **77.18** | **+5.59** |

**One honesty note on the aggregate**: the baseline was measured before #199 landed on `main`, which added 54 source lines and 8 tests to `logger.ts`. So the total mixes two contributions. Decomposed for that file: 66.28 → 81.40 is this branch's logger tests against the old source; 81.40 → 81.55 is #199 arriving (its tests roughly kept pace with its new code); 81.55 → 84.47 is this branch closing what the gate found in it.

`formatting.ts` is deliberately barely moved — see the scope note at the bottom. `article-order.ts` was already the strongest file and was left alone.

## The CI gate — and what it caught on its first real diff

**`diff-gate` (pull requests) mutates only the changed *lines*** and fails if any of those mutants survived. `pnpm test:mutation:diff <base> <head>` turns the diff into Stryker line ranges (`file:start-end`, which `--mutate` supports natively), mutates exactly those, and judges the report it just produced — one command, so a stale scope can never be paired with a fresh report.

File granularity would have judged a one-line change against every mutant in the file — touching `formatting.ts`, at 67 %, would fail on that file's history rather than on the PR. Measured on a two-line change: **10 mutants instead of the 43** in the whole file.

The gate asks for **no survivors, not a percentage**. On a handful of mutants a percentage is arbitrary — 1 of 3 fails at 67 %, 1 of 10 passes at 90 % — while "a change you made went unnoticed by every test" is predictable and explains itself. `// Stryker disable next-line <mutator>: <why>` is the escape for a genuinely equivalent mutant, which puts the reasoning in the diff where a reviewer sees it.

### It would have blocked #199

This PR cannot exercise the gate on itself (it changes no file in the mutate scope), so it was run against a real diff instead: **#199's own changed lines**. 27 mutants, 24 killed, **3 escaped**, in 26 seconds:

| Line | Code | What the test actually said |
| --- | --- | --- |
| `63` | `return '[function]'` | the `toJSON` test asserted the secret absent and `[REDACTED]` present — never that the callable had been *replaced* |
| `88` ×2 | `return { fields: '[unredactable]' }` | the throwing-getter test used `toContain('hostile')`, which matches the **event name**, present whatever redaction does |

The second is the sharper one: that test passed with the entire fallback removed. Losing the fields when an object fights back is acceptable; losing the trace that they were dropped is not.

Worth stating plainly, because it is the argument for making the check required: **#199 was written after all of this branch's work on loose assertions, by someone with that context, and reproduced the same pattern.** Both lines are now pinned in full and the gate reports 27/27 on #199's lines. Of `logger.ts`'s 16 remaining survivors, none are in #199's code — `git blame` confirms they are pre-existing (`renderValue`, `if (server)`, the `REDACTED_KEYS` entries).

**`baseline` (push to main) runs the full scope and is the only writer of the incremental cache.**

Two things worth a careful look, both places this could have gone wrong:

- **The cache key carries the staleness rules**, rather than a conditional `--force` someone has to keep in sync with the decision record. The prefix hashes every input incremental mode cannot diff for itself — `pnpm-lock.yaml`, `stryker.config.mjs`, `vitest.config.ts`, `tests/setup.ts`, `tests/msw-handlers.ts`, the `tests/integration/` scaffolding (`harness.ts`, `core-scenarios.ts`, `stdio-harness.ts`), `.nvmrc` — so changing one matches no entry, and no baseline means a cold run. Structural instead of remembered, and a unit test now asserts the hash covers every file under `tests/` the suite can load, since that list is the one thing here whose drift fails silently. `tests/functional/` is exempt by path: `vitest.config.ts` never loads it, and it gains a report per scenario run that would otherwise discard the baseline each time.
- **The key ends in the commit sha and the fallback is the prefix.** Actions cache entries are immutable, so a fixed key would have frozen the first baseline forever. A per-commit key keeps saves fresh and the prefix `restore-keys` lets a PR fall back to main's latest — sound *only* because the inputs hash sits inside the prefix. A broader restore-key would reintroduce the stale reuse this exists to prevent.

Also verified after merging `main`: `pnpm test:mutation:diff origin/main HEAD` returns **empty**. Merging main in does not make this PR answerable for main's mutants — the one behaviour here most likely to have been wrong.

**The gate does not block merges until branch protection marks the `Changed lines` check required.** Worth flipping now that a real diff has been through it.

## The TypeScript 7 workaround

Stryker 9.6.1 does not run on TS 7 out of the box — it throws before the first mutant:

```
ERROR Stryker TypeError: ts.parseConfigFileTextToJson is not a function
    at TSConfigPreprocessor.rewriteTSConfigFile
```

`@stryker-mutator/core` does a bare `await import('typescript')` (undeclared dependency) and calls an API TS 7 does not expose. Upstream: [stryker-js#6110](https://github.com/stryker-mutator/stryker-js/issues/6110), open, no assignee.

The fix points `tsconfigFile` at a file that does not exist. That preprocessor only rewrites relative `extends`/`references` paths; ours are package names, so it has nothing to rewrite and skipping it costs nothing. A pnpm scoped override to feed it the `typescript-legacy` alias is impossible (the dependency is undeclared, so there is nothing to override), and `inPlace: true` was rejected because it mutates the real sources. `@stryker-mutator/typescript-checker` is not installed for the same reason.

**The sentinel asserts its own absence** — creating the file would otherwise resurface the upstream crash as an unrelated-looking `ts.parseConfigFileTextToJson is not a function`, so the config throws with a pointer instead.

## Other tooling notes

- **`incremental` is not enabled by default.** It carries a correctness hazard (a verdict recorded against an older lockfile is replayed, not re-run), so `pnpm test:mutation` is a trustworthy cold run and `--incremental` is an explicit opt-in. CI passes it because the cache key guarantees the baseline matches the current inputs.
- **Two files are out of the mutate scope**, for the same reason and both temporarily. `src/tools/**` (its `.describe()` calls would each yield a `StringLiteral` survivor that says nothing about test quality) and `src/utils/formatting.ts` (171 escaped mutants, 58 % of them label strings, spread over 42 lines). A label-heavy file does not just score badly — it *booby-traps the gate*, since editing any line carrying a survivor fails CI for debt the author did not create. Both are excluded from *mutation* only: their tests still run and still count for coverage. [#203](https://github.com/fruggr/zendesk-mcp-server/issues/203) owns `formatting.ts`, and putting it back in scope is part of its acceptance criteria.
- **The gate mirrors Stryker's `!` semantics.** `scopeMatcher` replaced a `mutate.some(...)` check that ignored negations. That mattered: the gate hands its ranges to `--mutate`, which *replaces* the configured scope rather than intersecting with it, so a `some(...)` match through `src/utils/**/*.ts` would have mutated the excluded file anyway.
- **Vitest discovery is now an allowlist** (`include: ['tests/**/*.test.ts']`). The default repo-wide glob collects any *copy* of the tree: Stryker's sandbox was one instance, and a plain `pnpm test` was silently running every suite three times over against *mutated* sources, reporting 4 failures that do not exist. A git worktree or an unpacked tarball would do the same, so the allowlist closes all of them at once.
- **`.gitignore`'s `reports/` was unanchored** and also swallowed the tracked `tests/functional/reports/` — the next scenario report written there would have been invisible to `git status`. It hid those files from Biome too: `pnpm check` now sees 84 files where it saw 78.

## Tests added — no production code touched

Each of these kills mutants that survived the baseline:

- **`pagination`** — `buildOffsetParams` and `extractSearchPaginationMeta` were never imported by the suite at all. Adds the `count > page * perPage` boundary (a full final page is not "more"), and `count: 0` to prove the `!= null` check is not a truthiness check. Now at 100 %.
- **`logger`** — whole output lines pinned instead of fragments; all four level mappings (only `warn -> warning` was checked); key normalisation across case and separators (`ACCESS-TOKEN`, `refreshToken`, `Code_Verifier`), which the code documents but nothing verified; non-string field values; arrays kept as arrays; a *synchronous* throw from the MCP sink; and the two markers #199 left unasserted.
- **`package-info`** — the cache, the upward walk, and the shapes that must not satisfy it.
- **`validation`** — the full error message, pinning the sorted key list and its separators.
- **`article-sections`** — the out-of-range message, the last valid index, whitespace-only input, heading tag case.
- **`formatting`** — the truncation boundary at exactly `CHARACTER_LIMIT`, and the 120-character macro preview boundary.

## Author checklist
- [x] I checked no open or merged PR already addresses the linked issue
- [x] N/A — no linked issue
- [x] `pnpm test` passes locally (659 tests)
- [x] `pnpm check` is clean
- [x] `pnpm typecheck` passes
- [x] I ran a review pass on the diff and addressed its findings (see below)
- [x] Documentation updated (`docs/decisions/mutation-testing.md`, `AGENTS.md` Testing section)
- [x] N/A — no MCP tool added
- [x] Exempt — tooling, CI and tests. No runtime-observable behaviour changes: `src/` is untouched, so a running server behaves identically for a client.

## Review passes already applied

Four cleanup agents (reuse / simplification / efficiency / altitude), an automated review bot across several rounds, and a correctness pass. What they caught, in the commits above: the unanchored `reports/`, the Vitest denylist→allowlist change, `incremental` defaulting on without its safety net, five different `node:fs` mocking idioms in one file, a logger test the mutation report proved killed nothing, an order-agnostic regex that was weaker than pinning the message, an incomplete invalidation table, and an em dash my own tooling had escaped to `—` in the published `description`.

Efficiency found nothing, with measurements: the 25 k-character truncation fixtures cost 107 µs total (0.001 % of the suite), and `timeoutMS: 20000` costs ~3 % of a cold run for the 0.41 % of mutants that time out.

## Second review round — five defects in the gate machinery

A further review pass after the first round found five, all in the gate rather than in the tests. Recorded here because two of them made the gate wrong, not merely untidy, and one changes a behaviour a reviewer might otherwise read as a regression:

1. **`escapedMutants` judged mutants the run never executed.** It used *overlap*, but Stryker instruments a construct only when it is *contained* in a `--mutate` range (`locationIncluded` in the instrumenter's `syntax-helpers`, used by `babel-transformer`). Measured: asking for one line inside a 20-line expression instruments **0 mutants**. Under `--incremental` the non-contained ones are still in the report, replayed from the baseline with **main's** verdicts — so a PR editing one line of a long expression failed on a survivor it never created, unfixable except by testing someone else's code. Now containment. **The test that pinned the old behaviour was inverted**, with the reasoning in the test body, and a contained-multi-line counterpart added.
2. **The diff was taken against the wrong tree.** `base.sha...head.sha`, while Stryker mutates the checked-out *merge* commit — so once the base moved with any change to a file the PR also touches, every range pointed at the wrong lines: untouched code mutated, changed lines ungated. Now `base.sha...HEAD`.
3. **`git diff` ran without pinned path prefixes.** With `diff.noprefix` set the header is `+++ path`, the regex never matches, and the documented local gate **exits 0** on a PR full of in-scope changes. Reproduced, then fixed with `--no-ext-diff --src-prefix=a/ --dst-prefix=b/`.
4. **The cache key omitted two scaffolding files** (`core-scenarios.ts`, `stdio-harness.ts`). Both added, and the drift is now caught by a test rather than by memory.
5. **The job summary buried its own verdict.** Teeing Stryker's stdout dumped every survivor in the whole project (the report is project-wide under `--incremental`) while the gate's actionable list went to stderr and never appeared. `clearTextReporter.reportMutants: false`, list on stdout.

**A blind spot this makes explicit rather than introduces:** because Stryker needs containment, an edit inside a construct larger than the edit yields nothing to mutate. The gate now says so in as many words instead of reporting “nothing to gate”, and the ADR gains *What the gate cannot see* for why that beats the alternative — failing authors for mutants they cannot fix is how a gate gets switched off.

Every claim above was checked against Stryker 9.6.1's own source in `node_modules`, not inferred. The gate was then revalidated end to end on a scratch commit carrying an untested guard: 9 mutants on the changed lines, 2 escaped, step exit **1**.

## Notes for the reviewer

**Deliberately left out of scope:**

1. **`formatting.ts`'s remaining survivors** — now tracked in [#203](https://github.com/fruggr/zendesk-mcp-server/issues/203), which also carries the decision on how to treat the label strings. This PR takes the logic-bearing ones it could (truncation and preview boundaries) and defers the rest by excluding the file from the mutate scope, so the outstanding work cannot fail an unrelated PR. Putting it back in scope is one of #203's acceptance criteria.
2. **`src/tools/**`**, which needs `excludedMutations: ["StringLiteral"]` before it produces usable signal.
3. **Widening `pnpm check` beyond `src/tests/scripts`** so root config files are linted by the repo's own gate. This is how the `stryker.config.mjs` lint error went unseen locally; it is a pre-existing gap that already applies to `vitest.config.ts` and `tsdown.config.ts`, so it wants its own issue rather than being charged here.

**Worth a second opinion:** the `tsconfigFile` sentinel is load-bearing and non-obvious. If it is ever "cleaned up" to point at the real `tsconfig.json`, every mutation run dies. It now throws a named error rather than the upstream one, is commented in the config and explained in the ADR — but it is still the kind of thing a future tidy-up removes.

**One judgement call worth checking:** a handful of existing assertions were *strengthened* rather than added to — `toContain('platform=win32')` became an equality check on the whole line, and two of #199's assertions were tightened the same way. That is the point of the exercise and no test's intent changed, but those tests now fail for more reasons than before.

Cost, for planning: a cold run over the configured scope is ~1 700 mutants; `src/utils` alone takes ~9 min on 4 cores. Incremental reruns are 17 s (nothing changed) to 47 s (one test file changed). The PR gate, being line-scoped, came in at 26 s on #199's diff.

## Summary by CodeRabbit

* **Testing**
  * Added changed-code mutation testing with cached baselines and CI reporting.
  * Expanded coverage for formatting, logging, pagination, validation, package metadata, article sections, and mutation-scope behavior.
* **Documentation**
  * Documented mutation-testing practices, contributor requirements, and implementation decisions.
* **Chores**
  * Improved test discovery and ignored generated mutation-testing artifacts.
  * Added automated mutation-test reporting for pull requests and baseline updates.

---


dlecan:
All four review findings applied in a23d31f. Two are worth a note.

**`noDefaultExport` on `stryker.config.mjs` — real, and it exposed a blind spot.** Confirmed locally:

```
$ node scripts/biome.mjs check --error-on-warnings stryker.config.mjs
  × Avoid default exports.
  > 5 │ export default {
```

Green CI was not evidence against it: `pnpm check` runs over `src/ tests/ scripts/`, and lefthook's pre-commit glob is scoped the same way, so **no root-level config file is linted by this repo's own gate**. Fixed as suggested, by adding `**/*.config.mjs` to the exemption `biome.json` already grants `**/*.config.ts` and `**/*.config.mts` — Stryker's native ESM config requires the default export.

Worth flagging for a separate PR, since it is out of scope here: the same gap means `vitest.config.ts`, `tsdown.config.ts` and `biome.json` itself are unlinted. Widening `pnpm check`'s scope is the real fix.

**Stale incremental baselines — the finding is right, though the sentence it quotes was scoped.** "It costs time, never correctness" was about *Vitest's coarse per-file test detection*, where it does hold: marking a whole file's tests as changed only ever re-runs more than needed. But the broader point is correct and I had not covered it — incremental mode diffs mutated sources and test files and nothing else, so a dependency, environment or config change leaves verdicts that are *replayed rather than re-run*. That is a correctness problem, and it directly affects the CI design proposed in the ADR.

Added §4 "The baseline can go stale, and Stryker will not tell you": when `--force` is mandatory (lockfile, `stryker.config.mjs`, Node version), what the CI cache key has to hash so a change misses the cache instead of reusing it, and a note at the `incremental` option itself so it is visible at the point of use. The neighbouring Vitest sentence is now explicitly scoped so it cannot be read as covering incremental mode at large.

The other two — labelling `71.59 %` / 232 survivors as the pre-change baseline (with the `76.78 %` result alongside), `26` → `27` points, the `src/utils`-versus-repo-wide coverage scopes, and dating the ecosystem claim — are straightforward and applied as described.

Gate is green after the changes: `pnpm check`, `pnpm typecheck`, `pnpm test` (633).

---

.github/workflows/mutation.yml
```diff
@@ -0,0 +1,191 @@
...
+          else
+            echo "Mutating the changed lines: \`${specs}\`" >> "$GITHUB_STEP_SUMMARY"
+          fi
+
+      # `--incremental` is safe to pass unconditionally: the baseline is only on
+      # disk when a key matched, and the key encodes every input incremental mode
+      # cannot detect for itself.
+      - name: Mutate the changed lines
+        if: steps.scope.outputs.specs != ''
+        run: pnpm exec stryker run --incremental --reporters json,progress --mutate "${{ steps.scope.outputs.specs }}"
```
dlecan: Suggestion: use a pnpm script

.github/workflows/mutation.yml
```diff
@@ -0,0 +1,191 @@
...
+          cache: pnpm
+
+      - run: pnpm install --frozen-lockfile
+
+      # Same scheme as the gate job above — see the comment there for why the
+      # prefix is a hash of the inputs incremental mode cannot diff.
+      - name: Compute the baseline cache key
+        id: key
+        env:
+          PREFIX: mutation-baseline-${{ hashFiles('pnpm-lock.yaml', 'stryker.config.mjs', 'vitest.config.ts', 'tests/setup.ts', 'tests/msw-handlers.ts', 'tests/integration/harness.ts', '.nvmrc') }}
```
dlecan: Nitpick: this line should be mutualized with the other one in the same file
Even the full algorithm should be mutualized

.github/workflows/mutation.yml
```diff
@@ -0,0 +1,191 @@
...
+      - name: Restore the mutation baseline
+        uses: actions/cache/restore@v6
+        with:
+          path: reports/mutation/stryker-incremental.json
+          key: ${{ steps.key.outputs.exact }}
+          restore-keys: ${{ steps.key.outputs.prefix }}
+
+      # Never `break`s: this job reports and refreshes the baseline. The gate
+      # belongs on the diff, in the job above.
+      - run: pnpm exec stryker run --incremental --reporters json,html,clear-text,progress
```
dlecan: Suggestion: pnpm script?

.github/workflows/mutation.yml
```diff
@@ -0,0 +1,191 @@
...
+
+      - name: Score summary
+        if: always()
+        run: |
+          set -euo pipefail
```
dlecan: Nitpick: don't use inline script for that

AGENTS.md
```diff
@@ -81,6 +81,18 @@ quality bar still hold — the freedom is from Zendesk's shape, not from craft.)
 - Inter-LLM functional tests (proxy annotations, `[RO]` prefix on `tools/list`)
   live in `tests/functional/`; invoke via `/functional-testing`. Details in
   `tests/functional/README.md`.
+- `pnpm test:mutation` (StrykerJS) scores whether the tests *assert* anything,
```
dlecan: Nitpick: all of this seems useless for future use
Just describe this project uses Mutation testing this Stryker on a subset of test files
Will agents need something else in the future?

vitest.config.ts
```diff
@@ -5,6 +5,15 @@ export default defineConfig({
     globals: true,
     environment: 'node',
     setupFiles: ['./tests/setup.ts'],
+    // Every suite lives under `tests/`, so say that instead of relying on
```
dlecan: Question : is a so long comment necessary to explain tests are in tests/* ?

scripts/mutation-scope.mjs
dlecan, line 1: Question : a very complex script to do ... I don't really understand 
Why is this necessary?
How do other users of Stryker? Is it possible to use a third party lib to encapsulate this behaviour?

.github/workflows/mutation.yml
```diff
@@ -0,0 +1,152 @@
...
+  # and a path-filtered required check never reports at all, which would block
+  # every docs-only PR forever. The job is cheap when the diff touches nothing
+  # in scope — it says so and exits.
+  pull_request:
+  push:
+    branches: [main]
+    # The push job runs the *full* scope, so it is worth filtering: a docs or
+    # workflow commit cannot change a single verdict, and after a quiet week
+    # (Actions evicts caches unused for 7 days) it would rebuild cold.
+    paths:
```
dlecan: Question: this pattern seems pretty large. Why not all project's files?
How and when this list should evolve in the future?

.github/actions/mutation-baseline/action.yml
```diff
@@ -0,0 +1,54 @@
...
+    # The trailing commit sha keeps every save unique (Actions cache entries are
+    # immutable, so a fixed key would freeze the first baseline forever); the
+    # prefix restore-key is what allows falling back to main's latest, and it is
+    # safe precisely because the inputs hash sits inside it. A broader
+    # restore-key would reintroduce the stale reuse this exists to prevent.
+    - name: Compute the baseline cache key
+      id: key
+      shell: bash
+      env:
+        PREFIX: mutation-baseline-${{ hashFiles('pnpm-lock.yaml', 'stryker.config.mjs', 'vitest.config.ts', 'tests/setup.ts', 'tests/msw-handlers.ts', 'tests/integration/harness.ts', '.nvmrc') }}
```
dlecan: Question : is it the same list as in `mutation.yaml` ? Why another list?
How and when this list should evolve in the future?

scripts/mutation-scope.mjs
dlecan: Fair question, and the honest answer needed checking rather than asserting. Three findings, all verifiable:

1. **StrykerJS has no git-aware scoping.** Its 9.6.1 schema declares 53 options; none is `since`, `range`, `diff` or `changedFiles`:
   ```
   $ node -e 'const s=require("./node_modules/@stryker-mutator/core/schema/stryker-schema.json");
     console.log(Object.keys(s.properties).filter(k=>/sinc|rang|diff|baselin|git/i.test(k)))'
   []
   ```
   `--since` / `--with-baseline` do exist — in **Stryker.NET**, a different product. This is the most common wrong turn when reading about it; search results and LLMs conflate the two freely, and I nearly did earlier in this PR.

2. **Upstream knows it's missing.** [stryker-js#2843](https://github.com/stryker-mutator/stryker-js/issues/2843) — "Generate mutant only for the changed lines in a range of commits" — is the open request for exactly this, marked stale. `--incremental` is the nearest built-in and answers a different question: it makes a *full* run cheap by reusing verdicts, but it does not restrict the verdict to the diff, which is what a gate needs.

3. **The one third-party library is unmaintained and file-level.** `stryker-diff-runner`, last published 2.3.11 in **November 2022** (Stryker 6 era, against 9.6.1 today), single maintainer. It scopes by *file*, and #2843's own author describes the consequence: mutants are generated for the whole file, so the break threshold has to be disabled. That is precisely the failure this gate exists to avoid — a one-line change to `formatting.ts` would be judged against that file's whole history. Adopting it would mean writing the line-range half anyway, on top of an abandoned dependency.

So what's left is genuinely small: turn a unified diff into `--mutate` line specs, and read back a JSON report that has a published schema. Both halves are pure functions of a few lines. What earns them a file rather than inline YAML is that they're **tested** — `tests/unit/mutation-scope.test.ts` pins the hunk parsing and the range arithmetic, because an off-by-one there would silently stop guarding a line and nothing else in the suite would notice.

Recorded in the ADR as a new subsection, *Why a script and not a library*, including the exit condition: if #2843 ships, native line-range scoping replaces the first half and `thresholds.break` over a scoped run could replace the second. The file header now states the same in three lines instead of leaving you to infer it.

---

.github/workflows/mutation.yml
dlecan: You're right, and the fix is to delete it — done in 7f4edb2.

The list was a second copy of "what can change a mutation verdict", after the cache-key hash in the composite action. Two hand-synced copies of the same knowledge, and the failure modes are not symmetric: too broad costs a few minutes of CI, **one entry short leaves every later PR restoring a silently stale baseline**. That's the correctness hazard the invalidation rule exists to prevent, so reintroducing it here to save a few minutes was a bad trade. It was already incomplete, in fact — `package.json` was missing.

"Why not all project's files" is the right instinct, and there's a second reason it costs almost nothing: the run this filter skipped is the **cheap** one. With the baseline restored, a push touching nothing in scope is a dry run plus a report, not the ~1h cold run. And running on every push keeps the cache entry warm against Actions' 7-day eviction — so removing the filter actually helps the case the comment was worried about.

So: no list, no evolution rule, nothing to keep in sync. `push: branches: [main]`, unfiltered. Reasoning recorded in §4 of the ADR.

---

.github/actions/mutation-baseline/action.yml
dlecan: Different list, and after 7f4edb2 it is the **only** one — I deleted the `paths:` filter in `mutation.yml` rather than explain it, so there is nothing left to keep in sync.

They answered different questions, which is why they differed:

| | Question | Consequence of getting it wrong |
| --- | --- | --- |
| `paths:` (deleted) | should the baseline job run on this push? | a few wasted CI minutes, or a stale baseline |
| this hash | is an existing baseline still valid? | a replayed `Killed` verdict that was never re-run |

The key thing this list *is*: not "files that matter to mutation testing", but **the complement of what Stryker's incremental mode can diff for itself** — mutated sources plus the test files it discovers. That's why `src/**` and `tests/**/*.test.ts` are conspicuously absent: hashing them would throw the baseline away on every commit and buy nothing, since incremental mode already handles them correctly. Everything else that can flip a verdict must be hashed, because nothing else will notice it changed. `tests/setup.ts`, `tests/msw-handlers.ts` and `tests/integration/harness.ts` are in there for exactly that reason — Stryker diffs *discovered test files*, so shared scaffolding is invisible to it.

**When it evolves:** a new file that changes how the tests behave without being a test file Stryker discovers — another setup file, another shared fixture or harness, a runner config. Not docs, not workflows, not the mutate scope. Both the comment on this block and the allowlist table in §4 of the ADR now state that rule; the ADR's version also notes the enumeration cannot be trusted to stay complete, which is why it's phrased as an allowlist ("anything outside `src/**` and `tests/**/*.test.ts` needs `--force`").

---

AGENTS.md
dlecan: Agreed — cut from 12 lines to 6 in 7f4edb2, which is one bullet in the Testing section:

> - Mutation testing (StrykerJS) runs over the scope in `stryker.config.mjs`, and CI fails a PR on a mutant that survived **in a line it changed** (`pnpm test:mutation:diff origin/main HEAD` reproduces it). A survivor means an assertion is missing or too loose — tighten it, never weaken one to go green; for a genuinely equivalent mutant say why with `// Stryker disable next-line <mutator>: <reason>`. Rationale, scope and costs: `docs/decisions/mutation-testing.md`.

The file's own preamble says not to document what the code shows, and I'd broken it: the scope, the cost figures and the TS 7 workaround were all restatements of `stryker.config.mjs` and the ADR.

On "will agents need something else": what's left is the part *not* visible in the code, and it's the part that matters for behaviour — that the correct response to a survivor is to tighten the assertion, never to loosen it. That's the one instruction an agent will get wrong by default, since making the run green by weakening a `toContain` is the locally easiest move and defeats the entire point. The rest is a pointer.

I kept `stryker.config.mjs` named rather than saying "a subset of test files", because it's the *sources* that are scoped, not the tests — an agent reading "subset of test files" would look in the wrong place.

---

vitest.config.ts
dlecan: No. Cut to four lines in 7f4edb2:

```ts
// An allowlist, not the default repo-wide glob: the default also collects
// any *copy* of the tree (Stryker's `.stryker-tmp` sandbox, a worktree, an
// unpacked tarball), which runs every suite twice over against mutated
// sources.
include: ['tests/**/*.test.ts'],
```

The comment isn't there to explain that tests live in `tests/` — you're right that the line says that. It's there for the non-obvious half: why an **allowlist** rather than adding `.stryker-tmp` to `exclude`, which is the reflex fix. This wasn't hypothetical — a leftover sandbox made `pnpm test` collect 96 files instead of 32, with 4 failures that don't exist in the repo, because Vitest was running the suites against Stryker's *mutated* copies. An allowlist closes the worktree and unpacked-tarball variants at the same time; excluding one directory closes only that one.

Trimmed the war story, kept the reason.

---

.github/workflows/mutation.yml
dlecan: Done in 7f4edb2 — two new scripts, and the workflow calls those instead of `node scripts/…`:

```json
"test:mutation:diff": "node scripts/mutation-scope.mjs diff",
"test:mutation:summary": "node scripts/mutation-scope.mjs summary",
```

The step you commented on (`pnpm exec stryker run --incremental --reporters json,progress --mutate …`) is gone as well — 57552ca collapsed the three steps into one command, so the flags no longer live in YAML at all.

They're the documented entry point everywhere now: `CONTRIBUTING.md`, `AGENTS.md`, the ADR appendix, and the script's own usage string. No `--` separator anywhere, since pnpm 11 forwards it literally.

One thing I checked rather than assumed, given 026ad90 was about exactly this: the pnpm wrapper still propagates a failing exit code through the `tee` pipe. Verified end to end on a scratch commit with an untested guard — 9 mutants on the changed lines, 2 survived, step exited **1**; and the same shape without `shell: bash` still exits 0, so that line remains load-bearing.

---

.github/workflows/mutation.yml
dlecan: Both done, in 57552ca — and you were right that the *whole algorithm* was the thing to extract, not just the duplicated `hashFiles(...)` line.

It's now the composite action `.github/actions/mutation-baseline`, holding pnpm/Node setup, install, the cache-key computation and the restore. Both jobs are:

```yaml
- uses: ./.github/actions/mutation-baseline
```

The action's `description` states why it has to be one place: the two jobs must agree on cache identity, and only one of them writes — so a divergence wouldn't fail loudly, it would silently make PRs restore a baseline the push job never saved.

The follow-up you'd have hit next: `mutation.yml` still had a *second* copy of that knowledge in the `paths:` trigger filter. Deleted in 7f4edb2 (see the thread on line 28), so there's now exactly one list in the whole setup.

---

.github/workflows/mutation.yml
dlecan: This one is now `pnpm test:mutation --incremental` (57552ca) — the reporter list moved into `stryker.config.mjs` where it belongs, so the workflow no longer restates it.

Worth recording why there's no `--` in it: pnpm 11 forwards the separator to the script *literally*, so `pnpm test:mutation -- --incremental` becomes `stryker run -- --incremental` and fails with "too many arguments for 'run'". All four documented forms were wrong that way at one point; they're now all executed rather than written from memory.

---

.github/workflows/mutation.yml
dlecan: Agreed — the inline `node -e '…'` that parsed the report and built the Markdown table is gone (57552ca). It's now:

```yaml
- name: Score summary
  if: always()
  run: pnpm test:mutation:summary >> "$GITHUB_STEP_SUMMARY"
```

The tally and score arithmetic moved into `scripts/mutation-scope.mjs` as `tallyStatuses` / `scoreOf`, which are unit-tested — including that compile errors stay out of the denominator, which the inline version got wrong and nothing would have caught. That's the real argument against inline scripts in YAML here: not the length, but that it can't be tested or linted.

---

scripts/mutation-scope.mjs
dlecan: Follow-up on this thread, since a later review round changed the substance of the script rather than just its documentation.

Reviewing the gate itself turned up **five defects, two of which made it wrong** — and both were in the half I described above as "small". Worth recording here because it qualifies the "genuinely small" claim:

1. **It judged mutants the run never executed.** `escapedMutants` used *overlap*. But Stryker instruments a construct only when it is **contained** in the range — `locationIncluded` in the instrumenter's `syntax-helpers.js`, called from `babel-transformer.js:92`. Measured on this repo: asking for a single line in the middle of a 20-line expression in `formatting.ts` instruments **0 mutants**.

   Under `--incremental` the non-contained ones are still in the report, replayed from the baseline with **main's** verdicts (`incremental-differ`: *"old mutants that didn't run this time around"*). So a PR editing one line of a long expression failed on a survivor it never created — unfixable except by writing tests for someone else's code. Exactly the "judged on the file's history" failure I said line ranges avoid, arriving by a different door.

2. **The diff was taken against the wrong tree** — `base.sha...head.sha`, while Stryker mutates the checked-out *merge* commit. Once the base moves on a file the PR also touches, every range points elsewhere: untouched code mutated, changed lines ungated.

3. `git diff` ran without pinned path prefixes, so a contributor with `diff.noprefix` got `+++ path`, no hunk matched, and the documented local gate **exited 0** on a PR full of in-scope changes.

The other two were the cache key missing two scaffolding files, and the job summary drowning its own verdict in the project-wide survivor list.

**This sharpens rather than weakens the answer to your question.** The two halves really are small — but "small" turned out to mean "small and wrong in three places", and every one of those was caught by a test or a measurement, not by reading. `stryker-diff-runner` would have given me the file-level version of defect 1 permanently, with no test suite of my own to find it in.

One thing the fix makes explicit: with containment, an edit inside a construct larger than the edit yields **nothing to mutate**, so the gate is blind there. It now says so instead of printing "nothing to gate", and the ADR gained *What the gate cannot see* explaining why that beats the alternative — a gate that fails authors for mutants they cannot fix gets switched off.

---

tests/unit/mutation-scope.test.ts
```diff
@@ -0,0 +1,238 @@
...
+    );
+    const call = /hashFiles\(([^)]*)\)/.exec(action);
+    expect(call?.[1], 'the action must compute the prefix with hashFiles(...)').toBeDefined();
+    const hashed = [...(call?.[1] ?? '').matchAll(/'([^']+)'/g)].map((m) => m[1]);
+
+    // Sorted comparison, so the failure message names the missing file.
+    const alphabetically = (a: string, b: string) => a.localeCompare(b);
+    expect(hashed.filter((f) => f.startsWith('tests/')).sort(alphabetically)).toEqual(
+      nonTestFilesUnder('tests').sort(alphabetically),
+    );
```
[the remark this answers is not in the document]
dlecan: Taken in part, in 278f3cf — the diagnosis of my test was right, the prescription wasn't.

**What was genuinely wrong:** filtering candidates by `.ts` extension. It gives the correct answer today only by coincidence, and a JSON fixture or an `.mjs` helper that a suite reads at runtime would have slipped past the very check that exists to catch silent drift. The guard now compares against *every* file under `tests/`, whatever the extension. Verified both directions: adding `tests/fixtures/probe.json` fails it by name, and removing it passes.

**What I'm not doing: hashing `tests/functional/`.** Two reasons, and the second is decisive.

1. It cannot change a mutation verdict. `vitest.config.ts` is `include: ['tests/**/*.test.ts']`, so nothing in that directory is loaded by the suite Stryker runs. It's the inter-LLM harness driven by hand through `/functional-testing`.
2. It would make the cache *worse*. That harness writes a report per scenario run — `reports/*.report.md`, `*.verdict.md`, `*.raw.json`. Hashing them means every recorded run invalidates the baseline and the next PR pays a cold run (measured ~1h over the configured scope) for a file no test reads. That is the opposite of what the invalidation rule is for: it exists to prevent *stale* verdicts, not to discard *valid* ones.

So the exemption is now stated by path with that reasoning, in the test, in the action's comment and in the ADR — rather than falling out of an extension filter by accident, which was the fair part of your point.

Two notes on the finding itself, for the record: the OpenGrep `child_process.exec` warning on line 203 is a false positive — there is no `child_process` in this file, that line is `readFileSync` on the action YAML. And "fixtures, scenarios, reports, runners, and documentation" describes files that exist only under `tests/functional/`; there are no fixtures elsewhere under `tests/` today, which the new all-extension comparison would now catch if any appeared.

---

docs/decisions/mutation-testing.md
```diff
@@ -0,0 +1,382 @@
...
+"what can change a verdict", kept in sync by hand, with asymmetric failure modes:
+too broad wastes a few minutes of CI, one entry short leaves every later PR
+restoring a stale baseline. So the trigger is unfiltered. What that filter would
+have saved is the cheap run anyway — with the baseline restored, a push touching
+nothing in scope is a dry run plus a report — and running on every push keeps the
+entry warm against the 7-day eviction.
+
+That equivalence — the hash covering *every* non-`*.test.ts` file under
+`tests/` — is enforced by a unit test rather than by this paragraph, because
+forgetting one is exactly the silent failure the rule exists to prevent.
```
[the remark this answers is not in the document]
dlecan: Taken in part, in 278f3cf — flagging clearly that this is **not** full compliance, since the thread is marked addressed.

**Removed**, because these counts were incidental and fragile: Stryker's option total (the claim is "nothing in its schema is `since`/`range`/`diff`" — the total adds nothing and goes stale on any upstream release) and the number of `tests/integration/` scaffolding files (which goes stale precisely when someone adds a fourth, and that drift is now caught by a test instead).

**Kept: the measured figures.** The cited guideline does not apply to them, on two counts:

1. **The path scope quoted here doesn't exist.** There is no `**/*.{md,ts}` entry in `.coderabbit.yaml` — its `path_instructions` are `src/tools/**`, `tests/**` and `.github/workflows/**`. The real rule lives in `AGENTS.md`/`CLAUDE.md` and is about **tool** counts in the tool documentation: *"no per-section `(N tools)` or global tool totals"*. Same file, next sentence: *"Exact counts still live where they're load-bearing"*, naming `tests/unit/routing/registry.test.ts` and the `tests/unit/tools/*.test.ts` length assertions.
2. **This document is a measurement record.** Its own preamble is *"Every number below was measured on this repository, on the 4-vCPU CI container"*, and it carries a **Date** row for that reason. `0 mutants instrumented for a single line inside a 20-line expression` is not decoration — it is the evidence for the containment fix in the same commit, and without it the section asserts a behaviour instead of demonstrating it. Same for the cost table and the score deltas: strip them and the ADR stops being auditable, which is the only thing it is for.

The staleness concern is real and handled differently: figures are dated and scoped, and the doc says which parts to re-measure if the decision is revisited.

Two smaller points from the same comment: the LanguageTool "GitHub capitalisation" hit is on the literal path `.github/actions/mutation-baseline`, so it stays lowercase; and "exactly" is gone from that sentence, incidentally.

---

.github/actions/mutation-baseline/action.yml
```diff
@@ -0,0 +1,71 @@
...
+    # The trailing commit sha keeps every save unique (Actions cache entries are
+    # immutable, so a fixed key would freeze the first baseline forever); the
+    # prefix restore-key is what allows falling back to main's latest, and it is
+    # safe precisely because the inputs hash sits inside it. A broader
+    # restore-key would reintroduce the stale reuse this exists to prevent.
+    - name: Compute the baseline cache key
+      id: key
+      shell: bash
+      env:
+        PREFIX: mutation-baseline-${{ hashFiles('pnpm-lock.yaml', 'stryker.config.mjs', 'vitest.config.ts', 'tests/setup.ts', 'tests/msw-handlers.ts', 'tests/integration/harness.ts', 'tests/integration/core-scenarios.ts', 'tests/integration/stdio-harness.ts', '.nvmrc') }}
```
dlecan: Question : how are these files selected?

.github/workflows/mutation.yml
dlecan, line 1: Nitpick: compress all comments, there are too many of them

.github/actions/mutation-baseline/action.yml
```diff
@@ -0,0 +1,71 @@
...
+
+    - uses: actions/setup-node@v7
+      with:
+        node-version-file: .nvmrc
+        cache: pnpm
+
+    - run: pnpm install --frozen-lockfile
+      shell: bash
+
+    # THE INVALIDATION RULE — the only list of its kind in the setup, and the
```
dlecan: Nitpick: too much comments

.github/actions/mutation-baseline/action.yml
dlecan: By two mechanical rules — and that you had to ask, with 25 lines of comment sitting directly above the line, is the answer to your other two comments. Rewritten in a5bdc0e to state the rule instead of arguing for it:

```yaml
# THE INVALIDATION RULE. Selection is mechanical, in two parts:
#   1. the toolchain and runner inputs — lockfile, both configs, .nvmrc;
#   2. every file under tests/ except *.test.ts and tests/functional/.
# Part 2 is asserted by tests/unit/mutation-scope.test.ts, so this list
# cannot drift silently. `src/**` and *.test.ts are absent on purpose:
# Stryker's incremental mode diffs those itself.
```

Part 2 is the part worth trusting, because it isn't a judgement call any more — `tests/unit/mutation-scope.test.ts` walks `tests/` and compares against this exact `hashFiles(...)` list. Add a shared fixture and the test fails naming the file; that is now the mechanism, not the comment.

The two exclusions are the only choices in it:

- **`src/**` and `*.test.ts`** — Stryker's incremental mode diffs these itself, so hashing them would discard the baseline on every commit for no gain.
- **`tests/functional/`** — the by-hand inter-LLM harness. `vitest.config.ts` loads only `tests/**/*.test.ts`, so it cannot change a verdict, and it gains a report per scenario run that would otherwise throw away the baseline each time one is recorded.

Everything else lands in part 1 because nothing else notices it changed.

---

.github/workflows/mutation.yml
dlecan: Done in a5bdc0e. Measured, since "too many" deserves a number:

| | comment lines | total lines |
| --- | ---: | ---: |
| `mutation.yml` before | 72 | 159 |
| `mutation.yml` after | **17** | 104 |
| `action.yml` before | 29 | 71 |
| `action.yml` after | **11** | 52 |

45 % of the file was prose. Verified the change is comments only — non-comment content is byte-identical, checked by diffing both files stripped of comments against `HEAD`; the one exception is the composite action's `description`, shortened on purpose.

The reasoning now lives only in `docs/decisions/mutation-testing.md`, which is where it belonged. What I kept in the YAML is the two things that read as noise and break the gate when removed, in five lines instead of twenty-seven:

```yaml
# Two things here are load-bearing and look removable. `shell: bash` gives
# `-eo pipefail`; the default `bash -e` has none, so `tee` would swallow the
# gate's exit 1 and every failure would report green (measured). `HEAD`, not
# `head.sha`: checkout gives us the merge commit, and the line numbers must
# match the tree Stryker mutates. Both explained in full in the ADR, §4.
```

Both of those were real bugs in this PR, not hypotheticals, which is why they earn their lines while the rest didn't.

Worth naming the pattern, since it's mine: each of your earlier questions ("why another list", "how does it evolve") I answered by *adding* a paragraph to the YAML. Four rounds of that is how a workflow file ends up 45 % prose, and the file got harder to read with every answer — your comment on line 42 asking how the cache files are selected, sitting under 25 lines meant to explain exactly that, is the proof. The right move was to put the reasoning in the ADR once and make the rule mechanical enough that the code answers the question.

---

.github/actions/mutation-baseline/action.yml
dlecan: Done in a5bdc0e — 29 comment lines of 71 down to **11 of 52**.

The block you commented on was the worst offender: 25 lines, four paragraphs, arguing for a list. It is now six, stating the selection rule and naming the test that enforces it (full answer in the [thread on line 42](https://github.com/fruggr/zendesk-mcp-server/pull/197#discussion_r3714294493)). The `description:` at the top lost a sentence too.

What survives is the cache-key mechanics — prefix + sha because Actions entries are immutable, and why the prefix restore-key is safe to fall back on — in three lines. That one stays because a "simplification" to a fixed key silently freezes the first baseline forever, and nothing fails to tell you.

Everything else moved to `docs/decisions/mutation-testing.md`, §4.

---

.github/actions/mutation-baseline/action.yml
```diff
@@ -0,0 +1,52 @@
...
+    #   1. the toolchain and runner inputs — lockfile, both configs, .nvmrc;
+    #   2. every file under tests/ except *.test.ts and tests/functional/.
+    # Part 2 is asserted by tests/unit/mutation-scope.test.ts, so this list
+    # cannot drift silently. `src/**` and *.test.ts are absent on purpose:
+    # Stryker's incremental mode diffs those itself. Why each part is what it is:
+    # docs/decisions/mutation-testing.md, §4.
+    #
+    # Key = prefix + sha (cache entries are immutable, so a fixed key would
+    # freeze the first baseline); the prefix restore-key is safe to fall back on
+    # only because the inputs hash sits inside it.
```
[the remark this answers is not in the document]
dlecan: Correct, and worse than stated — fixed in d4ac52d. Verified each link in the chain:

- `pnpm/action-setup@v6` is used with **no `version:` input** here, so it does read the version from `package.json`.
- That field is pinned: `packageManager: pnpm@11.17.0+sha512...` — so it is a real, editable input, not a theoretical one.
- pnpm's layout is not incidental to this setup. Stryker's plugin auto-discovery already fails under it, which is why `plugins: ['@stryker-mutator/vitest-runner']` is declared explicitly in the config. A pnpm change is exactly the class of thing that can flip a verdict without touching a source or test file.

The part your finding doesn't mention, and the reason this is a defect rather than a hardening suggestion: **the ADR already claimed it was covered.** Its list of inputs incremental mode cannot detect reads *"`pnpm-lock.yaml` and `package.json` (dependency and devDependency versions)"* — so the hash contradicted the rule it exists to implement. The document was right and the code was wrong.

Worse, I had already spotted the omission and dropped it. In [the thread on the deleted `paths:` filter](https://github.com/fruggr/zendesk-mcp-server/pull/197#discussion_r3707867786) I wrote *"It was already incomplete, in fact — `package.json` was missing"* — about the trigger filter, then never carried it to the place that actually decides cache identity. Noticing a gap and fixing it somewhere else is not fixing it.

`package.json` is now in `hashFiles(...)`, and the comment says why it is there rather than leaving it to read as a duplicate of the lockfile:

```yaml
#   1. the toolchain and runner inputs — lockfile, package.json (its
#      `packageManager` field is where pnpm/action-setup reads the pnpm
#      version from, no `version:` input above), both configs, .nvmrc;
```

Not extending the guard test to part 1: enumerating "which toolchain files matter" in a test would only restate the list, with no independent source of truth to check it against — unlike part 2, where `tests/` on disk is that source. Part 1 stays a judgement call, which is why it now carries its reason inline.

---

PR #199, opened 2026-08-03
fix(logger): stop a circular field from crashing the caller

### LINKED ISSUE #198: Logging a field with a circular reference crashes the caller
## What happens

Passing a value containing a circular reference to any logger method throws `RangeError: Maximum call stack size exceeded` out of the call that emitted the log.

```ts
const log = createLogger('debug');
const circular: Record<string, unknown> = {};
circular['self'] = circular;

log.info('evt', { circular }); // throws
```

```
RangeError: Maximum call stack size exceeded
 ❯ isSensitive src/utils/logger.ts:54:39
 ❯ redactValue src/utils/logger.ts:64:18
 ❯ redactValue src/utils/logger.ts:64:52
 ❯ redactValue src/utils/logger.ts:64:52
 …
```

## Why

`renderValue` has a guard for exactly this class of value:

```ts
try {
  return JSON.stringify(value);
} catch {
  return '[unserializable]';
}
```

but it never gets the chance. `emit` runs `redactValue(fields)` *first*, and `redactValue` recurses into the object graph unconditionally — no cycle tracking:

```ts
const redactValue = (value: unknown): unknown => {
  if (Array.isArray(value)) return value.map(redactValue);
  if (value && typeof value === 'object') {
    const out: Record<string, unknown> = {};
    for (const [key, val] of Object.entries(value as Record<string, unknown>)) {
      out[key] = isSensitive(key) ? '[REDACTED]' : redactValue(val);
    }
    return out;
  }
  return value;
};
```

The stack blows inside the redaction walk. `emit` does not wrap it, so the `RangeError` propagates to whatever called `log.info`.

## Why it matters

A logger must never be able to break the flow that logged. The surrounding code already treats logging as best-effort — the MCP forward is wrapped in `try`/`catch` precisely so "it must never break the flow that emitted the log" — but the stderr path has no such guarantee.

Circular structures are not exotic in the values that reach this logger: Node error objects, HTTP request/response pairs and socket handles all self-reference. Any `log.error('…', { err })` with such an object turns a diagnostic into a crash.

## Where it came from

Found by the first StrykerJS run (#197) while writing a test for the `[unserializable]` fallback — the mutant on that `catch` was reported `NoCoverage`, i.e. nothing reached it. The test in #197 uses a `BigInt` instead, which reaches the same guard without tripping the recursion, and carries a comment pointing here. That PR is test-only by design, hence this issue rather than a fix in it.

## Suggested direction

Track visited objects during the walk (a `WeakSet`) and substitute a marker such as `'[circular]'` on a repeat visit. That keeps the redaction contract intact — a cycle must not become an escape hatch that skips redaction of the keys inside it — and leaves `renderValue`'s guard for the genuinely unserializable cases (`BigInt`, and values with a throwing `toJSON`).

Worth deciding alongside: whether `emit` should also wrap the stderr path defensively, so no future defect in formatting or redaction can propagate to the caller.

## Acceptance

- [ ] `log.info('evt', { circular })` does not throw and emits a line naming the cycle
- [ ] Keys inside a cycle are still redacted when sensitive
- [ ] The MCP sink receives the same sanitised payload
- [ ] Regression test added

dlecan:
## Summary

`redactValue` walked the field graph with no cycle tracking and ran *before* `renderValue`'s `[unserializable]` guard could ever see the value, so any `log.info('evt', { circular })` threw `RangeError: Maximum call stack size exceeded` out of the call that emitted the log. The walk now carries a `WeakSet` of the objects on the current branch and substitutes `[circular]` on a back-reference to an ancestor; `emit` additionally wraps redaction and the stderr write so no future defect in either can propagate to the caller.

Review turned up a second, related hole in the same walk — a caller-supplied `toJSON` survived redaction and re-exposed the secret at stringify time. Fixed here too; see *Scope addition* below.

## Linked issue

Closes #198

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tooling / CI

## Author checklist
- [x] `pnpm test` passes locally (600 tests, 32 files)
- [x] `pnpm typecheck` passes
- [x] Documentation is updated where needed — none needed: no tool, resource, prompt, flag or env var changes; the logger's redaction internals aren't documented outside the source comments, which are updated in the diff
- [ ] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md` — n/a
- [ ] If the change has **MCP-runtime-observable** behaviour … — see below

### On the functional validation plan

**No plan, deliberately — and here is the reasoning rather than a silent omission.** The fix is internal hardening on a path no client can currently reach. Every `logger.*` call site in `src/` passes primitives or pre-extracted strings (`status`, `port`, `errorCode`, `error: message`, `count`, `mode`, …); none forwards a raw error, request/response pair or socket handle, so no sequence of MCP tool calls can construct a field graph with a cycle or a hostile `toJSON`. A validator driving the running server through `mcp__*` tools would therefore exercise exactly the same observable behaviour before and after this diff.

What the fix protects against is the *next* call site that logs a self-referencing or caller-controlled object. That contract is pinned by unit tests, which is where it belongs.

## Changes

`src/utils/logger.ts`

- `redactValue(value, path = new WeakSet())` — tracks the objects on the **current branch** (add on entry, delete on exit) rather than every object ever visited. Consequences, all intentional:
  - a back-reference to an ancestor renders as `[circular]` instead of recursing until the stack blows;
  - the walk keeps descending everywhere else, so a cycle can't become an escape hatch that skips redaction of the keys inside it;
  - a value merely **shared between siblings** (a DAG, not a cycle) is still rendered in full — matching `JSON.stringify`, which also only rejects true cycles;
  - the returned value is always a finite tree, so `renderValue`'s `JSON.stringify` succeeds and both sinks receive the same serialisable payload.
- Function values are replaced with `[function]` — see *Scope addition*.
- `emit` — redaction and the stderr write are each wrapped. Redaction reads caller-supplied objects that may fight back (a getter that throws, an exotic proxy); losing the fields to `fields=[unredactable]` is acceptable, taking the caller down with them is not. This is the same best-effort contract the MCP forward already carried, now applied to the stderr path — which answers the "worth deciding alongside" question in the issue.

`tests/unit/utils/logger.test.ts` — seven regression tests, each written failing first: cycle doesn't throw and names the cycle; sensitive keys inside a cycle are still `[REDACTED]`; sibling-shared values are *not* falsely marked circular; the MCP sink gets the same sanitised, `JSON.stringify`-able payload; a hostile `toJSON` can't re-expose a redacted value on either sink; a throwing getter doesn't propagate; a failing stderr write doesn't propagate.

## Scope addition — `toJSON` redaction bypass

Raised in review, confirmed by reproduction before any code changed:

```
entries:      [ 'token', 'toJSON' ]
stringified:  {"hostile":{"token":"tojson-secret"}}
```

The walk copies own enumerable properties into the sanitised tree, functions included. A caller-supplied `toJSON` therefore survived redaction and was invoked later by `JSON.stringify` — in `renderValue` and again in the MCP transport — handing back the original, unredacted object. `token` was replaced, and the secret still reached both sinks.

Fixed by substituting `[function]` for any function value at the top of `redactValue`, rather than special-casing the `toJSON` key: it closes the class instead of one spelling of it (a key guard leaves `items: [fn]` reachable through the array branch), and loses nothing, since `JSON.stringify` already rendered a function as `undefined`.

**This predates the PR** — the previous `redactValue` copied functions through the same way. Landing it here rather than in a separate PR is a deliberate exception to scope discipline: it's a redaction bypass in the exact function this PR rewrites, under the same hostile-caller threat model as the throwing getter already handled. Happy to split it out if you'd rather keep the diff to the cycle fix.

- **Branch tracking vs. visited-set.** The issue suggests "a `WeakSet` of visited objects". A plain visited-set would also collapse sibling-shared subtrees to `[circular]`, silently dropping diagnostics that are perfectly loggable. Branch tracking costs one `delete` per node and keeps the output faithful; the DAG test pins the difference.
- **`[unredactable]` loses all fields, not just the hostile one.** Deliberate: a partial walk that resumed after a throwing getter would risk emitting values the interrupted redaction never got to inspect. Dropping the whole field set is the conservative choice for a redaction path.
- **Marker vocabulary.** `[function]` rather than reusing `[unserializable]`, which is `renderValue`'s word for values that *failed* to serialise — two different outcomes shouldn't share a marker.
- **Not touched:** `renderValue`'s `[unserializable]` guard, which still covers the genuinely unserializable cases (`BigInt`, a throwing `toJSON`). The test that reaches it lands with #197.
- Exponential blowup on a heavily shared DAG is theoretically possible with branch tracking (each shared subtree re-walked per path). Not guarded: log fields here are shallow diagnostic records, and the alternative loses correctness on the common case.

### DIFF
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -53,17 +53,13 @@ const REDACTED_KEYS = new Set([
 const isSensitive = (key: string): boolean =>
   REDACTED_KEYS.has(key.toLowerCase().replace(/[_-]/g, ''));
 
-// Recursively redact: a sensitive *key* anywhere in the tree (top-level or
-// nested in objects/arrays) has its value replaced, so `{ oauth: { token } }`
-// can't leak. Primitives are returned as-is.
-// `path` holds the objects on the current branch: a back-reference to an
-// ancestor becomes `[circular]` instead of recursing until the stack blows,
-// while the walk descends everywhere else so a cycle can't skip the redaction
-// of the keys inside it. A value shared between siblings is not a cycle.
+// Recursively redact: a sensitive *key* anywhere in the tree has its value
+// replaced, so `{ oauth: { token } }` can't leak. `path` is the current branch
+// only (add/delete around the descent), so a back-reference to an ancestor
+// becomes `[circular]` while a value shared between siblings is still walked.
+// Functions are dropped: an own enumerable `toJSON` copied into the output
+// would be invoked by `JSON.stringify` and hand back the unredacted original.
 const redactValue = (value: unknown, path: WeakSet<object> = new WeakSet()): unknown => {
-  // Functions never survive the walk: an own enumerable `toJSON` copied into
-  // the output would be invoked by `JSON.stringify` and hand back the original,
-  // unredacted object — a redaction bypass the caller controls.
   if (typeof value === 'function') return '[function]';
   if (!value || typeof value !== 'object') return value;
   if (path.has(value)) return '[circular]';
@@ -79,6 +75,18 @@ const redactValue = (value: unknown, path: WeakSet<object> = new WeakSet()): unk
       );
   path.delete(value);
   return out;
+};
+
+// Total over anything a caller can hand us: redaction reads objects that may
+// fight back (a throwing getter, an exotic proxy), and losing the fields beats
+// taking the caller down with them.
+const sanitise = (fields?: Fields): Fields => {
+  if (!fields) return {};
+  try {
+    return redactValue(fields) as Fields;
+  } catch {
+    return { fields: '[unredactable]' };
+  }
 };
 
 const renderValue = (value: unknown): string => {
@@ -115,21 +123,12 @@ export const createLogger = (level: LogLevel): Logger => {
   const emit = (lvl: LogLevel, event: string, fields?: Fields): void => {
     if (SEVERITY[lvl] < min) return;
 
-    // Redaction reads caller-supplied objects, which may fight back (a getter
-    // that throws, an exotic proxy). Losing the fields is acceptable; taking
-    // the caller down with them is not.
-    let safe: Fields;
-    try {
-      safe = fields ? (redactValue(fields) as Fields) : {};
-    } catch {
-      safe = { fields: '[unredactable]' };
-    }
+    const safe = sanitise(fields);
 
     try {
       console.error(formatLine(lvl, event, safe));
     } catch {
-      // Same contract as the MCP forward below: the stderr sink is
-      // best-effort and must never break the flow that emitted the log.
+      // Best-effort sink: a dead stderr must not break the caller.
     }
 
     if (server) {
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -56,14 +56,10 @@ const isSensitive = (key: string): boolean =>
 // Recursively redact: a sensitive *key* anywhere in the tree (top-level or
 // nested in objects/arrays) has its value replaced, so `{ oauth: { token } }`
 // can't leak. Primitives are returned as-is.
-//
-// `path` holds the objects on the current branch, so a back-reference to an
-// ancestor becomes `[circular]` instead of recursing until the stack blows
-// (Node errors, HTTP request/response pairs and socket handles all
-// self-reference). The walk still descends everywhere else — a cycle must
-// never become an escape hatch that skips redaction of the keys inside it —
-// and only true cycles are marked, not values merely shared between siblings.
-// The result is therefore always a finite tree that `JSON.stringify` accepts.
+// `path` holds the objects on the current branch: a back-reference to an
+// ancestor becomes `[circular]` instead of recursing until the stack blows,
+// while the walk descends everywhere else so a cycle can't skip the redaction
+// of the keys inside it. A value shared between siblings is not a cycle.
 const redactValue = (value: unknown, path: WeakSet<object> = new WeakSet()): unknown => {
   if (!value || typeof value !== 'object') return value;
   if (path.has(value)) return '[circular]';
--- a/src/utils/logger.ts
+++ b/src/utils/logger.ts
@@ -61,6 +61,10 @@ const isSensitive = (key: string): boolean =>
 // while the walk descends everywhere else so a cycle can't skip the redaction
 // of the keys inside it. A value shared between siblings is not a cycle.
 const redactValue = (value: unknown, path: WeakSet<object> = new WeakSet()): unknown => {
+  // Functions never survive the walk: an own enumerable `toJSON` copied into
+  // the output would be invoked by `JSON.stringify` and hand back the original,
+  // unredacted object — a redaction bypass the caller controls.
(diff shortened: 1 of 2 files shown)


src/utils/logger.ts
```diff
@@ -56,16 +56,29 @@ const isSensitive = (key: string): boolean =>
...
-  if (value && typeof value === 'object') {
-    const out: Record<string, unknown> = {};
-    for (const [key, val] of Object.entries(value as Record<string, unknown>)) {
-      out[key] = isSensitive(key) ? '[REDACTED]' : redactValue(val);
-    }
-    return out;
-  }
-  return value;
+//
+// `path` holds the objects on the current branch, so a back-reference to an
```
dlecan: Nitpick: this comment is too long

dlecan: Trimmed in 3336f08 — 8 lines down to 4. Kept only what the code can't show: why the walk carries a branch set, and that sibling sharing isn't a cycle (the DAG test pins that). The `JSON.stringify`/finite-tree consequence and the list of self-referencing value types are gone from the source; they live in the PR description if anyone needs the archaeology.

---

src/utils/logger.ts
```diff
@@ -56,16 +56,25 @@ const isSensitive = (key: string): boolean =>
...
+
+  path.add(value);
+  const out = Array.isArray(value)
+    ? value.map((item) => redactValue(item, path))
+    : Object.fromEntries(
+        Object.entries(value as Record<string, unknown>).map(([key, val]) => [
+          key,
+          isSensitive(key) ? '[REDACTED]' : redactValue(val, path),
+        ]),
+      );
```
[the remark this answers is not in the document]
dlecan: Confirmed and fixed in f07963f — reproduced before touching anything:

```
entries:      [ 'token', 'toJSON' ]
stringified:  {"hostile":{"token":"tojson-secret"}}
```

The walk does replace `token` with `[REDACTED]`, and then the copied `toJSON` hands the original object straight back at stringify time — on both sinks, since the MCP transport serialises the payload too.

I took a broader fix than the suggested diff: `typeof value === 'function'` → `'[function]'` at the top of `redactValue`, rather than special-casing `key === 'toJSON'`. Reasons:

- it closes the class instead of one spelling — the key-specific guard leaves a function reachable through an array (`items: [fn]` never goes through the object branch);
- nothing is lost: `JSON.stringify` already rendered a function value as `undefined`, so `[function]` is strictly more informative than what the line showed before;
- it avoids the nested ternary in the hot path.

Marker is `[function]`, not `[unserializable]` — the latter is `renderValue`'s vocabulary for values that *failed* to serialise, and reusing it here would blur two different outcomes.

Regression test added covering both sinks (nested `token` + malicious `toJSON`, asserting the secret appears in neither the stderr line nor `JSON.stringify` of the MCP payload). 600 tests green, `pnpm check` / `typecheck` clean.

Worth noting for the record: this predates the PR — the previous `redactValue` copied functions through the same way. I fixed it here rather than splitting it out because it's a redaction bypass in the exact function this PR rewrites, under the same hostile-caller threat model as the throwing getter already handled.

---

PR #215, opened 2026-08-05
chore(lint): check the whole repository, and gate Biome config drift

### LINKED ISSUE #202: pnpm check does not lint the root config files
## What

`pnpm check` and the lefthook pre-commit hook are both scoped to `src/ tests/ scripts/`:

```json
"check": "node scripts/biome.mjs check --error-on-warnings src/ tests/ scripts/",
```

So **no file at the repo root is ever linted by the repo's own gate** — including `vitest.config.ts`, `tsdown.config.ts`, `biome.json` itself, and `server.json`.

## How it surfaced

In #197, `stryker.config.mjs` was added at the root and tripped `style/noDefaultExport` — which native Stryker ESM configs require. Neither `pnpm check` nor pre-commit saw it; it was reported by the review bot, which lints the whole repository:

```
$ node scripts/biome.mjs check --error-on-warnings stryker.config.mjs
  × Avoid default exports.
  > 5 │ export default {
```

The immediate fix there was to extend the existing `noDefaultExport` exemption to `**/*.config.mjs`, alongside the `**/*.config.ts` / `**/*.config.mts` entries `biome.json` already carried. That those two entries exist at all is the tell: the same class of finding has come up before and been suppressed per-extension, without anyone noticing that the files are unlinted either way.

## Why it is worth fixing

The exemptions are load-bearing for files nothing checks, so nobody knows whether they are still needed or still sufficient. And a genuine finding in a root config — a real bug, not a style rule — currently reaches us only via a third-party bot, or not at all.

## Suggested direction

Widen the scope, roughly:

```json
"check": "node scripts/biome.mjs check --error-on-warnings src/ tests/ scripts/ *.ts *.mjs *.json"
```

with the same widening in `lefthook.yml`'s glob, so pre-commit and CI stay identical — that parity is deliberate per [`docs/decisions/lint-tooling.md`](docs/decisions/lint-tooling.md) §6.2 ("CI is a verification step rather than a discovery step").

Worth checking while doing it:

- **Whether the tree is already clean** for the newly-included files. `AGENTS.md` requires a rule to be green before it is enabled, and the same logic applies to widening a scope: it must not land red. `stryker.config.mjs` is clean today; the others are unverified.
- **`package.json` and `pnpm-lock.yaml`.** Biome formats JSON, and its opinion may not match what pnpm and semantic-release write back. The lockfile at minimum should stay excluded.
- **`vcs.useIgnoreFile`.** Biome already honours `.gitignore`, so `dist/`, `coverage/` and `reports/` stay out on their own — but confirm rather than assume.
- **The `PostToolUse` hook** in `.claude/settings.json` already lints whatever file was edited, root configs included. It is the only thing that currently checks them, which is why the finding above appeared during editing and not in CI. Widening `pnpm check` makes that consistent rather than incidental.

## Not urgent

Nothing is broken today. This is a gap in coverage of the linter itself, and it wants its own PR rather than riding along with unrelated work.

dlecan:
## Summary
Widens Biome's check perimeter from `src/ tests/ scripts/` to the whole repository, and adds a CI gate that catches Biome **configuration** drift the same way `pnpm check` already catches formatter and lint drift. Also applies the already-pending `rules.recommended` → `rules.preset` migration while it is still a rename.

## Linked issue
Closes #202

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Tooling / CI

## Author checklist
- [x] `pnpm test` passes locally — 32 files, 600 tests
- [x] `pnpm test:coverage` meets the thresholds — statements 96.81%, branches 83.42%, functions 98.54%, lines 97.48%
- [x] `pnpm check` is clean (lint + format) — 94 files, up from 83
- [x] `pnpm typecheck` passes
- [ ] I ran a Claude Code review on the diff and addressed its findings — **not run**; see reviewer notes
- [ ] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md` — n/a
- [x] If the change has **MCP-runtime-observable** behaviour: n/a — pure tooling, exempt. A running server behaves identically for a client; the standard gates plus CI are the validation.

## What changed

**1. Perimeter (`package.json`, `lefthook.yml`)**

`pnpm check` / `check:fix` now pass `.` instead of three paths, and the lefthook glob spans the repository. `biome.json` `files.includes` becomes the single source of truth on both sides, so pre-commit and CI cannot disagree — previously two path lists had to be kept in sync by hand.

Eleven files were neither linted nor format-checked:

```
.claude/settings.json   biome.json      renovate.json    tsdown.config.ts
.mcp.json               glama.json      server.json      vitest.config.ts
.releaserc.json         package.json    tsconfig.json
```

That includes `tsdown.config.ts` and `vitest.config.ts` — the only two TypeScript files in the tree that no lint rule ever reached. The existing `**/*.config.ts` → `noDefaultExport: off` override in `biome.json` already anticipated them, and they pass unmodified.

The tree is **already clean** for the wider perimeter, so this commit turns the coverage on without a reformatting pass: `Checked 94 files. No fixes applied.`

**2. Config drift gate (`.github/workflows/ci.yml`)**

`pnpm check` runs Biome in check mode, so an upgrade that changes formatter or lint output fails on the bump PR itself rather than surfacing later in an unrelated contributor's diff. Configuration had no equivalent: a deprecated field is reported as an `info` diagnostic, which `--error-on-warnings` does not fail on, so it slides through silently until the next major turns it into a hard error.

`biome migrate` has no check mode — it exits 0 whether or not a migration is pending — so the step writes and diffs. On the passing path nothing is written.

**3. `rules.recommended` → `rules.preset` (`biome.json`)**

Biome 2.5.6 already reports the old field as deprecated. `biome migrate --write` produces a one-line diff (verified: it does not reformat the config), and re-running it is idempotent.

**4. Docs (`AGENTS.md`, `docs/decisions/lint-tooling.md`, `docs/decisions/biome-rules.md`)**

Two new durable rules in `AGENTS.md` (don't reintroduce a path list; keep the migrate gate). The lint-tooling and biome-rules decision records are corrected where they described the *current* perimeter; the historical benchmark and diagnostic-count measurements are left intact, with a clause noting the scope they were taken at.

- **#202 was found late, and its suggested direction was not followed.** The issue proposed `check ... src/ tests/ scripts/ *.ts *.mjs *.json` — a path list plus an extension list. Both were tried here and both drifted: the path list is what left the root configs unchecked in the first place, and the extension list omitted `.mts`, `.tsx`, `.cts`, `.jsx` and `.css`, which Biome parses (`.mts` notably, since `biome.json` carries an override for it). Passing `.` and dropping lefthook's `glob` leaves `biome.json` `files.includes` as the single perimeter instead. The issue also asked whether the tree was already clean for the newly-included files — it is, so no reformatting pass was needed.
- **Deliberately not in scope.** Biome cannot handle YAML, Markdown or shell, so 58 files (`.github/workflows/*.yml`, `lefthook.yml`, all of `docs/`, `.claude/hooks/session-start.sh`, …) stay outside any checker. Adding `actionlint` / `markdownlint` / `shellcheck` was considered and rejected: without an auto-formatter for those formats there is no drift-on-upgrade risk, which is the problem this PR addresses.
- **Generated files now in the perimeter.** `server.json` is produced by `pnpm build:server-json` and rewritten by semantic-release. Verified that a freshly regenerated `server.json` is byte-identical to the committed one *and* already conforms to Biome's JSON formatter, so there is no generator-vs-formatter conflict that could turn `main` red after a release. Same for `package.json` (semantic-release only touches `version`).
- **Cost.** Negligible: 3.1 s → 3.3 s wall for `pnpm check` locally. The scan was already repository-wide because of `files.includes: ["**"]`; only the number of files linted changed (83 → 94). The `!!**/node_modules` force-ignore stays essential and is untouched.
- **Gate verified by construction, not just by a green run.** Snapshot `biome.json`, run `migrate --write`, diff → no change. Before the `preset` migration the same sequence produced the one-line diff, i.e. the gate does fail when it should.
- **Pre-commit glob verified for real.** The commit's own pre-commit run reported `Checked 2 files` (`biome.json`, `package.json`); under the previous glob the root-level pattern matched nothing.
- **Checklist honesty.** I did not run `/code-review` on this diff. It is 7 files of tooling configuration with no `src/` changes; happy to run it if you'd rather have it before merge.
- **Pre-existing staleness I left alone**, since it is unrelated to this diff and would muddy it: `docs/decisions/biome-rules.md` and `docs/decisions/lint-tooling.md` still say "Biome 2.5.5" after the 2.5.6 bump in #204.

https://claude.ai/code/session_012pudHuJza4TLT7oTpP98pF

---
_Generated by [Claude Code](https://claude.ai/code/session_012pudHuJza4TLT7oTpP98pF)_

## Summary by CodeRabbit

* **Chores**
  * Expanded code-quality checks to cover the entire repository.
  * Updated formatting and linting commands to use centralized file-selection rules.
  * Added CI verification for automatic Biome configuration migrations.

* **Documentation**
  * Clarified repository-wide checking, pre-commit behavior, benchmarking, and troubleshooting guidance.
  * Updated Biome configuration to use the current recommended preset format.

---


AGENTS.md
```diff
@@ -109,6 +109,19 @@ test: if a running server behaves no differently for a client, it's tooling.
 ## Code style
 
 - TypeScript strict; Biome for lint/format (`pnpm check`).
+- Biome's perimeter is the **whole repository**, and `biome.json`
```
dlecan: Nitpick: agents.md must stay concise
Relive all of this, it's useless for future use of the project

lefthook.yml
```diff
@@ -10,12 +10,19 @@
...
   jobs:
     - name: biome
-      # Same paths as `pnpm check`, so pre-commit and CI cannot disagree.
-      # Two patterns on purpose: `**/` requires at least one directory level in
-      # lefthook's matcher, so it alone would silently skip src/index.ts,
-      # src/config.ts, tests/setup.ts and every other top-level file.
-      glob:
-        - '{src,tests,scripts}/*.{ts,mjs,cjs,js,json,jsonc}'
-        - '{src,tests,scripts}/**/*.{ts,mjs,cjs,js,json,jsonc}'
+      # No `glob` on purpose: every staged file goes to Biome and Biome decides,
```
dlecan: Nitpick: make a simple one line comment

.github/workflows/ci.yml
```diff
@@ -26,6 +26,19 @@ jobs:
 
       - run: pnpm check
 
+      # A Biome upgrade can deprecate configuration as well as change lint or
```
dlecan: suggestion: shorter comment

AGENTS.md
dlecan: Done in 905e8fb, across all three threads: one line in `lefthook.yml`, two in `ci.yml`, and here two bullets collapsed into one.

The detail is relocated rather than deleted — `docs/decisions/lint-tooling.md` now carries the extension list that had drifted (`.mts`, `.tsx`, `.cts`, `.jsx`, `.css`), the `Checked 0 files` cost of passing `{staged_files}` unfiltered, and why CI needs a second Biome step. What stays in `AGENTS.md` is only the part an agent has to obey: don't filter the perimeter on either side.

---

AGENTS.md
```diff
@@ -109,6 +109,12 @@ test: if a running server behaves no differently for a client, it's tooling.
 ## Code style
 
 - TypeScript strict; Biome for lint/format (`pnpm check`).
+- Biome's perimeter is the whole repository and `biome.json` `files.includes` is
```
dlecan: Nitpick: remove that, that's internal details useless permanently in your context
Reading the code will tell you that information if you need it one day

AGENTS.md
dlecan: Removed in f0dc2ce — `AGENTS.md` is now byte-identical to `main`, so this PR no longer touches it at all.

For the record on where the rule went, since the code shows the *what* but not the *don't*: `docs/decisions/lint-tooling.md` states that the glob was removed rather than widened and that widening would only have moved the drift. `lefthook.yml`'s one-liner points there, as does the neighbouring `AGENTS.md` bullet on the per-edit/pre-commit split.

---

AGENTS.md
dlecan: Not applying this one — the guideline is about tool counts, not about measurements.

The rule in `AGENTS.md` lives in the **Documentation maintenance** section and reads: "Prose and section headers are deliberately count-free (no per-section `(N tools)` or global tool totals)". Its stated load-bearing exceptions are namespace counts in `tests/unit/routing/registry.test.ts` and the length assertions in `tests/unit/tools/*.test.ts` — both about the tool surface. It exists so the tool tables and the README don't carry `(N tools)` labels that go stale when a tool is added.

Reading it as "no numbers anywhere under `docs/`" would gut the two files flagged here, because measurement *is* what they are:

- `lint-tooling.md` opens with "Everything below is measured on this repository, not quoted from vendor benchmarks", carries a measurement-provenance warning about which tables are comparable, and its entire argument rests on the timings (`458.5 ms` → `137.0 ms`, the `3.3×`). The two figures added in this PR sit in that same stage table and are the same kind of fact.
- `biome-rules.md` states outright: "Counts in this document are that measurement, not an estimate."

The duplication trace in the glob explanation (`root.json src/deep/nested.ts src/top.ts top.ts src/deep/nested.ts src/top.ts`) is there for the same reason: it's the evidence that lefthook's `*` crosses `/`, which is why the glob was removed rather than widened. Without it the paragraph is an assertion.

The sibling finding on this review was real and is fixed in 05377b8.

---

PR #218, opened 2026-08-06
fix(config): fail loudly on flags and env vars with a missing or empty value

### LINKED ISSUE #174: CLI parser silently ignores value-taking flags with a missing or empty value
## Problem

Every value-taking flag branch in `parseCliArgs` (`src/config.ts`) is guarded by `arg === '--flag' && next`. When the value is missing or empty, the guard is falsy and the flag is **silently dropped** instead of failing at startup. This affects all value-taking flags equally: `--mode`, `--namespace`, `--tool`, `--log-level`, `--transport`, `--host`, `--port`, `--public-url`, `--cors-origin`, `--callback-port` and (once #173 lands) `--hc-resource-scheme`.

Two concrete failure shapes:

1. **Flag as last argument (value forgotten).** E.g. a container CMD edited to `["mycompany", "--hc-resource-scheme"]`: the bare flag matches no branch and is discarded, the server boots with the default scheme, and nothing in the logs points at the cause. Clients and runbooks expecting `wiki://topology` see `zendesk-hc://topology` with zero diagnostic.

2. **Empty value (shell expansion).** E.g. `zendesk-mcp-server --hc-resource-scheme "$SCHEME" mycompany` with `SCHEME` unset: `next` is `''` (falsy), the flag is ignored, and the empty string then falls through to the positional-subdomain branch — it becomes the subdomain, `mycompany` is ignored (`positionalIndex` already consumed), and startup fails with the misleading `ZENDESK_SUBDOMAIN is required` instead of naming the actually-broken flag.

This is inconsistent with the parser's own posture elsewhere: an *invalid* value fails loudly at config parse time (strict `--port` parsing, the strict scheme validation from #169), but a *missing* value is silently swallowed. Surfaced during the review pass on PR #173 (see "Notes for the reviewer"); left out of that PR because it is a cross-cutting parser change, not specific to the new flag.

## Proposed solution

Make `parseCliArgs` fail loudly, for the whole class at once:

- A value-taking flag at the end of `argv`, or followed by an empty string, throws a clear ASCII error naming the flag (`Missing value for --hc-resource-scheme.`), consistent with the existing no-echo policy for values (`parsePort` comment).
- An empty value stops falling through to the positional-subdomain branch (fixing the misleading `ZENDESK_SUBDOMAIN is required` shape).
- Ideally driven by a single list/table of value-taking flags rather than one `else if` per flag, so the fix cannot drift per-flag — and that list could then be exported and reused by `scripts/mcp-live.ts` `VALUE_FLAGS`, whose "Keep in sync with `parseCliArgs`" comment documents exactly this drift hazard (it was found out of sync by six flags in #173).

## Alternatives considered

- **Also reject unknown `--*` flags** (typos like `--hc-ressource-scheme` are silently ignored today too). Same "fail loudly at startup" philosophy, but a behavior change with more blast radius (e.g. forward-compat of deploy scripts across versions); could be a follow-up or an explicit non-goal — to be decided in this issue.
- **Keep the silent-ignore and only document it.** Zero code change, but the misconfiguration shapes above stay undiagnosable from logs, which is the actual pain.

## Decision — agreed scope

Settling the open question above, and widening this issue to the class it belongs to. The whole of the following is in scope for #174 and is implemented in PR #218:

| Behaviour | Verdict |
| --- | --- |
| A value-taking flag with a missing or empty value fails at startup, naming the flag | in scope, the original ask |
| **Unknown flags** are rejected (`--moed all`) | **in scope.** The "Alternatives considered" question above is decided *yes*: this is the same failure the issue objects to, reached by a typo. `--hc-ressource-scheme wiki mycompany` boots against tenant `wiki` and drops `mycompany` — a wrong Zendesk tenant, silently |
| **Empty single-value environment variables** are rejected, naming the variable | **in scope.** Failure shape 2 above is a shell handing over `''` for an unset variable; the identical mistake in a compose file (`HC_RESOURCE_SCHEME=`) produced the identical silent misconfiguration. Fixing only the CLI half would leave the more common deployment path broken. Two exceptions: `CORS_ORIGIN`, where an empty *list* legitimately means "no extra origins", and any variable the command line overrides, which is never consulted |
| **Extra positional arguments** are rejected | **in scope.** Found while reviewing #218: `--namespace tickets help_center mycompany` booted against subdomain `help_center` with only `tickets` registered. The last silently-dropped argv shape, and the same wrong-tenant outcome |
| A value given to a **standalone flag** is rejected (`--read-only=false`) | in scope, and free from the same strict parser |

One consequence worth stating separately, because it reverses an existing spec: an empty `HC_RESOURCE_SCHEME` used to mean "unset" and applied the default, matching the PORT-style convention. That convention is what hid the broken deployment described above, so it is now a startup error, and the test asserting the old behaviour is inverted.

## Scope

- [x] Read-only (no data is modified in Zendesk)

## Additional context

- Touch-points: `src/config.ts` (`parseCliArgs`), `tests/unit/config.test.ts`, `scripts/mcp-live.ts` (`VALUE_FLAGS`, if the shared-list option is taken).
- Purely a startup/config concern: no MCP-runtime-observable behavior changes for a correctly configured server (a misconfigured one now fails at startup instead of running with silently-wrong config — that startup failure is the feature).
- Split out of the review discussion on #173 / #169.
- Implementation went through `node:util` `parseArgs` in `strict` mode rather than a hand-rolled parser or a third-party one; commander and convict were both tried and both accept an empty CLI value and an empty env var, reproducing this defect. Rationale in PR #218.

dlecan:
## Summary

A value-taking flag with a missing or empty value was dropped silently, so the server booted with defaults and nothing pointed at the cause. Detection now goes through `node:util` `parseArgs` driven by one declarative flag table, which also rejects unknown flags, a flag whose value is another flag, and a stray second positional; empty single-value environment variables fail the same way. A correctly configured server is unaffected — a misconfigured one refuses to boot instead of running with silently-wrong config.

## Linked issue

Closes #174

## Scope — wider than the issue, and approved

Issue #174 asks for one thing: a value-taking **CLI flag** whose value is missing or empty must fail instead of being dropped. This PR does that, and three more things:

| Beyond the issue | Status in #174 | Why it is here |
| --- | --- | --- |
| Rejecting **unknown flags** (`--moed`) | listed under *Alternatives considered*, explicitly left "to be decided in this issue" | `parseArgs` in `strict` mode gives this for free; opting out would mean deliberately re-allowing typos. Same failure shape as the issue's own case 2 — `--hc-ressource-scheme wiki mycompany` boots against tenant `wiki` |
| Rejecting **empty environment variables** | **not mentioned at all** | The issue's case 2 is a shell handing over `''` for an unset variable. The identical mistake in a compose file (`HC_RESOURCE_SCHEME=`) produced the identical silent misconfiguration, so fixing only the CLI half would have left the more common deployment path broken |
| Rejecting **extra positionals** | not mentioned | Found by review of this PR: `--namespace tickets help_center mycompany` booted against subdomain `help_center` with only `tickets` registered. The last silently-dropped argv shape, and the same wrong-tenant outcome the issue objects to |
| Rejecting a value given to a **standalone flag** (`--read-only=false`) | not mentioned | Also free from `strict` mode |

The maintainer has confirmed the environment-variable rule is wanted, and with it the rest of the widened scope. Issue #174 now records that decision in its own body, so the four rows above are the agreed shape of this change and nothing is being split out.

## Type of change
- [x] Bug fix
- [ ] New feature
- [x] Documentation
- [ ] Tooling / CI

## Author checklist
- [x] If this PR resolves an issue, its description above links it with a closing keyword (`Closes #`) so it auto-closes on merge to `main`
- [x] `pnpm typecheck` passes
- [ ] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md` — n/a, no tool added
- [x] If the change has **MCP-runtime-observable** behaviour: this PR carries a `## Functional validation plan` — see the scope note in it, the negative shapes are not reachable from the validator's environment

## What was wrong

Every value-taking branch was guarded by a truthiness check on the next argument, so a missing or empty value made the guard falsy and the flag was discarded. Verified against `loadConfig` before the fix:

| argv | before |
| --- | --- |
| `["mycompany","--hc-resource-scheme"]` | boots on the default scheme, zero diagnostic |
| `["--hc-resource-scheme","","mycompany"]` | `''` becomes the positional subdomain, `mycompany` is dropped, startup fails with `ZENDESK_SUBDOMAIN is required` — naming the wrong knob |

Three further shapes surfaced while fixing it, none of them in the issue:

- `["mycompany","--host","--read-only"]` yielded `host === '--read-only'` with `readOnly: false`. On stdio `host` is ignored entirely, so the **only** effect was a server meant to be read-only exposing its write tools, silently.
- `["--hc-ressource-scheme","wiki","mycompany"]` booted against subdomain `wiki`: the typo'd flag was skipped, `wiki` taken as the positional, `mycompany` dropped. A wrong Zendesk tenant, not a wrong flag.
- `["--namespace","tickets","help_center","mycompany"]` booted against subdomain `help_center` with only the `tickets` namespace registered and `mycompany` dropped — the wrong tenant *and* a narrowed tool surface at once. A space-separated list is the natural wrong guess for a `multiple: true` flag, and extra positionals were the last silently-dropped argv shape left standing.

## What changed

`parseArgs` in `strict` mode rejects the unknown-flag, missing-value, flag-as-value and value-to-a-standalone-flag shapes. Two rules are ours, because they are the two `parseArgs` accepts: an empty value, and more than one positional. Node's messages already name the offending flag and never echo the value after an `=`, so they pass through rather than being re-wrapped; the `Empty ...` and `Expected one positional ...` messages are ours and stable. Neither of ours echoes a value — the positional one reports a count.

Empty single-value environment variables now fail naming the variable, across the nine `loadConfig` reads. Two deliberate exceptions, both documented in `docs/configuration.md`:

- `CORS_ORIGIN` stays tolerant — as a list, an empty value legitimately means "no extra origins".
- A variable the command line overrides is never consulted, so `--port 8080` next to a stray `PORT=` still boots, and so does the positional `<subdomain>` next to an empty `ZENDESK_SUBDOMAIN`. Validation applies to the value the server actually uses.

The value-flag list is now derived once and consumed by `scripts/mcp-live.ts`, replacing a hand-maintained copy under a "Keep in sync" comment. One correction to the premise, since #174 cites it: the issue notes that copy "was found out of sync by six flags in #173", which was true then but is no longer — on `1b669c4` its 11 entries match the parser's 11 value-taking flags exactly, verified by diffing the two lists. Deriving it removes the standing invitation to drift; it is not repairing a live defect.

## Notes for the reviewer

**One deliberate spec change, and it rewrites an existing test.** `tests/unit/config.test.ts` used to assert that an empty `HC_RESOURCE_SCHEME` meant "unset" and applied the default, matching the PORT-style convention. That convention is what hid the broken deployment in the issue, so the test now asserts rejection. It is not a test weakened to go green; it is a test inverted because the spec it encoded is the bug. This follows from the same decision that keeps the environment-variable rule in scope, but it is the single most reversible thing here if you want the old convention back.

**A third-party parser was evaluated and rejected, empirically.** commander 15 and convict 6 were both installed and exercised against the real shapes. Both **accept an empty CLI value and an empty env var** — they reproduce this exact defect rather than fixing it — and commander additionally accepts `--host --read-only`, which the built-in rejects. convict also echoes offending values into its messages (`value was "nope"`), contradicting the no-echo policy documented in `src/config.ts`, and brings its own schema language alongside `ConfigSchema`, which that file designates as the single authority. `node:util` `parseArgs` is a built-in, stable since Node 20.0.0, matching `engines: node>=20` and the `smoke-node20` job. `allowNegative` is deliberately not used (Node 22.4+); `--no-topology` and `--no-promoted-articles` stay declared as their own booleans.

**The `Map`-not-object-literal comment was removed with the maps it explained.** That guard existed because the old parser indexed raw argv against a lookup. `parseArgs` rejects `--toString` / `--__proto__` / `--constructor` as unknown options, returns a null-prototype `values` object, and collects positionals into an array that is never a lookup key — a bare `toString` argument correctly becomes the subdomain, and there is a test for it.

**`parseCliArgs` builds its result with a local mutable accumulator, and two attempts to remove that were backed out.** Both collide with lint rules this repo has enabled, so the "cleaner" versions cannot land:

- Explicit per-flag assignments (which would also restore the compile-time checking that `Object.assign(result, { [field]: value })` loses to a computed key): 15 straight-line `if`s put the function at cognitive complexity 22 against the ceiling of 15 in `noExcessiveCognitiveComplexity`.
- A `reduce` spreading the accumulator, as CodeRabbit suggested: two `noAccumulatingSpread` errors, and `pnpm check` runs `--error-on-warnings`.

The function is pure either way — same argv in, same `CliResult` out, and the accumulator never escapes — so what is left is a style preference the lint config has already settled. A comment at the accumulator records this. The computed-key gap is covered behaviourally instead: every `FIELD_BY_FLAG` entry is asserted by a test, and the mutation gate kills all string-literal mutants on those lines.

**Mutation gate drove three simplifications.** The first run left three survivors. None was a missing assertion — each marked a branch that could not affect the outcome, so they were removed rather than annotated with `Stryker disable`: a redundant `undefined` guard, a guard that only satisfied `exactOptionalPropertyTypes`, and an `oauthClientId` fallback the schema made unreachable. That last one cuts the failure report for a missing subdomain from two issues to the one the operator can act on.

### Scope note — read first

This change is a **startup/config** concern: every rejection happens in `loadConfig`, before any transport is connected. That collides with the validator profile this repo's `run-validation-plan` skill assumes — a session already on the branch **with the server running and authenticated**, forbidden from building or starting it. A server that refuses to boot cannot be observed from a session whose server is already up.

So the negative shapes below are marked **not validator-reachable**, honestly rather than dressed up as tool calls. They are covered by the unit tests in `tests/unit/config.test.ts` (parametrised over all 11 value-taking flags), a 100% mutation score on every changed line, and direct probes of the built binary recorded under "Implementer evidence". What the validator *can* and should confirm is the claim that actually touches a running server: **nothing changed for a correctly configured one.**

### Context

The running server in the validator's environment was started with valid config, so it exercises exactly the regression-risk path: the rewritten parser produced the config it is running on. `--mode all` means the full tool surface is registered, which is the most sensitive observable — a mis-mapped flag would show up as a wrong surface or a missing resource.

Not exercised, and not worth testing: per-tool behaviour, Zendesk API shapes, auth flows. This PR touches none of them.

### Setup & prerequisites

None beyond the ready environment. No state to seed, no probe script, no credentials beyond what is already wired in. Capture `git rev-parse HEAD` and report the SHA so the verdict is anchored to a commit.

### Scenarios

| id | Validates | Action | Expected observable |
| --- | --- | --- | --- |
| V1 | The rewritten parser produced a correct config for the running server | `tools/list` | The full `--mode all` surface. On this branch that is 52 tools: `help_center` 29, `tickets` 18, `users` 5. Compare against `createAllTools` on the commit you actually test rather than against that total, which moves whenever tools are added — what matters is that no namespace is missing and the count is not short |
| V2 | `--hc-resource-scheme` default survives the rewrite | `resources/list` | The topology resource is exposed as `zendesk-hc://topology` — the default scheme, not an empty or mangled one |
| V3 | The topology resource still reads | Read `zendesk-hc://topology` | Returns the locales / category-section tree normally. Confirms the scheme round-trips through the SDK's registry lookup, the thing an empty scheme would have broken |
| V4 | A real read path is unaffected | `get_current_user` (or any cheap read) | Succeeds as before. Any failure here is unrelated to this PR but worth reporting |

V1 through V3 are the priority and take one call each. V4 is a sanity check.

### Not validator-reachable — for the record

These are the PR's actual subject. They need a server launched with deliberately broken config, which the validator profile excludes. Listed so a human reviewer can run them by hand from a checkout if they want independent confirmation:

| Invocation | Expected |
| --- | --- |
| `node dist/index.js mycompany --hc-resource-scheme` | `option '--hc-resource-scheme <value>' argument missing` |
| `node dist/index.js --hc-resource-scheme "" mycompany` | `Empty value for --hc-resource-scheme.` — and **not** `ZENDESK_SUBDOMAIN is required` |
| `node dist/index.js mycompany --host --read-only` | `Option '--host' argument is ambiguous` |
| `node dist/index.js --hc-ressource-scheme wiki mycompany` | `Unknown option '--hc-ressource-scheme'` |
| `node dist/index.js mycompany --read-only=false` | `Option '--read-only' does not take an argument` |
| `node dist/index.js --namespace tickets help_center mycompany` | `Expected one positional argument (the subdomain), got 2.` — and the offending values are **not** in the message |
| `HC_RESOURCE_SCHEME= node dist/index.js mycompany` | `Empty HC_RESOURCE_SCHEME. Set it to a value, or unset it entirely.` |
| `ZENDESK_SUBDOMAIN= node dist/index.js mycompany` | boots — the positional overrides the variable, so the empty value is never consulted |
| `node dist/index.js mycompany --oauth-token=s3cr3t` | `Unknown option '--oauth-token'` — the value is **not** in the message |
| `node dist/index.js mycompany --mode all --read-only --hc-resource-scheme wiki` | boots normally |

### Long-running behaviours

None. Nothing in this change is time-based, so there is no timer to simulate.

### Implementer evidence

All rows above were run against the built binary on this branch and produced the expected output, including the two no-echo cases. Gates on the current head, with `main` merged in: `pnpm typecheck` and `pnpm check` clean, `pnpm test` 768 passing, `pnpm test:coverage` at the thresholds with `config.ts` at 100% branch coverage, `pnpm test:mutation:diff origin/main HEAD` 124/124 mutants killed on the changed lines. `pnpm tsx scripts/mcp-live.ts list -- --mode namespace` was run to exercise the shared `VALUE_FLAG_NAMES` import on its drift-sensitive path. CI is green.

The test and mutant totals rose from 725 and 100 as `main` was merged in: the branch now carries the Help Center translation tools from #224 and the release-tooling tests, none of which this PR touches.

### Reporting instructions

Post one PR comment in English: per-id verdict (`OK` / `FAIL` / `BLOCKED`) with the observed evidence — the tool count and namespace breakdown for V1, the exact resource URI for V2, a snippet of the payload for V3 — the HEAD SHA you tested, and an overall green/failing summary. If a scenario's evidence is ambiguous, say so rather than guessing a verdict.

### One environment note

The functional harness forwards `env: process.env` to the server it spawns (`tests/functional/bin/dump-tools-list.mjs`). An empty exported `PORT` or `HC_RESOURCE_SCHEME` in a shell will now fail those scenarios. That is the intended behaviour of this change, not a harness bug.

## Summary by CodeRabbit

- **New Features**
  - Added stricter startup validation for command-line arguments and environment variables.
  - Supports documented flag formats, repeatable options, positional subdomains, and CLI-over-environment precedence.
  - Provides clear, secret-safe errors for unknown, missing, empty, or extra arguments.

- **Bug Fixes**
  - Invalid configuration now fails consistently at startup instead of silently using unintended values.
  - Empty single-value environment variables are rejected, while empty `CORS_ORIGIN` remains valid.
  - Preserved fallback behavior for resource-scheme configuration.

- **Documentation**
  - Added configuration and troubleshooting guidance for validation errors, accepted syntax, and corrective actions.

## Summary by CodeRabbit

* **New Features**
  * Added stricter startup validation for command-line arguments and environment variables.
  * Added support for repeatable options and multiple accepted flag formats.
  * Improved handling of empty values, positional arguments, and configuration precedence.
  * Protected secrets from appearing in startup error messages.

* **Bug Fixes**
  * Empty configuration values now fail clearly when invalid, while valid empty CORS settings remain supported.

* **Documentation**
  * Added configuration and troubleshooting guidance for startup validation errors, accepted syntax, and corrective actions.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -48,6 +48,24 @@ Options:
 
 `--namespace` and `--read-only` are applied before the proxies are registered, so they narrow the surface in every mode — in the default `namespace` mode, `--namespace help_center` registers a single proxy (`zendesk_help_center`) instead of the full set of namespace proxies.
 
+### A malformed invocation fails at startup
+
+The server refuses to boot rather than run with config it cannot honour. Four
+shapes are rejected, each naming the flag at fault:
+
+| Invocation | Why it is rejected |
+| --- | --- |
+| `--mode` (nothing after it) | the value was forgotten |
+| `--mode ""` or `--mode=` | an empty value, typically `--mode "$VAR"` with `VAR` unset |
+| `--host --read-only` | the value is another flag, so `--host` would swallow it |
+| `--read-only=false` | a standalone flag takes no value |
+| `--moed all` | unknown flag — a typo is not silently ignored |
+
+Values are never echoed back in these messages, so a mistyped
+`--token=<secret>` reports only the flag name. Both `--flag value` and
+`--flag=value` are accepted. Only the first positional argument is read as the
+subdomain.
+
 **Examples:**
 
 ```bash
@@ -102,6 +120,20 @@ Scope and limits, by design:
   sources to reload. It is meant for `tsx`-from-source development.
 
 ## Environment variables
+
+An **empty** variable is a misconfiguration, not "unset": `PORT=` in a compose
+file, and `PORT="$VAR"` with `VAR` unset, both arrive as an empty string, and
+applying the default there would boot a server whose config silently disagrees
+with the deployment. Every single-value variable below therefore fails at
+startup when set but empty, naming the variable — unset it to get the default.
+
+Two deliberate exceptions:
+
+- `CORS_ORIGIN` is a *list*, where an empty value legitimately means "no extra
+  origins" on top of the built-in allowlist.
+- A variable that a CLI flag overrides is never consulted, so `--port 8080`
+  alongside a stray `PORT=` still boots. Validation applies to the value the
+  server actually uses.
 
 ### `ZENDESK_SUBDOMAIN`
 **Required:** yes (or the CLI `<subdomain>` argument) · **Default:** —
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -59,7 +59,7 @@ shapes are rejected, each naming the knob at fault:
 | `--mode ""` or `--mode=` | an empty value, typically `--mode "$VAR"` with `VAR` unset |
 | `--host --read-only` | the value is another flag, so `--host` would swallow it |
 | `--read-only=false` | a standalone flag takes no value |
-| `--moed all` | unknown flag — a typo is not silently ignored |
+| `--moed all` | an unknown flag. A typo is not silently ignored |
 | `mycompany extra` | a second positional argument, so one of them would be dropped |
 
 Values are never echoed back in these messages, so a mistyped
@@ -67,10 +67,10 @@ Values are never echoed back in these messages, so a mistyped
 `--flag=value` are accepted.
 
 Exactly one positional argument is read, as the subdomain. A repeatable flag
-(`--namespace`, `--tool`, `--cors-origin`) has to be repeated —
-`--namespace tickets --namespace help_center`, never
-`--namespace tickets help_center`, which would leave `help_center` sitting where
-the subdomain belongs.
+(`--namespace`, `--tool`, `--cors-origin`) has to be repeated:
+`--namespace tickets --namespace help_center`. Writing
+`--namespace tickets help_center` instead leaves `help_center` sitting where the
+subdomain belongs.
 
 **Examples:**
 
@@ -131,7 +131,7 @@ An **empty** variable is a misconfiguration, not "unset": `PORT=` in a compose
 file, and `PORT="$VAR"` with `VAR` unset, both arrive as an empty string, and
 applying the default there would boot a server whose config silently disagrees
 with the deployment. Every single-value variable below therefore fails at
-startup when set but empty, naming the variable — unset it to get the default.
+startup when set but empty, naming the variable. Unset it to get the default.
 
 Two deliberate exceptions:
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -50,8 +50,8 @@ Options:
 
 ### A malformed invocation fails at startup
 
-The server refuses to boot rather than run with config it cannot honour. Four
-shapes are rejected, each naming the flag at fault:
+The server refuses to boot rather than run with config it cannot honour. These
+shapes are rejected, each naming the knob at fault:
 
 | Invocation | Why it is rejected |
 | --- | --- |
@@ -60,11 +60,17 @@ shapes are rejected, each naming the flag at fault:
 | `--host --read-only` | the value is another flag, so `--host` would swallow it |
 | `--read-only=false` | a standalone flag takes no value |
 | `--moed all` | unknown flag — a typo is not silently ignored |
+| `mycompany extra` | a second positional argument, so one of them would be dropped |
 
 Values are never echoed back in these messages, so a mistyped
 `--token=<secret>` reports only the flag name. Both `--flag value` and
-`--flag=value` are accepted. Only the first positional argument is read as the
-subdomain.
+`--flag=value` are accepted.
+
+Exactly one positional argument is read, as the subdomain. A repeatable flag
+(`--namespace`, `--tool`, `--cors-origin`) has to be repeated —
+`--namespace tickets --namespace help_center`, never
+`--namespace tickets help_center`, which would leave `help_center` sitting where
+the subdomain belongs.
 
 **Examples:**
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -62,9 +62,10 @@ shapes are rejected, each naming the knob at fault:
 | `--moed all` | an unknown flag. A typo is not silently ignored |
 | `mycompany extra` | a second positional argument, so one of them would be dropped |
 
-Values are never echoed back in these messages, so a mistyped
-`--token=<secret>` reports only the flag name. Both `--flag value` and
-`--flag=value` are accepted.
+Values are never echoed back in these messages, so an unsupported flag written
+as `--anything=<secret>` is reported by name alone, with the value withheld. For
+the flags that take a value, both `--flag value` and `--flag=value` work;
+standalone flags such as `--read-only` accept no value in either form.
 
 Exactly one positional argument is read, as the subdomain. A repeatable flag
 (`--namespace`, `--tool`, `--cors-origin`) has to be repeated:
@@ -137,9 +138,10 @@ Two deliberate exceptions:
 
 - `CORS_ORIGIN` is a *list*, where an empty value legitimately means "no extra
   origins" on top of the built-in allowlist.
-- A variable that a CLI flag overrides is never consulted, so `--port 8080`
-  alongside a stray `PORT=` still boots. Validation applies to the value the
-  server actually uses.
+- A variable that the command line overrides is never consulted, so `--port
+  8080` alongside a stray `PORT=` still boots, and so does the positional
+  `<subdomain>` alongside an empty `ZENDESK_SUBDOMAIN`. Validation applies to
+  the value the server actually uses.
 
 ### `ZENDESK_SUBDOMAIN`
 **Required:** yes (or the CLI `<subdomain>` argument) · **Default:** none
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -8,29 +8,29 @@ variables referenced below.
 
 The server would rather not boot than run with config it cannot honour, so a
 malformed invocation fails immediately. The message names the knob at fault, and
-never repeats the value you passed — a mistyped `--token=<secret>` reports only
+never repeats the value you passed: a mistyped `--token=<secret>` reports only
 the flag name.
 
 | What you get | What to look for |
 | --- | --- |
 | `option '--mode <value>' argument missing` | the flag is the last thing on the command line; its value was forgotten |
-| `Empty value for --mode.` | the value is an empty string — usually `--mode "$VAR"` with `VAR` unset in the shell or compose file |
+| `Empty value for --mode.` | the value is an empty string, usually `--mode "$VAR"` with `VAR` unset in the shell or compose file |
 | `Option '--host' argument is ambiguous` | the next token is another flag, so `--host` would have swallowed it. To pass a value that really does start with a dash, use `--host=-value` |
 | `Unknown option '--moed'` | a typo, or a flag from a different version. Check it against the [CLI reference](configuration.md#cli-reference) |
 | `Option '--read-only' does not take an argument` | a standalone flag was given a value |
-| `Expected one positional argument (the subdomain), got 2.` | a stray argument. Most often a repeatable flag given a space-separated list — write `--namespace tickets --namespace help_center` |
+| `Expected one positional argument (the subdomain), got 2.` | a stray argument. Most often a repeatable flag given a space-separated list, where `--namespace tickets --namespace help_center` is the right form |
 | `Empty PORT. Set it to a value, or unset it entirely.` | the variable is set but empty. Remove it entirely to fall back to the default |
 
 The rows starting `option` / `Option` / `Unknown option` come from Node's own
-argument parser, so treat that wording as indicative rather than exact — it can
-change between Node releases. The `Empty …` and `Expected one positional …`
+argument parser, so treat that wording as indicative rather than exact. It can
+change between Node releases. The `Empty ...` and `Expected one positional ...`
 messages are ours and stable.
 
 Two things that are *not* errors: `CORS_ORIGIN=` is a valid way to say "no extra
 origins", and a variable that a CLI flag overrides is never read, so `--port
 8080` alongside a stray `PORT=` starts normally.
 
-If a deployment that used to work now fails here, that is the point — it was
+If a deployment that used to work now fails here, that is the point. It was
 running with config that had been silently dropped. The most common case is an
 empty `HC_RESOURCE_SCHEME`, which used to fall back to `zendesk-hc` while
 runbooks expected a branded scheme.
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -18,11 +18,13 @@ the flag name.
 | `Option '--host' argument is ambiguous` | the next token is another flag, so `--host` would have swallowed it. To pass a value that really does start with a dash, use `--host=-value` |
 | `Unknown option '--moed'` | a typo, or a flag from a different version. Check it against the [CLI reference](configuration.md#cli-reference) |
 | `Option '--read-only' does not take an argument` | a standalone flag was given a value |
+| `Expected one positional argument (the subdomain), got 2.` | a stray argument. Most often a repeatable flag given a space-separated list — write `--namespace tickets --namespace help_center` |
 | `Empty PORT. Set it to a value, or unset it entirely.` | the variable is set but empty. Remove it entirely to fall back to the default |
 
-The wording of the first, third, fourth and fifth comes from Node's own argument
-parser, so treat the examples above as indicative rather than exact — they can be
-reworded between Node releases. The `Empty …` messages are ours and stable.
+The rows starting `option` / `Option` / `Unknown option` come from Node's own
+argument parser, so treat that wording as indicative rather than exact — it can
+change between Node releases. The `Empty …` and `Expected one positional …`
+messages are ours and stable.
 
 Two things that are *not* errors: `CORS_ORIGIN=` is a valid way to say "no extra
 origins", and a variable that a CLI flag overrides is never read, so `--port
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -4,6 +4,35 @@ Common issues running the [Zendesk MCP Server](../README.md) and how to diagnose
 them. See also [Configuration](configuration.md) for the flags and environment
 variables referenced below.
 
+## The server refuses to start and names a flag or a variable
+
+The server would rather not boot than run with config it cannot honour, so a
+malformed invocation fails immediately. The message names the knob at fault, and
+never repeats the value you passed — a mistyped `--token=<secret>` reports only
+the flag name.
+
+| What you get | What to look for |
+| --- | --- |
+| `option '--mode <value>' argument missing` | the flag is the last thing on the command line; its value was forgotten |
+| `Empty value for --mode.` | the value is an empty string — usually `--mode "$VAR"` with `VAR` unset in the shell or compose file |
+| `Option '--host' argument is ambiguous` | the next token is another flag, so `--host` would have swallowed it. To pass a value that really does start with a dash, use `--host=-value` |
+| `Unknown option '--moed'` | a typo, or a flag from a different version. Check it against the [CLI reference](configuration.md#cli-reference) |
+| `Option '--read-only' does not take an argument` | a standalone flag was given a value |
+| `Empty PORT. Set it to a value, or unset it entirely.` | the variable is set but empty. Remove it entirely to fall back to the default |
+
+The wording of the first, third, fourth and fifth comes from Node's own argument
+parser, so treat the examples above as indicative rather than exact — they can be
+reworded between Node releases. The `Empty …` messages are ours and stable.
+
+Two things that are *not* errors: `CORS_ORIGIN=` is a valid way to say "no extra
+origins", and a variable that a CLI flag overrides is never read, so `--port
+8080` alongside a stray `PORT=` starts normally.
+
+If a deployment that used to work now fails here, that is the point — it was
+running with config that had been silently dropped. The most common case is an
+empty `HC_RESOURCE_SCHEME`, which used to fall back to `zendesk-hc` while
+runbooks expected a branded scheme.
+
 ## The browser doesn't open during OAuth login
 
 The OAuth flow opens your default browser on the first tool call. The first call
--- a/src/config.ts
+++ b/src/config.ts
@@ -266,6 +266,10 @@ const parseCliArgs = (args: string[]): CliResult => {
     );
   }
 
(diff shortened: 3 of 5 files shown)


dlecan:
CI is red on `build-and-test`, `guard` and `Changed lines`, and none of it is this branch: all three failed in **`Prepare all required actions`**, before running a single step.

```
Getting action download info
Failed to resolve action download info. Error: Service Unavailable
Retrying in 14 seconds
Failed to resolve action download info. Error: Service Unavailable
##[error]Service Unavailable
```

`Changed lines` got as far as checking out `pull/218/merge` and then failed resolving the next action, so Stryker never ran — the job's own log ends with `No files were found with the provided path: reports/mutation/`, which is the absent report, not a mutation failure. `guard` reported `Internal Server Error` on its first two attempts before degrading to `Service Unavailable`. A re-run of `build-and-test` failed identically, so this is not a one-off.

CodeQL passed on both analyses (it started before the outage deepened).

Nothing to fix here, so no commit: this needs GitHub Actions to recover, then a re-run. I'll retry and report back.

For the record, the same gates all pass locally on `c3fa8f3`:

| Gate | Result |
| --- | --- |
| `pnpm typecheck` | clean |
| `pnpm check` | clean |
| `pnpm test` | 719 passing, 33 files |
| `pnpm test:coverage` | thresholds met, `config.ts` at 100% branch coverage |
| `pnpm test:mutation:diff origin/main HEAD` | 116/116 killed, 100% |

---

dlecan:
Every substantive gate has passed on CI on `c3fa8f3`:

| Job | Result |
| --- | --- |
| `build-and-test` | all steps green at 17:14 — `pnpm check`, `pnpm typecheck`, `pnpm test:coverage`, `pnpm build`, `scripts/smoke-test.mjs`, `pnpm pack` |
| `Changed lines` (Stryker) | success — 116/116 mutants killed on the changed lines |
| `guard` | success |
| `CodeQL` + both `Analyze` | success |

`smoke-node20` is the sole exception, and it never ran: across repeated attempts it was assigned no runner (`runner_id: 0`, `runner_name: ""`) and was cut at the 15-minute mark each time. It needs a plain re-run once Actions capacity recovers — no code change.

One caveat on what the checks currently display: re-running the CI workflow to retry `smoke-node20` also cleared `build-and-test`'s recorded success, and that fresh attempt has since sat queued without materialising any jobs. So the green tick for `build-and-test` is missing from the UI even though the job did pass — the run linked above is the evidence.

---

src/config.ts
```diff
@@ -158,133 +162,137 @@ const parsePort = (raw: string, label: string): number => {
...
     }
 
-    // Only the first non-flag argument is taken; `subdomain` being unset is
-    // exactly the "no positional seen yet" state.
-    if (!arg.startsWith('-') && result.subdomain === undefined) {
-      result.subdomain = arg;
-    }
+    const field = FIELD_BY_FLAG.get(flag);
+    if (field) Object.assign(result, { [field]: value });
+  }
```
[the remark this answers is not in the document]
dlecan: Not applying this one: the suggested shape is rejected by this repo's own lint config. I did implement it to check rather than assume — `Object.entries(values).reduce<CliResult>((acc, …) => ({ ...acc, ...effect }), …)` — and `pnpm check` fails on it twice:

```
src/config.ts:283:28 lint/performance/noAccumulatingSpread
  ! Avoid the use of spread (`...`) syntax on accumulators.
src/config.ts:286:24 lint/performance/noAccumulatingSpread
  ! Avoid the use of spread (`...`) syntax on accumulators.
```

`noAccumulatingSpread` is enabled in `biome.json`, and `pnpm check` runs with `--error-on-warnings`, so a reducer built from object copies cannot land here. The two guidelines genuinely pull against each other on this function, and the lint gate is the one that blocks CI.

On the substance: `parseCliArgs` is already pure — same argv in, same `CliResult` out, no side effects, and the accumulator is a function-local object that never escapes. What is mutated is not shared or observable state.

For completeness, the other route out of the mutation — replacing both lookup maps with explicit per-flag assignments, which would also have restored the compile-time checking the computed key loses — was tried earlier in this PR and backed out too: 15 straight-line `if`s put `parseCliArgs` at a cognitive complexity of 22 against the ceiling of 15 in `noExcessiveCognitiveComplexity`.

I have added a comment at the accumulator recording this, so the next reader does not have to rediscover it (9666362).

---

docs/configuration.md
```diff
@@ -48,6 +48,30 @@ Options:
 
 `--namespace` and `--read-only` are applied before the proxies are registered, so they narrow the surface in every mode — in the default `namespace` mode, `--namespace help_center` registers a single proxy (`zendesk_help_center`) instead of the full set of namespace proxies.
 
+### A malformed invocation fails at startup
```
dlecan, line 51: Suggestion: use this skill https://github.com/blader/humanizer ti humanize this text

docs/troubleshooting.md
```diff
@@ -4,6 +4,37 @@ Common issues running the [Zendesk MCP Server](../README.md) and how to diagnose
 them. See also [Configuration](configuration.md) for the flags and environment
 variables referenced below.
 
+## The server refuses to start and names a flag or a variable
```
dlecan, line 7: Suggestion: use this skill https://github.com/blader/humanizer ti humanize this text

docs/configuration.md
```diff
@@ -48,6 +48,30 @@ Options:
...
+| `--mode` (nothing after it) | the value was forgotten |
+| `--mode ""` or `--mode=` | an empty value, typically `--mode "$VAR"` with `VAR` unset |
+| `--host --read-only` | the value is another flag, so `--host` would swallow it |
+| `--read-only=false` | a standalone flag takes no value |
+| `--moed all` | an unknown flag. A typo is not silently ignored |
+| `mycompany extra` | a second positional argument, so one of them would be dropped |
+
+Values are never echoed back in these messages, so a mistyped
+`--token=<secret>` reports only the flag name. Both `--flag value` and
+`--flag=value` are accepted.
```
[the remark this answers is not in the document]
dlecan: Both findings were right, and both are fixed in aa1d179.

**The syntax claim.** "Both `--flag value` and `--flag=value` are accepted" is contradicted by the `--read-only=false` row three lines above it, so it is now scoped to the flags that take a value, with standalone flags called out explicitly.

**The `--token` example.** Confirmed there is no such flag, and the README rules out static API-token auth. Rather than dropping the concrete case, I kept what it was there to show and made the premise explicit, since the no-echo guarantee matters most exactly when an operator misspells a flag that carries a secret:

```
Values are never echoed back in these messages, so an unsupported flag written
as `--anything=<secret>` is reported by name alone, with the value withheld. For
the flags that take a value, both `--flag value` and `--flag=value` work;
standalone flags such as `--read-only` accept no value in either form.
```

**The out-of-diff finding on the precedence bullet** was also correct, and it was the more interesting one. Verified against `loadConfig` before changing anything:

```
ZENDESK_SUBDOMAIN='' + argv ['mycompany']  -> boots, subdomain="mycompany"
ZENDESK_SUBDOMAIN='' + argv []             -> throws "Empty ZENDESK_SUBDOMAIN..."
```

The bullet said "a variable that a CLI *flag* overrides", and the positional `<subdomain>` is not a flag, yet it shadows the variable through the same `??` chain. The wording now covers the command line rather than just flags, and names this case. The behaviour had no test either: the existing one only exercised `--port` and `--hc-resource-scheme`, so a test for the positional case comes with the fix (725 passing).

---

dlecan:
CI is green on `01d1b2b`, the current head, with the branch up to date with `main`.

| Job | Result |
| --- | --- |
| `build-and-test` | success — `pnpm check`, `pnpm typecheck`, `pnpm test:coverage`, `pnpm build`, `scripts/smoke-test.mjs`, `pnpm pack` |
| `smoke-node20` | success — the packed tarball on Node 20 |
| `Changed lines` (Stryker) | success — mutation gate on the changed lines |
| `CodeQL` + `Analyze` (actions, javascript-typescript) | success |
| `Full scope` | skipped, as designed |

This supersedes my earlier comment on `c3fa8f3`, which was written while the platform was degraded and reported `smoke-node20` as never having run. It has since run and passed, first on `aa1d179` and again here on a clean first attempt.

`01d1b2b` merges `main` at `854fef9` (pnpm 11.19.0 and the two lock-file maintenance PRs), so the branch is no longer behind.

The one thing still open is not technical: the scope question in the PR description. Rejecting empty environment variables and rejecting unknown flags both go beyond what #174 asks for, and that issue carries no discussion to point at, so those calls are mine alone and are yours to confirm or send back.

---

docs/configuration.md
dlecan: Done in 1ef5b89, then refined in aa1d179.

I read the skill and applied it to this section. The rule that actually bit was the one against em dashes, and I could check it against your own precedent rather than guess: line 49 of this file, which #217 touched, replaces an em dash with a period. So the three in this section became a colon, a comma or a sentence break, whichever the clause called for:

- `unknown flag — a typo is not silently ignored` → `an unknown flag. A typo is not silently ignored`
- `has to be repeated — --namespace tickets …, never …` → a colon, then a separate sentence for the wrong form
- `naming the variable — unset it to get the default` → a sentence break

Also checked, and clean here: no curly quotes, no typographic ellipsis, none of the skill's flagged vocabulary.

aa1d179 then fixed two accuracy problems CodeRabbit found in the same paragraph, which are worth more than the wording: the `--token` example named a flag this server does not have, and the claim that both `--flag value` and `--flag=value` are accepted is contradicted by the `--read-only=false` row three lines above.

---

docs/troubleshooting.md
dlecan: Done in 1ef5b89.

Same treatment as the `docs/configuration.md` section. Three em dashes in this one, each replaced by what the clause actually needed:

- `never repeats the value you passed — a mistyped …` → a colon, since what follows is the illustration
- `an empty string — usually --mode "$VAR" …` → a comma
- `that is the point — it was running with config …` → a sentence break

Two more in the table rows became a comma and a relative clause. I also dropped the typographic ellipsis from `Empty …` in favour of `Empty ...`.

One thing I changed beyond the wording, because adding a row exposed it: the paragraph under the table used to identify the Node-generated messages by position ("the first, third, fourth and fifth"). That ordering would have silently gone wrong the moment anyone inserted a row, and its closing sentence did not cover the new `Expected one positional …` message. It now identifies them by how they read (`option` / `Option` / `Unknown option`), which no longer depends on row order.

---

PR #225, opened 2026-08-11
feat(help-center): section & category translation tools (#224)

### LINKED ISSUE #224: No section or category translation tools: articles get published in a target locale under sections that only exist in the source locale
### Problem

Translations are only reachable at the article level. `list_article_translations` takes an `article_id` and nothing else, and `create_article_translation` / `update_article_translation` are article-scoped too. Sections and categories have no translation tool at all.

Reading is partly possible as a side effect: `list_sections` and `list_categories` accept a `locale`, and the filtered listing only returns the ones that have a translation in that locale. That is enough to spot a gap, not enough to explain it. A section absent from the target-locale listing may have no translation at all, or one that exists but is still a draft, and those need different fixes. Writing is not possible at all.

### Why it matters

Publishing an article in a second locale does not make it reachable in that locale. Readers browse the tree, so when the parent section has no translation, a fully translated article sits under a section title that only exists in the source locale.

We hit this on a bilingual Help Center: several untranslated sections turned out to hold articles already published in the target locale. The parent categories were translated, so the section was the only missing link in each case. Detecting it took a handful of read calls; fixing it took a human in Guide admin, one section at a time, because the server offers no way to do it.

It also undercuts a rule worth enforcing on the caller side, that an article is created or updated in both locales in the same pass. The rule holds for article bodies and silently fails for the tree they live in.

### Proposed solution

Section and category translation tools mirroring the article ones:

- list the translations of a section or a category, with their `draft` state
- create a translation
- update a translation, including publishing it by flipping `draft`

The endpoints belong to the same API family as the article ones ([Translations](https://developer.zendesk.com/api-reference/help_center/help-center-api/translations/)): `GET` and `POST` on `/api/v2/help_center/sections/{section_id}/translations`, `PUT` on `/api/v2/help_center/sections/{section_id}/translations/{locale}`, and the same shape under `categories/{category_id}`. The translation object carries a writable `draft` attribute.

Worth considering on top of that: report, per active locale, which sections and categories have no published translation. `zendesk-hc://topology` already walks the whole tree and knows the active locales, so it is the natural place to expose it.

### Alternatives considered

- Deducing gaps from `list_sections?locale=…`. Works as a detector, but it cannot tell a missing translation from an unpublished one, and it fixes nothing.
- Fixing it in Guide admin by hand. Acceptable once on a short list, not for a recurring consistency check across two locales.

### Scope

- [x] Read-only (listing section and category translations, reporting gaps)
- [x] Write operation (creating and publishing section and category translations)

### Additional context

Observed with `@fruggr/zendesk-mcp-server@latest` in `single` mode, admin account, two active locales.

dlecan:
## Summary

Translations were only reachable at the article level, so publishing an article in a second locale left it unreachable: readers browse the tree, and a translated article sat under a section title that only existed in the source locale. This adds five `help_center` tools that read, audit and write section and category translations, so "translate the article in both locales in the same pass" also holds for the tree it lives in.

## Linked issue

Closes #224

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tooling / CI

## What is added

| Tool | Mode | Shape |
|------|------|-------|
| `list_section_translations` | read | `{ section_id }` → per locale: localized name, description set/empty, draft state |
| `list_category_translations` | read | `{ category_id }` → same |
| `find_translation_gaps` | read | `{ locale, category_id? }` → every node with *no translation* vs *draft translation* |
| `set_section_translation` | write | `{ section_id, locale, name?, description?, draft? }` → create-or-update |
| `set_category_translation` | write | `{ category_id, locale, name?, description?, draft? }` → create-or-update |

The `set_*` tools probe the node's translations, then `POST` when the locale is absent and `PUT` when it exists, sending **only** the fields passed. So publishing a draft is a single `draft: false`, and omitting `description` never blanks an existing one.

`find_translation_gaps` fetches `/locales` plus the two listings, then one `translations?locales=…` request per node, in waves of 5, bounded by `ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES` (default 60). Categories are scanned first; the report names what it left unscanned. An inactive locale is flagged as a warning rather than reported as a tree-wide catastrophe.

The tools speak `name` / `description` — the vocabulary `list_sections` and `list_categories` already return — and map onto the API's `title` / `body`, so a caller never has to know those two fields change meaning on a section or a category.

**Why the audit probes each node instead of diffing two listings.** A locale-filtered listing answers neither question: a node with no translation in the locale is absent from it *without saying why*, and a node whose translation is an unpublished **draft** is still returned, under the draft's own name (measured with an admin token during validation — see below). So absence there is ambiguous and presence does not mean published. The listing-diff design would have been far cheaper and was considered; it would also have missed every draft-translated node, i.e. exactly the case #224 was opened about. The per-node fan-out is the price of the honest answer, which is what the cap and the wave size exist to bound.

**Two deliberate departures from the shape proposed in the issue**, both following the usage-first rule in `AGENTS.md` (also flagged on the issue before starting):

1. **One upsert per level instead of a `create_*` / `update_*` pair.** A split pair forces a listing call before every write, or eats the `400 "translation already exists"` the API returns on a duplicate `POST`. Two write tools instead of four, and no guaranteed error path. The cost is a naming divergence from `create_article_translation` / `update_article_translation`.
2. **The gap report is its own opt-in tool, not a `zendesk-hc://topology` section** (which the issue suggested). Topology is ambient context, auto-read by clients and cached per session, so a per-locale scan there would make every read pay for it. `docs/help-center-context.md` is therefore unchanged.

Kept from the issue: separate tools per level, mirroring `list_sections` / `list_categories`, rather than one tool with a `type` discriminator that would add a wrong-type-404 failure mode.

**Sections and categories are treated as iso** — same endpoint family, same translation object, one shared `upsertNodeTranslation`, parameterised by kind rather than written twice. The API reference backs this: identical endpoint shapes for list/show/create/update, the same Translation object with `title` mandatory on both, and `name`/`description` documented identically on the Section and Category objects. The rule is stated on `TreeNodeKind`, and the unit tests enforce it by running every read and write case over both levels from one table (12/12 parity), so a new case cannot land on one level only.

**API surface confirmed** against the [Translations reference](https://developer.zendesk.com/api-reference/help_center/help-center-api/translations/): the endpoints, `title`/`body`/`draft`/`locale` writable, `title` mandatory on create, the `locales` filter on list endpoints. There is **no** missing-translations endpoint for sections or categories (only `articles/{id}/translations/missing`), which is why the audit has to fan out. That `title` is the localized *name* and `body` the localized *description* is **not** documented for either level; validation scenario **S1** confirmed it on a live section.

**Commit trail** — everything after the first commit is a review or validation pass, not new feature work:

- `2276cb8`: the scoped audit reads its category directly (filtering a one-page listing silently reported "0 categories scanned" for a category further down); a no-op update is refused instead of reported as a write; an empty level no longer claims to be fully translated.
- `aee3d5e`: the scan probes in waves of 5 instead of firing up to 60 requests at once (Zendesk caps concurrent requests per account, there is no retry here, and one 429 discarded the whole audit — follows the `VIEW_COUNT_BATCH` pattern in `tickets.ts`); a level the cap never reached is no longer rendered like a genuinely empty one; the `locales` filter value is normalized so it cannot disagree with the case-insensitive local comparison.
- `2140595`: corrects the overclaim S3/S7 caught (below). Wording only, no logic.
- `1267a5f`: both-levels test table, so nothing is asserted on sections alone; category write over the namespace proxy in the integration suite.
- `2194ad5`: records the untaken sideload route (#226) next to the cost it would remove.

**Mutation gate** reports `No changed lines inside the mutate scope` — `stryker.config.mjs` scopes mutation to `src/auth`, `src/client`, `src/routing`, `src/utils` (minus `formatting.ts`) and `src/config.ts`, none of which this PR touches.

## Author checklist
- [x] If this PR resolves an issue, its description above links it with a closing keyword (`Closes #224`) so it auto-closes on merge to `main`
- [x] `pnpm test` passes locally (705 tests)
- [x] `pnpm typecheck` passes
- [x] I ran a review pass on the diff and addressed its findings — see `2276cb8` and `aee3d5e`
- [x] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md`
- [x] If the change has **MCP-runtime-observable** behaviour: this PR carries a `## Functional validation plan` below, executed by an independent validator, reports posted as PR comments

### Validation outcome — green

Independently validated end-to-end against a live tenant, twice.

| Run | Commit | Result |
|-----|--------|--------|
| 1 | `aee3d5e` | [report](https://github.com/fruggr/zendesk-mcp-server/pull/225#issuecomment-5255363078) — 11/13 OK, **S3 + S7 FAILED** on one shared point, S12 write half skipped |
| 2 | `2140595` | [report](https://github.com/fruggr/zendesk-mcp-server/pull/225#issuecomment-5255534401) — **12/13 OK**, S12 write half skipped |

**What run 1 caught.** `list_sections({ locale })` **does** return sections whose translation in that locale is an unpublished draft (admin token), rendered under the draft name. Three tool descriptions, two code comments and the `docs/configuration.md` cost note claimed such a node was omitted — which would lead an agent to read "listed ⇒ published", wrong on exactly the path these tools address. The other half of the claim, that a node with *no* translation is omitted, was verified exhaustively across all 45 sections and 13 categories. Fixed in `2140595`, wording only. **The implementation never depended on the false half**: the audit reads the draft flag per node rather than diffing the two listings, a choice made in the plan precisely so as not to rest on undocumented listing behaviour — a listing-diff audit would have missed every draft-translated node.

S3 and S7 in the table below are **amended**: their original wording expected a draft-translated node to be *absent* from the locale listing, which is what run 1 disproved. A re-run should expect such a node to remain listed.

**Later commits are not covered by run 2, and do not need to be.** `git diff 2140595 2194ad5 -- src/` with comment lines filtered out is empty: everything since is tests, an integration scenario and comments. Nothing run 2 measured can have moved.

**Still unvalidated, accepted:** S12's write half. There is no expendable category on this production tenant, so the category *write* path has unit and integration coverage only. `1267a5f` brings it to full parity with the section path, but those assertions run against MSW — they prove the request built and the text rendered, not that Zendesk answers `POST /categories/{id}/translations` like its section twin. The API reference says they are one endpoint family with an identical object; that is documentation, not a measurement. First thing to suspect if a category write ever misbehaves.

**Follow-up filed:** #226 — both list endpoints document a `translations` sideload that could collapse the per-node fan-out into the two listings already fetched. Not taken here: the docs never say whether a sideloaded translation carries `draft`, which is the whole point, and settling it needs one live call.

> Placeholders below are written `{like_this}` — substitute before calling.

### Context

Five new tools on the `help_center` namespace, exercised by calling `mcp__zendesk-local__list_section_translations`, `list_category_translations`, `find_translation_gaps`, `set_section_translation` and `set_category_translation` directly. Nothing about transports, auth or the resource surface changed, so the HTTP transport and the OAuth paths need no coverage here; `zendesk-hc://topology` is deliberately untouched and its output must be **identical** to `main`.

Unit tests already cover the rendering, the POST-vs-PUT decision, the payload built on each path, the cap arithmetic, the wave bound and the inactive-locale warning against MSW mocks. What they cannot cover — and what this plan is for — is whether the real Zendesk API agrees with our reading of it on two points: that `title`/`body` on a section/category translation really are the name and the description, and that `draft` behaves as the publish switch.

### Setup & prerequisites

- No build, checkout or credential work: the environment is on this branch with the server running and authenticated. Record `git rev-parse HEAD` and put it in the report.
- **Pick a throwaway section** for the write scenarios — one whose localized name you may change. Get candidates from `zendesk-hc://topology` or `list_sections`. Note its id, and capture its current translations (`list_section_translations`) **before touching anything**, so the original state can be restored and pasted into the report.
- **Pick the target locale**, `{locale}`: a non-default active locale from `zendesk-hc://topology`.
- Everything is observable in the tool output; no log level to raise.
- A `403` on the write scenarios is a `BLOCKED` finding — report it, do not go looking for another token.

### Scenarios

| id | Validates | Manipulation / call | Expected observable |
| **S1** | **`title` is the name, `body` the description** — the assumption the whole mapping rests on | `list_sections` (default locale); note a section's name and description. Then `list_section_translations` with that `section_id` | The default-locale entry's `**Name**` equals the name from `list_sections`, and `**Description**` reads `set` when the section has a description, `empty` when it has none. **If `Name` shows anything other than the section name, stop and report — the rest of the plan is invalid.** |
| **S2** | Reads report draft state per locale | `list_section_translations` and `list_category_translations` on a section and a category translated in `{locale}` | One entry per locale, each with `Name`, `Description`, `Draft`, `Updated`. `Draft` is `false` for a published translation. No article-style `**Title**` label anywhere. |
| **S3** | The audit finds real gaps and classifies them | `find_translation_gaps({ locale: "{locale}" })` | A `# Translation gaps` report with `## Categories (N scanned)` / `## Sections (N scanned)`. Cross-check against `list_sections({ locale: "{locale}" })`: every node reported as *no translation* is **absent** from it; every node reported as *draft translation* **may still be present** (under its draft name); every node **not** reported is present. A node reported *no translation* yet present in the listing, or an unreported node missing from it, is a FAIL — paste both listings side by side. |
| **S4** | The audit's classification matches the per-node read | For one node reported *draft translation* (if any) and one reported *no translation*, call `list_section_translations` on it | *draft* ⇒ an entry for `{locale}` with `Draft: true`. *no translation* ⇒ no entry for `{locale}` at all. |
| **S5** | Create path: a locale with no translation | On the throwaway section: `set_section_translation({ section_id: {id}, locale: "{locale}", name: "{a recognizable test name}", description: "validation probe" })` | `Translation created for section #{id} in "{locale}" (published).` Then `list_section_translations` shows a new `{locale}` entry with `Draft: false`, `Description: set`. Then the section appears in `list_sections({ locale: "{locale}" })` under the new name — **this is the user-facing point of the whole PR**. |
| **S6** | Update path leaves omitted fields alone | `set_section_translation({ section_id: {id}, locale: "{locale}", name: "{a different test name}" })` — no `description`, no `draft` | `Translation updated …`, the name changes, and `Description` is **still** `set` (the description from S5 survived). |
| **S7** | `draft` is the publish switch, both ways | `set_section_translation({ section_id: {id}, locale: "{locale}", draft: true })`, then the same with `draft: false` | First call: `(draft, not visible to end users)`, `Draft: true` on the per-node read, and `find_translation_gaps` reports the node as *draft translation (not published)*. The node **stays** in `list_sections({ locale: "{locale}" })`, under the draft's own name — that is the documented behaviour, not a failure. Second call: `(published)`, and the node drops out of the audit. |
| **S8** | No-op writes are refused, not faked | `set_section_translation({ section_id: {id}, locale: "{locale}" })` — nothing else | An error saying `Nothing to write`, naming the locale. **No** "Translation updated" message. |
| **S9** | Creation requires a name | `set_section_translation({ section_id: {id}, locale: "{an active locale this section has no translation in}" })` | An error saying the section has no translation in that locale yet and that `name` is required. Nothing is created — confirm with `list_section_translations`. |
| **S10** | Scoping reads the category itself | `find_translation_gaps({ locale: "{locale}", category_id: {a real category id} })`, then again with a **nonexistent** category id | First: exactly `## Categories (1 scanned)`, that category only, and its own sections under `## Sections`. Second: a Zendesk `404`-shaped error, **not** an empty report. |
| **S11** | Inactive locale is flagged, not reported as catastrophe | `find_translation_gaps({ locale: "zz" })` | A warning that `"zz"` is not an active locale, naming the real active locales, above the (all-missing) listing. |
| **S12** | Category write path | `set_category_translation` on a category — same create-then-publish shape as S5/S7. If no category is expendable, run only the read half and mark the write half `SKIPPED` with the reason | Same messages with `category #{id}`, and the category becomes visible in `list_categories({ locale: "{locale}" })`. |
| **S13** | Nothing else regressed | `zendesk-hc://topology`, `list_sections`, `list_categories`, `list_article_translations`, `compare_translations` | Output shape unchanged from `main` — in particular topology carries **no** new translation-coverage section. |

**Cleanup.** After S5–S7 and S12, restore the throwaway node with `set_section_translation` back to the name captured in Setup. If the `{locale}` translation did not exist beforehand, say so in the report: there is no delete-translation tool here, so a leftover test translation must be flagged explicitly for manual removal in Guide.

### Long-running behaviours

None. No timers, no TTL, no polling. The two bounded-cost paths (the 60-node cap and the wave size) are covered by unit tests — the wave test observes peak in-flight requests — and neither needs a 60+-node Help Center to validate.

### Priorities

Do **S1 first and block on it** — it is the ground-truth check for the field mapping every other scenario assumes. Then **S3, S5, S7**: the audit-then-publish loop is the actual user story from the issue. **S6, S8, S9, S10** are the guard rails. **S2, S4, S11, S12, S13** last.

### Reporting instructions

Post a PR comment (English) with: the commit SHA tested, one line per scenario id with `OK` / `FAIL` / `BLOCKED` / `SKIPPED` plus the observed evidence (paste the relevant tool output, trimmed), an explicit note on whether the throwaway node was restored, and a one-line overall verdict. For any FAIL, include both what you expected and what you got.

## Summary by CodeRabbit

- **New Features**
  - Added tools to list, create, update, and publish Help Center translations for sections and categories.
  - Added translation-gap audits identifying missing and draft translations.
  - Added locale filtering, scoped audits, and scan-limit reporting.
- **Documentation**
  - Documented translation tools and configurable scan limits.
- **Bug Fixes**
  - Improved locale handling, validation, and preservation of existing translation values.

### DIFF
--- a/docs/mcp-tools-reference.md
+++ b/docs/mcp-tools-reference.md
@@ -48,6 +48,9 @@ out every `write` tool before the proxies are built.
 | `list_articles` | List articles with sorting and translation info | read |
 | `list_promoted_articles` | List the promoted ("featured") articles (scans + filters client-side, page-capped) | read |
 | `list_article_translations` | List available translations for an article | read |
+| `list_section_translations` | List a section's translations with their draft state, telling an untranslated section apart from an unpublished one | read |
+| `list_category_translations` | List a category's translations with their draft state, telling an untranslated category apart from an unpublished one | read |
+| `find_translation_gaps` | Audit the category → section tree for a locale and report every node with no translation or an unpublished draft (one request per node scanned, capped) | read |
 | `list_article_attachments` | List attachments on an article | read |
 | `list_permission_groups` | List Guide permission groups (needed to create articles; requires Guide-admin / Help Center manager rights) | read |
 | `list_content_tags` | List Guide content tags (end-user visible), cursor-paginated with name-prefix filter and sort | read |
@@ -60,6 +63,8 @@ out every `write` tool before the proxies are built.
 | `archive_article` | Archive (soft-delete) an article; recoverable only via the Guide admin UI | write |
 | `create_article_translation` | Create a translation for an article | write |
 | `update_article_translation` | Update an article's translation (full body) | write |
+| `set_section_translation` | Create or publish a section's translation in one locale (creates when absent, updates otherwise; only the fields passed are written) | write |
+| `set_category_translation` | Create or publish a category's translation in one locale (creates when absent, updates otherwise; only the fields passed are written) | write |
 | `update_article_section` | Replace a single section of an article | write |
 | `create_content_tag` | Create a new Guide content tag | write |
 | `create_article_attachment` | Upload a file to a Help Center article (returns the created attachment) | write |
--- a/docs/mcp-tools-reference.md
+++ b/docs/mcp-tools-reference.md
@@ -50,7 +50,7 @@ out every `write` tool before the proxies are built.
 | `list_article_translations` | List available translations for an article | read |
 | `list_section_translations` | List a section's translations with their draft state, telling an untranslated section apart from an unpublished one | read |
 | `list_category_translations` | List a category's translations with their draft state, telling an untranslated category apart from an unpublished one | read |
-| `find_translation_gaps` | Audit the category → section tree for a locale and report every node with no translation or an unpublished draft (one request per node scanned, capped) | read |
+| `find_translation_gaps` | Audit a locale across every category **and** every section, reporting each one that has no translation or only an unpublished draft; scope it to one category with `category_id` (one request per node scanned, capped) | read |
 | `list_article_attachments` | List attachments on an article | read |
 | `list_permission_groups` | List Guide permission groups (needed to create articles; requires Guide-admin / Help Center manager rights) | read |
 | `list_content_tags` | List Guide content tags (end-user visible), cursor-paginated with name-prefix filter and sort | read |
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -47,8 +47,9 @@ export const ARTICLE_RESOURCES_SCAN_MAX_PAGES = positiveIntEnv(
 
 // Max category + section nodes probed by find_translation_gaps. Unlike articles,
 // sections and categories have no "missing translations" endpoint, and the
-// locale-filtered listing cannot tell an absent translation from a draft one, so
-// the only honest answer costs one translations request per node. This bounds
+// locale-filtered listing cannot answer the question either — it omits a node with
+// no translation without saying why, and still returns one whose translation is a
+// draft — so the only honest answer costs one translations request per node. This bounds
 // that fan-out on a large Help Center; nodes beyond the cap are left unscanned
 // and the tool says so. Override via ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES.
 export const TRANSLATION_GAP_SCAN_MAX_NODES = positiveIntEnv(
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -45,6 +45,17 @@ export const ARTICLE_RESOURCES_SCAN_MAX_PAGES = positiveIntEnv(
   20,
 );
 
+// Max category + section nodes probed by find_translation_gaps. Unlike articles,
+// sections and categories have no "missing translations" endpoint, and the
+// locale-filtered listing cannot tell an absent translation from a draft one, so
+// the only honest answer costs one translations request per node. This bounds
+// that fan-out on a large Help Center; nodes beyond the cap are left unscanned
+// and the tool says so. Override via ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES.
+export const TRANSLATION_GAP_SCAN_MAX_NODES = positiveIntEnv(
+  'ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES',
+  60,
+);
+
 // Local port the OAuth PKCE flow listens on for the browser callback. Must match
 // the redirect URL registered in the Zendesk OAuth client. Deliberately picked
 // outside the usual dev range (3000/5000/8080…) and below the OS ephemeral
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -45,22 +45,12 @@ export const ARTICLE_RESOURCES_SCAN_MAX_PAGES = positiveIntEnv(
   20,
 );
 
-// Max category + section nodes probed by find_translation_gaps. Unlike articles,
-// sections and categories have no "missing translations" endpoint, and the
-// locale-filtered listing cannot answer the question either — it omits a node with
-// no translation without saying why, and still returns one whose translation is a
-// draft — so the only honest answer costs one translations request per node. This bounds
-// that fan-out on a large Help Center; nodes beyond the cap are left unscanned
-// and the tool says so. Override via ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES.
-//
-// A cheaper route may exist and is deliberately not taken yet: the sections and
-// categories LIST endpoints document a `translations` sideload
-// (`?include=translations`), which would collapse the whole fan-out — and this cap
-// with it — into the two listings already fetched. What the docs do not say is
-// whether a sideloaded translation carries `draft`, and `draft` is the entire
-// point here, so adopting it has to be measured against a live tenant first.
-// Tracked separately; until then this stays on the endpoint that is known to
-// return the flag.
+// Max category + section nodes probed by find_translation_gaps. Sections and
+// categories have no "missing translations" endpoint and the locale-filtered
+// listing cannot substitute for one, so the answer costs one request per node;
+// this bounds the fan-out and the tool reports what it left unscanned. A
+// `translations` sideload could remove it — untaken, see #226. Override via
+// ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES.
 export const TRANSLATION_GAP_SCAN_MAX_NODES = positiveIntEnv(
   'ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES',
   60,
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -52,6 +52,15 @@ export const ARTICLE_RESOURCES_SCAN_MAX_PAGES = positiveIntEnv(
 // draft — so the only honest answer costs one translations request per node. This bounds
 // that fan-out on a large Help Center; nodes beyond the cap are left unscanned
 // and the tool says so. Override via ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES.
+//
+// A cheaper route may exist and is deliberately not taken yet: the sections and
+// categories LIST endpoints document a `translations` sideload
+// (`?include=translations`), which would collapse the whole fan-out — and this cap
+// with it — into the two listings already fetched. What the docs do not say is
+// whether a sideloaded translation carries `draft`, and `draft` is the entire
+// point here, so adopting it has to be measured against a live tenant first.
+// Tracked separately; until then this stays on the endpoint that is known to
+// return the flag.
 export const TRANSLATION_GAP_SCAN_MAX_NODES = positiveIntEnv(
   'ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES',
   60,
--- a/src/tools/help-center.ts
+++ b/src/tools/help-center.ts
@@ -105,6 +105,13 @@ const listTranslations = (
 // localized *name* and `body` the localized *description*. The tools speak
 // name/description — the vocabulary list_sections and list_categories already
 // return — and do the mapping here, so a caller never has to know about it.
+//
+// The two levels are treated as ISO: same endpoints, same object, same
+// behaviour, parameterised by kind rather than implemented twice. Anything done
+// to one level must be done to the other — schema, description, error message,
+// report line, test. The unit tests enforce that by running every case over both
+// levels from one table (see NODE_LEVELS in the tests), which is also what
+// stands in for the category write path's missing live validation (#225, S12).
 type TreeNodeKind = 'sections' | 'categories';
 
 const NODE_LABEL: Record<TreeNodeKind, string> = {
--- a/src/tools/help-center.ts
+++ b/src/tools/help-center.ts
@@ -221,9 +221,13 @@ const nodeTranslationWriteText = (
 // --- find_translation_gaps.
 //
 // A node is a gap when the target locale has no translation at all, or has one
-// that is still a draft. Those need different fixes (write a translation vs flip
-// `draft`), which is exactly what `list_sections?locale=…` cannot tell you: it
-// omits both cases alike.
+// that is still a draft. Those need different fixes — write a translation vs flip
+// `draft` — and `list_sections?locale=…` answers neither question. Measured on a
+// live tenant with an admin token (#225 validation): a node with no translation is
+// absent from that listing, but a node whose translation is a DRAFT is still
+// returned, rendered with the draft's own name. So absence there is ambiguous and
+// presence does not mean published, which is why this audit probes the
+// translations endpoint per node instead of diffing the two listings.
 type GapReason = 'missing' | 'draft';
 
 interface TranslationGap {
@@ -1132,7 +1136,7 @@ export const createHelpCenterTools = (ctx: ToolContext): ToolDefinition[] => {
       readOnly: true,
       title: 'List Section Translations',
       description:
-        'List the translations of a Help Center section: for each locale, the localized name, whether a description is set, and whether the translation is published or still a draft. Reach for this when a section looks untranslated in a locale — list_sections with a locale omits a section that has no translation AND one whose translation is an unpublished draft, and only this tool tells the two apart. Fix either case with set_section_translation; to sweep every category and section at once, use find_translation_gaps.',
+        'List the translations of a Help Center section: for each locale, the localized name, whether a description is set, and whether the translation is published or still a draft. Reach for this when a section looks wrong in a locale, because list_sections with that locale cannot settle it: a section with no translation is omitted from it, while a section whose translation is an unpublished draft may still be listed there under the draft name — so appearing in that listing does not mean published, and the draft flag here is what decides. Fix either case with set_section_translation; to sweep every category and section at once, use find_translation_gaps.',
       inputSchema: z.object({
         section_id: z
           .number()
@@ -1162,7 +1166,7 @@ export const createHelpCenterTools = (ctx: ToolContext): ToolDefinition[] => {
       readOnly: true,
       title: 'List Category Translations',
       description:
-        'List the translations of a Help Center category: for each locale, the localized name, whether a description is set, and whether the translation is published or still a draft. Reach for this when a category looks untranslated in a locale — list_categories with a locale omits a category that has no translation AND one whose translation is an unpublished draft, and only this tool tells the two apart. Fix either case with set_category_translation; to sweep every category and section at once, use find_translation_gaps.',
+        'List the translations of a Help Center category: for each locale, the localized name, whether a description is set, and whether the translation is published or still a draft. Reach for this when a category looks wrong in a locale, because list_categories with that locale cannot settle it: a category with no translation is omitted from it, while a category whose translation is an unpublished draft may still be listed there under the draft name — so appearing in that listing does not mean published, and the draft flag here is what decides. Fix either case with set_category_translation; to sweep every category and section at once, use find_translation_gaps.',
       inputSchema: z.object({
         category_id: z
           .number()
@@ -1197,7 +1201,7 @@ export const createHelpCenterTools = (ctx: ToolContext): ToolDefinition[] => {
       readOnly: true,
       title: 'Find Help Center Translation Gaps',
       description:
-        'Audit the Help Center tree for a target locale and report every category and section that has no translation, or one that is still an unpublished draft. Use it before or after translating articles: an article published in a second locale is unreachable while its parent section only exists in the source locale, and that gap is invisible to list_sections. Costs one extra request per node scanned, capped (the report says so when the cap bites) — pass category_id to narrow it. Fix what it reports with set_section_translation / set_category_translation.',
+        'Audit the Help Center tree for a target locale and report every category and section that has no translation, or one that is still an unpublished draft. Use it before or after translating articles: an article published in a second locale is unreachable while its parent section only exists in the source locale. Listing sections in that locale cannot answer this — a node with no translation is simply absent, without saying why, and a node whose translation is an unpublished draft may still be listed under its draft name — so this audit reads the draft flag on each node instead of trusting that listing. Costs one extra request per node scanned, capped (the report says so when the cap bites) — pass category_id to narrow it. Fix what it reports with set_section_translation / set_category_translation.',
       inputSchema: z.object({
         locale: z
           .string()
--- a/src/tools/help-center.ts
+++ b/src/tools/help-center.ts
@@ -99,19 +99,12 @@ const listTranslations = (
 
 // --- Section / category translations.
 //
-// Sections and categories live on the same Translations endpoint family as
-// articles (`/{sections|categories}/{id}/translations`) and return the same
-// translation object, with two fields carrying different meanings: `title` is the
+// Same endpoint family as articles, same translation object, but `title` is the
 // localized *name* and `body` the localized *description*. The tools speak
-// name/description — the vocabulary list_sections and list_categories already
-// return — and do the mapping here, so a caller never has to know about it.
+// name/description (what list_sections/list_categories return) and map here.
 //
-// The two levels are treated as ISO: same endpoints, same object, same
-// behaviour, parameterised by kind rather than implemented twice. Anything done
-// to one level must be done to the other — schema, description, error message,
-// report line, test. The unit tests enforce that by running every case over both
-// levels from one table (see NODE_LEVELS in the tests), which is also what
-// stands in for the category write path's missing live validation (#225, S12).
+// The two levels are ISO: anything done to one must be done to the other. Tests
+// enforce it over both from one table (NODE_LEVELS).
 type TreeNodeKind = 'sections' | 'categories';
 
 const NODE_LABEL: Record<TreeNodeKind, string> = {
@@ -119,14 +112,10 @@ const NODE_LABEL: Record<TreeNodeKind, string> = {
   categories: 'category',
 };
 
-// `locales` narrows the response when the caller only cares about one language.
-// The result is filtered locally anyway (see findTranslation), so the tools stay
-// correct even where the server-side filter is not applied. The value is
-// lower-cased so the server-side filter cannot disagree with that local,
-// case-insensitive comparison: Zendesk stores locales lower-cased, and whether it
-// matches the filter case-sensitively is not documented — if it does, a caller's
-// "FR" would come back empty and every node would read as "no translation".
-// Normalizing costs nothing if the filter is case-insensitive after all.
+// `locales` only narrows the response; the result is filtered locally anyway, so
+// an unapplied server-side filter stays correct. Lower-cased because whether that
+// filter is case-sensitive is undocumented: if it is, a caller's "FR" comes back
+// empty and every node reads as untranslated.
 const listNodeTranslations = (
   subdomain: string,
   token: string,
@@ -141,8 +130,7 @@ const listNodeTranslations = (
     locale ? { locales: locale.toLowerCase() } : undefined,
   ).then((res) => res.translations ?? []);
 
-// Locales are matched case-insensitively: Zendesk normalizes its own to lower
-// case, but a caller (or a copy-paste from Guide) may well pass "fr-FR".
+// Case-insensitive: a caller may well pass "FR" or a copy-paste from Guide.
 const findTranslation = (
   translations: ZendeskTranslation[],
   locale: string,
@@ -152,14 +140,10 @@ const findTranslation = (
 };
 
 /**
- * Create-or-update a section/category translation in one call.
- *
- * The probe that decides POST vs PUT is the very call `list_*_translations`
- * makes, so the caller pays one request instead of being forced to list first —
- * or to recover from the 400 the API returns when a POST targets a locale that
- * already has a translation. Only the fields actually passed are sent on update,
- * so omitting `description` never blanks an existing one and omitting `draft`
- * never republishes (or unpublishes) by accident.
+ * Create-or-update a section/category translation in one call. The POST-vs-PUT
+ * probe spares the caller a listing round-trip, or the 400 a duplicate POST
+ * returns. Only the fields passed are sent on update, so omitting `description`
+ * never blanks it and omitting `draft` never (un)publishes by accident.
  */
 const upsertNodeTranslation = async (
   subdomain: string,
@@ -193,15 +177,13 @@ const upsertNodeTranslation = async (
   if (name !== undefined) updates['title'] = name;
   if (description !== undefined) updates['body'] = description;
   if (draft !== undefined) updates['draft'] = draft;
-  // An empty payload would round-trip to Zendesk and come back reported as an
-  // update, which is a lie. Say what is missing instead.
+  // An empty payload would round-trip and be reported as an update: a lie.
   if (Object.keys(updates).length === 0) {
     throw new Error(
       `Nothing to write: ${NODE_LABEL[kind]} #${nodeId} already has a "${existing.locale}" translation, so pass at least one of "name", "description" or "draft" to change it (draft: false publishes it).`,
     );
   }
-  // Zendesk's own spelling of the locale, not the caller's: the PUT path has to
-  // match the existing translation even when the input was cased differently.
+  // Zendesk's spelling, not the caller's, so the PUT path matches.
   const { translation } = await helpCenterPut<{ translation: ZendeskTranslation }>(
     subdomain,
     token,
@@ -227,14 +209,11 @@ const nodeTranslationWriteText = (
 
 // --- find_translation_gaps.
 //
-// A node is a gap when the target locale has no translation at all, or has one
-// that is still a draft. Those need different fixes — write a translation vs flip
-// `draft` — and `list_sections?locale=…` answers neither question. Measured on a
-// live tenant with an admin token (#225 validation): a node with no translation is
-// absent from that listing, but a node whose translation is a DRAFT is still
-// returned, rendered with the draft's own name. So absence there is ambiguous and
-// presence does not mean published, which is why this audit probes the
-// translations endpoint per node instead of diffing the two listings.
+// A gap is no translation at all, or a draft one — different fixes, and
+// `list_sections?locale=…` shows neither. Measured with an admin token (#225): a
+// missing translation is absent there, a DRAFT one is still listed under its
+// draft name. So absence is ambiguous and presence is not "published", hence the
+// per-node probe rather than a listing diff.
 type GapReason = 'missing' | 'draft';
 
 interface TranslationGap {
@@ -282,10 +261,8 @@ const renderGapLines = (
       ...gaps.map((gap) => `- **${gap.name}** (${gap.id}) — ${GAP_REASON_TEXT[gap.reason]}`),
     ];
   }
-  // "every one scanned has a translation" would be a wrong reading of an empty
-  // scan (a category with no sections, a scope that matched nothing) — and
-  // "nothing to scan" would be a wrong reading of a level the node cap never
-  // reached, which is the opposite of an all-clear.
+  // Three distinct states, and conflating any two reads as a false all-clear:
+  // nothing at this level, nothing left after the cap, nothing missing.
   if (scanned === 0) {
     return [
       header,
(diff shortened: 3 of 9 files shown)


dlecan:
## Functional validation report — aee3d5efc92bcaaec18075f716af4f3bd508d87d

Independent validator run, on the PR branch against the live tenant. Tenant values are redacted to `{placeholders}`; numeric ids are kept where a scenario needs them for reproduction, never paired with the localized text they belong to.

**Environment as read from `zendesk-hc://topology`**: default locale `fr`, active locales `en-us` + `fr` — so `{locale}` = `en-us`. Tree size: 13 categories, 45 sections = 58 nodes, which sits **below** the default 60-node cap, so the cap and its "left unscanned" wording were never exercised on this tenant (the plan assigns that to unit tests).

| ID | Verdict | Evidence |
| -- | ------- | -------- |
| S1 | OK | Two sections cross-checked against `list_sections` (default locale). Section `35103402757789`: `list_sections` name/description → translation `fr` entry shows the **same** `**Name**` and `**Description**: set`. Section `19918665202845` (`No description` in `list_sections`) → `**Description**: empty`. `title`=name / `body`=description holds. |
| S2 | OK | `list_section_translations` (`19918665202845`) and `list_category_translations` (`19918231873565`) each return one block per locale with `**Name**` / `**Description**` / `**Draft**` / `**Updated**`, `Draft: false` on published entries. No `**Title**` label anywhere — that label stays on `list_article_translations`, as intended. |
| S3 | **FAIL** | Category half is exact: 13 scanned, 3 reported as `no translation`, and `list_categories({locale:"en-us"})` returns exactly the other 10. Section half is off by one node: of 45 scanned, 18 reported and 27 not, and all 27 are present in `list_sections({locale:"en-us"})` — but that listing returns **28** sections. The extra one is `25471740471197`, which the audit reports as `draft translation (not published)`. Per the plan's cross-check ("every section it reports … is absent from `list_sections`"), that is a mismatch. Root cause below — the audit's classification is correct; the premise about `list_sections` is not. |
| S4 | OK | `25471740471197` → `en-us` entry with `Draft: true` (matches `draft translation`). `9816536711069` → **no** `en-us` entry at all, `fr` only (matches `no translation`). |
| S5 | OK | Section `37276895449117` (only `fr` beforehand, confirmed by `list_section_translations`): `set_section_translation({section_id, locale:"en-us", name:"Getting started with AI Governance", description:"{probe description}"})` → `Translation created for section #37276895449117 in "en-us" (published).` Then `list_section_translations` shows a new `en-us` entry, `Draft: false`, `Description: set`. Then `list_sections({locale:"en-us", category_id:29193628989213})` returns the section under the new English name with the description — the user-facing point of the PR, confirmed against the real API. |
| S6 | OK | Same section, `set_section_translation({…, name:"Getting started"})` with **no** `description`, **no** `draft` → `Translation updated for section #37276895449117 in "en-us" (published).`, name changed, `Description` still `set`. Independently re-read through `list_sections({locale:"en-us"})`: the S5 description survived verbatim. |
| S7 | **FAIL** (partial) | `draft: true` → `Translation updated … (published)` replaced by `(draft, not visible to end users)`, `Draft: true`, and the audit re-classifies the node as `draft translation (not published)` — both as specified. But the section did **not** disappear from `list_sections({locale:"en-us"})`: it stayed listed, still rendered with its *draft* localized name and description. `draft: false` → `(published)`, node drops out of the audit again. So `draft` is the publish switch and the tools report it correctly; the plan's `list_sections` observable is the part that fails, same root cause as S3. |
| S8 | OK | `set_section_translation({section_id:37276895449117, locale:"en-us"})` and nothing else → error `Nothing to write: section #37276895449117 already has a "en-us" translation, so pass at least one of "name", "description" or "draft" to change it (draft: false publishes it).` No "Translation updated" message. |
| S9 | OK | On section `32303268073501` (no `en-us` translation): `set_section_translation({section_id, locale:"en-us"})` → error `section #32303268073501 has no "en-us" translation yet, so one has to be created and "name" is required. Pass the localized name, or call list_section_translations to see which locales already exist.` Nothing created — `list_section_translations` still shows `fr` only. |
| S10 | OK | `find_translation_gaps({locale:"en-us", category_id:14560863527965})` → exactly `## Categories (1 scanned)` with that category alone and `## Sections (2 scanned)`, its own two sections. With a nonexistent category id → `Resource not found. Please verify the ID is correct. (Not Found)`, not an empty report. |
| S11 | OK | `find_translation_gaps({locale:"zz"})` → `> ⚠ "zz" is not an active locale of this Help Center (active: en-us, fr), so everything below reads as untranslated. Check the spelling, or activate the language in Guide first.` above the all-missing listing (13 + 45 nodes). Warning, not a catastrophe verdict. |
| S12 | Read half OK / write half **SKIPPED** | `list_category_translations` validated under S2 (both locales, correct labels), and the create/publish path is exercised category-side by no scenario: the tenant owner ruled out category writes — there is no expendable category on this production tenant. Write half not run, per the plan's own escape clause. |
| S13 | OK | `zendesk-hc://topology` carries the same four sections (`Locales`, `Categories → sections`, `Visibility`, `Permission groups`) and **no** translation-coverage section. `list_sections`, `list_categories`, `list_article_translations` (still `**Title**`, `Draft`, `Updated`) and `compare_translations` (Freshness / outdated flag / structure / per-section table) all render as before. Corroborated statically: `git diff origin/main...HEAD -- src/` is **600 insertions, 0 deletions** across `src/constants.ts`, `src/tools/help-center.ts`, `src/utils/formatting.ts`; the shared-file diffs are a new `TRANSLATION_GAP_SCAN_MAX_NODES` constant and a new `formatNodeTranslationSummary` function, no existing symbol touched, and no resource/instructions file in the diff. |

### The failing observable — `list_sections(locale)` does list draft-translated sections

Observed twice, once on pre-existing data and once deliberately reproduced:

1. Section `25471740471197` has an `en-us` translation with `Draft: true` (confirmed by `list_section_translations`), is reported by `find_translation_gaps` as `draft translation (not published)`, **and** is returned by `list_sections({locale:"en-us"})`.
2. S7 reproduced it on demand: flipping section `37276895449117` to `draft: true` left it in `list_sections({locale:"en-us"})`, rendered with the draft's own localized name and description — so the listing cannot even be used to infer draft state from a fallback-to-source-locale name.

What the new tools *compute* is right; what their descriptions *claim* is not. Three places state the premise unconditionally:

- `list_section_translations` / `list_category_translations`: "list_sections with a locale omits a section that has no translation AND one whose translation is an unpublished draft, and only this tool tells the two apart."
- `find_translation_gaps`: "that gap is invisible to list_sections".
- The PR summary makes the same claim.

Half of it is verified true: a node with **no** translation in the locale is indeed omitted — checked across all 45 sections and all 13 categories, the omission sets match the audit exactly apart from the one draft node. The draft half is false for the caller used here, an **admin**-authenticated token (role `admin`, per topology). I could not test with an end-user or anonymous token, so whether Zendesk hides draft translations from unprivileged callers is **unverified** — plausible, and it would explain the behaviour, but the descriptions do not say "for end users", and an agent reading them will conclude "present in `list_sections(locale)` ⇒ published", which is wrong on exactly the path these tools exist to fix.

Suggested resolution (author's call): narrow the claim to what holds unconditionally (absent translations are omitted; a draft translation may still be listed, and the listing cannot tell it from a published one), or qualify it by caller role once the end-user behaviour is confirmed. No code change is implied by this finding — the audit, the classification and the writes are all correct.

### State left on the tenant

The plan asked for a *throwaway* node and a restore. The tenant owner redirected: these sections are genuinely due an English translation (English articles are already published under them), so the write scenarios ran on a real node and the final state is **intended, not a leftover**.

- Section `37276895449117`: had `fr` only; now has a **published** `en-us` translation named `Getting started`. The `{probe description}` written in S5 was cleared afterwards with a final `description: ""` call, aligning it with the `fr` node which has none — that call also exercised the documented clear path (`Description` went `set` → `empty`), which no scenario covered.
- Section `32303268073501`: **untouched**. S9 only exercised the refused-create path; verified afterwards that it still has `fr` only.
- No category was written.
- Nothing needs manual removal in Guide.

One transient `Zendesk authentication required` re-auth prompt occurred mid-run and resolved on the next retry; no scenario was blocked by it.

## Summary

11 of 13 scenarios OK, S12 read half OK with its write half skipped for lack of expendable category test data. **S3 and S7 FAIL on the same single point**: `list_sections(locale)` lists sections whose translation in that locale is an unpublished draft, so the "only this tool tells the two apart" / "invisible to `list_sections`" claim in three tool descriptions and in the PR summary is not what the real API does for an admin caller. Everything the plan set out to prove about the implementation itself holds: `title`/`body` really are the localized name and description (S1), the audit's absent-vs-draft classification matches per-node reads (S3 category half, S4), create/update/publish/clear all behave as documented and only write the fields passed (S5–S7), and the guard rails refuse no-op and nameless-create writes with actionable messages (S8, S9).

dlecan:
Thanks — S3/S7 is a real finding and it is now fixed in `2140595`.

**The claim was half wrong, and your evidence pins which half.** I checked both observations: pre-existing node `25471740471197` (draft `en-us`, reported as *draft* by the audit, yet returned by `list_sections({locale:"en-us"})`) and the on-demand reproduction in S7 on `37276895449117`. Two independent observations, one deliberately reproduced, and the reproduction rules out a stale-cache explanation. So:

- **absent ⇒ no translation**: verified true, and you checked it exhaustively (45 sections, 13 categories, omission sets matching the audit exactly apart from the one draft node).
- **draft ⇒ absent**: false for an admin caller.

Corrected in the three tool descriptions, the gap-classifier comment, the scan-cap constant comment and the cost note in `docs/configuration.md`. The new wording states only what was measured and attributes it to an admin token, since the unprivileged-caller behaviour was not testable and I am not claiming it either way. A test pins all three descriptions so the old phrasing cannot creep back.

**No logic change was needed, and that is not luck.** `find_translation_gaps` probes the translations endpoint per node rather than diffing the full and locale-filtered listings. The diff was the cheaper design and I considered it while planning; I rejected it precisely because it would have rested on this undocumented behaviour. Your finding is what that would have cost: a listing-diff audit would have missed **every** draft-translated node — the exact case the issue was opened about. The cost of the safer choice is the one-request-per-node fan-out that the cap and the wave size exist to bound.

Worth recording that **issue #224's own premise was partly wrong** in the same way: it stated that a section absent from the target-locale listing "may have no translation at all, or one that exists but is still a draft". Per your measurement, a draft one is not absent. The issue's conclusion still holds — the listing can't answer the question — but for the opposite reason on that half.

Two notes on your run, both fine by me:

- **S12 write half skipped**: agreed, no expendable category on a production tenant. Category writes share `upsertNodeTranslation` with the section path end-to-end — only `NODE_LABEL` and the URL segment differ — so S5–S9 passing section-side plus the S2/S10 category reads is the most that can be had here. Not worth manufacturing a throwaway category for.
- **Tenant state**: noted that the write scenarios ran on a genuinely-needed translation rather than a throwaway, at the owner's direction, and that `37276895449117` is intentionally left with a published `en-us` translation. Thanks for exercising the `description: ""` clear path on the way out — no scenario asked for it and it is a documented behaviour that had only unit coverage.

The cap wording was never exercised (58 nodes vs a 60 default), as the plan anticipated; it stays on the unit tests, including the one that observes peak in-flight requests for the wave bound.

---

dlecan:
S12's write half stays unvalidated in live — accepted, the category path ships as is. Taking the iso rule seriously changes what that costs, so `1267a5f` closes the gap it left.

**The two levels were iso in the code but not in the tests.** 9 write cases and 2 read cases on the section side, against 3 and 1 on the category side. The category path was effectively riding on "the shared `upsertNodeTranslation` makes its own coverage redundant" — which is precisely the assumption that cannot be leaned on when the live pass for that half was skipped. The unit tests are the only thing behind it.

Both read and write cases now run over one `NODE_LEVELS` table, so each is asserted on section **and** category from a single definition, and a new case cannot land on one level only. Parity is 12/12. Category-side coverage added for the POST/PUT path actually built, only-the-fields-passed on update, description defaulting and explicit clearing, the locale-casing normalization, the refused no-op update and the draft-state wording. Two cases are new on both levels: `draft: true` honoured on creation, and the draft flag surfaced by the list tools. The integration suite gains a category write over the namespace proxy next to the section one.

I also audited the two tool definitions for real asymmetry by normalizing the section/category vocabulary and diffing: they are byte-identical apart from one intentional difference — the category tool says "where its **sections or** articles are already translated". The iso rule is now stated on `TreeNodeKind` where a future change will see it.

705 tests, `pnpm check` / `typecheck` / `build` green, coverage unchanged.

One thing this does **not** buy, and I would rather name it than let the green ticks imply otherwise: every category assertion runs against MSW, so it proves the request we build and the text we render, not that Zendesk answers `POST /categories/{id}/translations` the way it answers the section one. S1 confirmed the `title`/`body` mapping on a live section; the equivalent for a category rests on the two endpoints being the same family, which the API reference states but no call in this PR has demonstrated. If a category write ever misbehaves in production, that is the assumption to suspect first.

---

dlecan:
## Functional validation report (re-run) — 21405952bd237e3159fb1bf7aabb63ab1dfbd9e1

Second independent run, after `2140595`, against the live tenant with the MCP server restarted on this commit. Fresh report rather than an edit of [issuecomment-5255363078](https://github.com/fruggr/zendesk-mcp-server/pull/225#issuecomment-5255363078). Tenant values redacted to `{placeholders}`; numeric ids kept where reproduction needs them, never paired with the localized text they belong to.

**Criterion used for S3 / S7.** The amended plan keeps both scenarios' original wording and annotates the disproven half, stating that "a re-run should expect a draft-translated node to remain listed". That is the expectation applied here — a draft-translated node **remaining** in `list_sections({locale})` is now an OK, not a FAIL.

**Corrected wording is what the server actually serves.** All three descriptions re-fetched from the restarted server carry the `2140595` text, e.g. `find_translation_gaps`: "a node with no translation is simply absent, without saying why, and a node whose translation is an unpublished draft may still be listed under its draft name — so this audit reads the draft flag on each node instead of trusting that listing." `list_section_translations` and `list_category_translations` carry the symmetric per-level version. No description still claims the listing omits a draft.

| ID | Verdict | Evidence |
| -- | ------- | -------- |
| S1 | OK | Re-checked on two sections. `35103402757789`: `fr` translation `**Name**` equals the `list_sections` name, `**Description**: set` against a non-empty description. `19918665202845` (`No description`) → `**Description**: empty`. `title`=name / `body`=description still holds. |
| S2 | OK | `list_section_translations` (`35103402757789`) and `list_category_translations` (`19918231873565`): one block per locale, `**Name**` / `**Description**` / `**Draft**` / `**Updated**`, `Draft: false` on published. No `**Title**` label. |
| S3 | OK | 13 categories scanned, 3 reported `no translation`; `list_categories({locale:"en-us"})` returns exactly the other 10. 45 sections scanned, 17 reported = 16 `no translation` + 1 `draft translation (not published)` (`25471740471197`). `list_sections({locale:"en-us"})` returns 29 = the 28 unreported ones **plus** the draft node, and none of the 16 `no translation` ids appear in it. Exact match against the amended criterion. Footer arithmetic consistent: `20 node(s)` = 3 + 17. |
| S4 | OK | `25471740471197` → `en-us` entry `Draft: true`. `9816536711069` → no `en-us` entry, `fr` only. Matches both classifications. |
| S5 | OK | Section `37276911319837`, `fr` only beforehand (captured). `set_section_translation({…, name:"Governing", description:"Governance probe"})` → `Translation created for section #37276911319837 in "en-us" (published).`, `Draft: false`, `Description: set`. `list_sections({locale:"en-us", category_id:29193628989213})` then returns it under the English name with the description. |
| S6 | OK | Same section, `{…, name:"Govern"}` only → `Translation updated … (published)`, name changed, `Description` still `set`. Verified independently through `list_sections({locale:"en-us"})`: the S5 description is still rendered verbatim after the name-only write. |
| S7 | OK | `draft: true` → `Translation updated for section #37276911319837 in "en-us" (draft, not visible to end users).`, `Draft: true`, and `find_translation_gaps` re-classifies the node as `draft translation (not published)`. The section **stays** in `list_sections({locale:"en-us"})`, rendered under the draft's own name — the now-documented behaviour, per the amended expectation. `draft: false` → `(published)` and the node drops out of the audit again. |
| S8 | OK | `set_section_translation({section_id:37276911319837, locale:"en-us"})` alone → `Nothing to write: section #37276911319837 already has a "en-us" translation, so pass at least one of "name", "description" or "draft" to change it (draft: false publishes it).` No update message. |
| S9 | OK | `set_section_translation({section_id:32303268073501, locale:"en-us"})` → `section #32303268073501 has no "en-us" translation yet, so one has to be created and "name" is required. Pass the localized name, or call list_section_translations to see which locales already exist.` Nothing created — re-read shows `fr` only. |
| S10 | OK | `category_id:14560863527965` → exactly `## Categories (1 scanned)` with that category and `## Sections (2 scanned)` with its own two. Nonexistent id → `Resource not found. Please verify the ID is correct. (Not Found)`, not an empty report. |
| S11 | OK | `locale:"zz"` → `> ⚠ "zz" is not an active locale of this Help Center (active: en-us, fr), so everything below reads as untranslated. Check the spelling, or activate the language in Guide first.` above the all-missing listing (13 + 45). |
| S12 | Read half OK / write half **SKIPPED** | Read half re-run under S2. Write half still skipped: no expendable category on this production tenant (owner's call, unchanged since the first run). |
| S13 | OK | `zendesk-hc://topology` unchanged — same four sections (`Locales`, `Categories → sections`, `Visibility`, `Permission groups`), no translation-coverage section. `list_sections`, `list_categories`, `list_article_translations` (still `**Title**` / `Draft` / `Updated`) and `compare_translations` (Freshness / outdated flag / structure / per-section table) all render as before. `git show --stat 2140595` touches only `docs/configuration.md`, a comment in `src/constants.ts`, three descriptions + the gap-classifier comment in `src/tools/help-center.ts`, and the test file — no logic. `pnpm vitest run tests/unit/tools/help-center.test.ts` → 116 passed. |

### Extra path exercised (not in the plan)

`description: ""` on an existing translation cleared it: `Description` went `set` → `empty` while name and draft state stayed put, matching the schema's "pass an empty string to clear it". No scenario covered the clear path; it works.

### On the previous report's quotation

My first report quoted the pre-fix wording as the text of the tool descriptions. After the restart I cannot re-establish which wording the server served me at the moment I fetched those schemas — the schema blocks in my session now show the corrected text. The quotation matches, word for word, the text `2140595` removes, so the commit's own diff corroborates it; but treat the diff as its provenance, not a live observation of mine. The behavioural finding it supported was measured directly, twice, and is unaffected.

### State left on the tenant

Both write scenarios ran on sections genuinely due an English translation (English articles already published under them), per the tenant owner's instruction, so the end state is intended.

- Section `37276895449117` (first run): published `en-us` translation, name `Getting started`, no description. Untouched this run.
- Section `37276911319837` (this run): had `fr` only; now a published `en-us` translation, name `Govern`, description cleared to match the `fr` node which has none.
- Section `32303268073501`: untouched — S9 only exercises the refused-create path; re-verified `fr` only afterwards.
- No category written. Nothing to remove manually in Guide.

## Summary

Green. 12 of 13 scenarios OK, S12's write half skipped for lack of expendable category test data on a production tenant — unchanged and not a defect. The two FAILs from the first run are resolved: the behaviour they measured is unchanged (it was never a logic bug), and the descriptions, the classifier comment, the scan-cap comment and the `docs/configuration.md` cost note now state what the API actually does, which is what an agent reading them needs. The remaining gap in evidence is one the correction itself acknowledges rather than papers over: whether an unprivileged caller sees draft-translated nodes in a locale-filtered listing was not testable here, and the new wording no longer claims it either way.

dlecan:
Re-run accepted, and I checked the API docs on the section/category symmetry rather than leaving it as an assumption. That check found something worth a follow-up.

**On the re-run being against `2140595` while head is now `2194ad5`.** Not stale in any way that matters, and I verified rather than assumed it: `git diff 2140595 2194ad5 -- src/`, with comment lines filtered out, is **empty**. Everything since is the both-levels test table, a category write in the integration suite, and two comment blocks. Nothing the report measured can have moved, so no third run is needed.

**Spot-checking your S3 arithmetic** — it reconciles across both runs, which is what makes it convincing. Run 1: 18 sections reported, 27 unreported, listing returned 28. Run 2: 17 reported, 28 unreported, listing returned 29. The delta of exactly one is the section your own run-1 S5 published and left published (`37276895449117`). 45 = 17 + 28 ✓, 29 = 28 + 1 draft ✓, footer `20` = 3 + 17 ✓.

**Thank you for retracting the quotation.** You had no way to re-establish which wording the server served you once the session restarted, and saying so is worth more than the quote was. It changes nothing here: the wording you quoted is verbatim what `2140595` deletes, so the diff is its provenance, and the behavioural finding was measured directly on the wire, twice.

### The doc check: section/category symmetry confirmed, as far as documentation can

- **Endpoints** identical in shape for list / show / create / update, `sections/{id}/translations` ↔ `categories/{id}/translations`.
- **Translation object** the same for both, `title` mandatory on create for both.
- **Section vs Category objects** document `name` and `description` identically — `name` writable and mandatory, `description` writable and optional, plus `locale` and `source_locale` on both.
- What stays undocumented is undocumented **for both equally**: neither page states that `name` is backed by the translation's `title` and `description` by its `body`. So the gap in evidence is not an asymmetry between the two levels, which is what the iso assumption actually needed. S1 pinned the mapping on a live section; the category side rests on the two being one endpoint family, which the reference asserts.

### What the check turned up — filed as #226

Both **list** endpoints document a `translations` sideload — sections: *"the section and category translations, if any"*, categories: *"the category translations, if any"* — on the list endpoints, not only show-by-id. If it behaves as named, `GET /sections?include=translations&page[size]=100` plus the same for categories returns the whole tree with its translations in **two** requests, collapsing the per-node fan-out, the waves, and `ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES` entirely.

I am not taking it in this PR. The docs never say which fields a sideloaded translation carries, and `draft` is the whole point of the audit — adopting it on the strength of a sideload's *name* would be the same class of mistake as the claim your S3 just caught. It needs one live call to settle, which I cannot make from here. #226 has the measurement steps, starting with "does the sideloaded object carry `draft`? if not, close as not-viable", and asks for the raw payload in the issue before any redesign. `2194ad5` records the lead next to the cost it would remove, so it is not rediscovered from scratch.

That keeps this PR as validated: the per-node route is the one *known* to return the flag.

Note: CodeRabbit has been rate-limited since `1267a5f` and has not reviewed the last two commits (both comment/test-only). Its last completed review was clean on `2140595`. I have a check-in armed to request one once its limit resets.

---

dlecan:
@coderabbitai review

Requesting a review of the head commit: the automatic runs for `1267a5f` and `2194ad5` were both refused on the per-developer rate limit, so the last completed review is on `2140595`. The two unreviewed commits are the both-levels test table plus an integration scenario, and two comment blocks — no logic change.

---

docs/mcp-tools-reference.md
```diff
@@ -48,6 +48,9 @@ out every `write` tool before the proxies are built.
 | `list_articles` | List articles with sorting and translation info | read |
 | `list_promoted_articles` | List the promoted ("featured") articles (scans + filters client-side, page-capped) | read |
 | `list_article_translations` | List available translations for an article | read |
+| `list_section_translations` | List a section's translations with their draft state, telling an untranslated section apart from an unpublished one | read |
+| `list_category_translations` | List a category's translations with their draft state, telling an untranslated category apart from an unpublished one | read |
+| `find_translation_gaps` | Audit the category → section tree for a locale and report every node with no translation or an unpublished draft (one request per node scanned, capped) | read |
```
dlecan: Question: why just "categories" and not "sections" too?

src/constants.ts
```diff
@@ -45,6 +45,27 @@ export const ARTICLE_RESOURCES_SCAN_MAX_PAGES = positiveIntEnv(
   20,
 );
 
+// Max category + section nodes probed by find_translation_gaps. Unlike articles,
```
dlecan: Nitpick: shorter comment please

src/tools/help-center.ts
dlecan, line 1: Nitpick: shorten all comments in this file

docs/mcp-tools-reference.md
dlecan: Second reading: it handles both, so the row was too restrictive. Fixed in `f204c29`.

"Audit the category → section tree" parses as *a category's* section tree, i.e. category-scoped — which is exactly the wrong impression. The row now says it audits every category **and** every section, and that `category_id` is what narrows it.

The generic name stays right: the tool reports both levels, and the two `list_*_translations` / `set_*_translation` pairs already carry the level in their names when it matters.

Worth noting the tool's own `description` — the text an LLM actually reads — was already explicit: *"report every category and section that has no translation, or one that is still an unpublished draft"*. So this was a human-facing docs gap, not a behaviour an agent would have got wrong. That the two drifted apart is the lesson: the table is a second copy of the same claim.

---

src/constants.ts
dlecan: Done in `a143b66` — 19 lines to 6.

It was carrying the whole argument for the per-node cost plus the full rationale for not taking the sideload. The cost claim stays (it is why the constant exists); the sideload paragraph is now a pointer to #226, which is where that reasoning belongs anyway.

---

src/tools/help-center.ts
dlecan: Done in `a143b66` — the comments this PR adds went from 54 lines to 27.

They were arguing a case where a claim was enough. What I kept is the part a reader cannot re-derive from the code: the `title`=name / `body`=description mapping, that a draft node is still listed in a locale-filtered listing (measured, #225), why the locale filter is lower-cased, and why the burst is bounded. What went: the reasoning leading to each, which is in the commits and #226.

**One deliberate narrowing, flag it if you meant otherwise.** You said "all comments in this file"; I only touched the ones this PR introduced. Reflowing the pre-existing ones — the `compare_translations` signal blocks, `autoSortNotice`, the `localePrefix` composition note, `scanCostNote` — would add churn across a file this PR barely touches otherwise and bury the actual change. Happy to do it as a separate PR if you want the whole file tightened; it is a real improvement, just not one that belongs in this diff.

---

PR #228, opened 2026-08-11
perf(help-center): classify translation gaps from the listing sideload (#226)

### LINKED ISSUE #226: find_translation_gaps: try the `translations` sideload to drop the per-node fan-out
### Context

`find_translation_gaps` (added in #225, closes #224) audits the category → section tree for a target locale and reports every node with no translation or an unpublished draft. It gets the draft state by calling `GET /{sections|categories}/{id}/translations` **once per node**, in waves of 5, bounded by `ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES` (default 60). On a 58-node tenant that is ~61 requests for one audit, and past the cap the report has to admit it left nodes unchecked.

That design was chosen because a locale-filtered listing cannot answer the question — validated on a live tenant in #225: a node with no translation is absent from it *without saying why*, and a node whose translation is a **draft** is still returned, under the draft name.

### The lead

Both list endpoints document a `translations` sideload:

- Sections — `translations`: "the section and category translations, if any"
- Categories — `translations`: "the category translations, if any"

Documented on the **list** endpoints, not just show-by-id (`List Sections`, `List Sections in Category`, `List Sections by Locale`, `Show Section`, …). If it behaves as named, then

```
GET /api/v2/help_center/sections?include=translations&page[size]=100
GET /api/v2/help_center/categories?include=translations&page[size]=100
```

returns the whole tree *with* its translations in **two** requests — collapsing the fan-out, the wave logic, and `ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES` along with the "left unscanned" wording in the report and the cost note in `docs/configuration.md`.

### What has to be measured first

The docs do **not** state which fields a sideloaded translation carries. `draft` is the entire point of this audit, so:

1. On a live tenant, call `GET /api/v2/help_center/sections?include=translations&page[size]=100` and inspect one sideloaded translation object. **Does it carry `draft`?** If not, stop — the sideload cannot replace the per-node read and this issue closes as not-viable.
2. If it does: confirm it carries every locale's translation, not only the current/source one (a node with a `fr` translation and no `en-us` one must be distinguishable from one with both).
3. Check the shape: sideloads are usually a top-level `translations` array keyed back by `source_id`, not nested per node — the classifier would group by `source_id` / `source_type`.
4. Confirm it holds under pagination, and that `page[size]=100` with the sideload does not get throttled or truncated differently.
5. Check whether the locale-prefixed variant (`/{locale}/sections?include=translations`) changes what is returned — irrelevant if the unprefixed listing already carries all locales.

Worth capturing the raw JSON in the issue before any code changes, since that payload is the whole basis for the redesign.

### If viable

- `find_translation_gaps` fetches the two listings with `include=translations` and classifies from the sideload; `classifyGap` and the report are unchanged (they already take a translations array).
- `TRANSLATION_GAP_SCAN_MAX_NODES`, `GAP_SCAN_WAVE_SIZE` and `probeInWaves` go away, along with the cap note in the report and the cost note in `docs/configuration.md`.
- `listNodeTranslations` stays — `list_section_translations` / `list_category_translations` still need the per-node endpoint.
- The `has_more` "only the first page was considered" note stays relevant, and becomes the *only* coverage caveat.
- Removing a documented env var is user-visible: keep reading it and ignore it, or note the removal in the release notes.

### Not urgent

The current implementation is correct and twice validated end-to-end (#225); this is a cost and simplicity win, not a fix. The per-node route is also the one *known* to return `draft`, so it stays the fallback if step 1 comes back negative.

dlecan:
## Summary

`find_translation_gaps` read the `draft` flag with one `GET /{sections|categories}/{id}/translations` per node (waves of 5, capped at `ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES`), so a 58-node tenant cost ~61 requests and the report had to admit what it left unchecked. The `translations` sideload on both list endpoints now carries that flag instead: the audit classifies straight from the two listings it already fetched, whatever the tree size.

The issue gated this on a live measurement, since the docs never state which fields a sideloaded translation carries. That measurement ran and is posted in [#226](https://github.com/fruggr/zendesk-mcp-server/issues/226#issuecomment-5272129090): `draft` is present on 100% of sideloaded translations at both levels, and a field-by-field cross-check against the per-node endpoint agrees on every sampled node, **including a real draft** created for the purpose and deleted afterwards.

The read-only probe that produced that evidence lived on this branch, ran, and is removed again at review request: it was step 1's instrument, not a lasting tool, and its payload belongs in the issue rather than in `scripts/`. So the net diff here is the redesign alone.

## Linked issue

Closes #226

## Type of change
- [ ] Bug fix
- [ ] New feature
- [x] Documentation (the no-regression rule on tool descriptions, see below)
- [ ] Tooling / CI

## What changed

`ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES`, `GAP_SCAN_WAVE_SIZE` and `probeInWaves` are gone, along with the cap note in the report and the env var's section in `docs/configuration.md`. The variable is no longer read at all. Tree size no longer bounds the audit, so the `has_more` "only the first page was considered" note is the only coverage caveat left.

Two deliberate calls beyond the issue's outline:

**Still two listings, not one.** `include=categories,translations` on `/sections` does return categories with their own translations (measured), which would collapse the audit to a single request. It only returns the categories *referenced by the sections on the page*, though, so a category holding no section would silently vanish from the audit. A missed category is exactly the gap this tool exists to find, so the second listing stays.

**One rule amended.** `docs/mcp-metadata.md` forbade shortening a tool or param description, "a stricter agent may key off exact text". Enriching, which the same rule waves through, changes that exact text just as much — so the rationale did not support the rule, while the ratchet it created collided with the `Conciseness & Structure` dimension documented sixty lines above it, and left no way to delete a sentence that had become false. Descriptions are now held to the facts they carry. This PR is its first customer twice over: the "one request per node, capped" sentence had to go, and the replacement I wrote first grew the description by 56% before being cut back.

**An absent sideload is not a missing translation.** Zendesk drops an `include` it doesn't know silently, with a 200 and no key. Reading that as "no translation" would report an entire Help Center as one big gap, so nodes whose `translations` key is absent are counted as unclassified, named as such in the report, and left out of the gap list.

## Author checklist
- [x] `pnpm test` passes locally (34 files, 770 tests, rebased on `main`)
- [x] `pnpm typecheck` passes
- [x] I ran a Claude Code review on the diff and addressed its findings (one taken, one declined on evidence — see the comments)
- [x] Documentation is updated where needed (`docs/mcp-tools-reference.md` cost column, `docs/configuration.md` env var removed)
- [x] If this PR adds an MCP tool: n/a, no tool added
- [x] If the change has **MCP-runtime-observable** behaviour: plan below

### Context

The change is entirely inside `find_translation_gaps`. Exercise it through `mcp__zendesk-local__find_translation_gaps`; every other tool is untouched. The transport does not matter (the tool is transport-agnostic), so validating over the session's existing stdio connection is enough.

Two properties are already locked down by unit tests and need no runtime re-proof: that the audit issues **zero** per-node `/{kind}/{id}/translations` calls, and that both listings are asked for `include=translations` (`tests/unit/tools/help-center.test.ts`). What runtime validation adds is whether the report tells the **truth about a real tenant**, which no mock can settle.

### Setup & prerequisites

Nothing to build or configure. If the session's tools predate this branch's commit, call `mcp__zendesk-local__reload_tools` first so the handler under test is the one loaded. Record `git rev-parse HEAD` in the report.

S3 writes to the tenant. It adds a **draft** translation in a locale where the section has none, so nothing that is currently visible to end users changes: drafts are invisible by definition, and the node was already reported as a gap in that locale. Deleting it afterwards needs `DELETE /api/v2/help_center/translations/{id}`, which no MCP tool exposes; if you cannot issue that call, say so in the report and leave the draft named as it is rather than publishing it.

### Scenarios

| id | Validates | Do this | Expected |
|----|-----------|---------|----------|
| S1 | The audit covers the whole tree, uncapped | `find_translation_gaps { locale: "fr" }` | Report headers count **every** category and section of the tenant (cross-check against `zendesk-hc://topology`). No `60-node cap`, no `left unchecked`, no `ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES` anywhere in the output |
| S2 | The verdicts are true, not just plausible | Take 3 nodes from S1: one reported as a gap, one not reported, one of each kind if possible. Call `list_section_translations` / `list_category_translations` on each | Per-node truth matches the report every time: a node listed as "no translation" has no `fr` entry; a node listed as "draft" has `fr` with `Draft: true`; an unreported node has `fr` with `Draft: false` |
| S3 | A draft is reported as a draft, not as "no translation" | Take a section S1 reports as `no translation` in some locale L. `set_section_translation { section_id, locale: L, name: "TEMP …", draft: true }`, re-run the audit for L, then delete that translation and re-run | While drafted: S flips to `— draft translation (not published)`. After deleting: S is back to `no translation` and the rest of the report is byte-identical |
| S4 | Scoping still works, and still reads the scoped category itself | `find_translation_gaps { locale: "fr", category_id: <a category with sections> }` | Only that category and its own sections appear. The category itself is classified (it is fetched directly, not filtered out of a listing) |
| S5 | An inactive locale warns instead of crying gap | `find_translation_gaps { locale: "es" }` | The `is not an active locale` warning is present and lists the tenant's active locales. Everything below reads as untranslated, as the warning says |
| S6 | Locale casing | `find_translation_gaps { locale: "FR" }` | Same gap list as S1, and **no** inactive-locale warning |
| S7 | The cost actually dropped | The tool output cannot show this (a tenant under the old 60-node cap rendered the same report either way), so measure the two strategies directly against the tenant with a read-only script: three requests with the sideload, versus the two listings plus one per node in waves of five | The sideload path issues **3** requests where the per-node path issues `2 + nodes + 1`, and classifies every node from the listing |

### Long-running behaviours

None. Nothing here is time-based, and no scenario needs a wait or a fake timer.

### Priorities

S1 and S2 first: they are the user-facing path and the only check that the sideload tells the truth on a real tenant. S3 next, since a draft read as published is the one failure mode that would make this change worse than what it replaces. S4 to S7 after.

### Reporting

Post one PR comment (English) with the tested SHA, a per-id `OK` / `FAIL` / `BLOCKED` verdict with the observed evidence (the report text, the per-node translation listings, timings), and an overall green-or-failing summary. For S3, state explicitly that the section was restored to `draft: false`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hjs9kdjNJA2KM9wXpbSBmy)_


dlecan:
## Functional validation report — `e87363b`

**Caveat on independence:** this run was executed by the implementer, at the repo owner's explicit request, not by an independent validator. Read the verdicts with that in mind; every one of them is backed by evidence below, and an independent re-run is still worth having.

Environment: live `fruggr` tenant (13 categories, 45 sections, active locales `en-us` and `fr`), stdio transport, `--mode all`, tools refreshed with `reload_tools` before the first call (the reloaded tool description matched this branch's, confirming the handler under test was the new one).

| id | Verdict | Evidence |
|----|---------|----------|
| S1 | OK | `find_translation_gaps { locale: "fr" }` reported `## Categories (13 scanned)` and `## Sections (45 scanned)`, i.e. all 58 nodes of the tree. No `60-node cap`, no `left unchecked`, no `ZENDESK_TRANSLATION_GAP_SCAN_MAX_NODES` in the output |
| S2 | OK | Three verdicts checked against the per-node endpoint. Section 36166963504285, reported `no translation` in `en-us`: `list_section_translations` returns `fr` only. Category 16301205121949, same verdict: `list_category_translations` returns `fr` only. Section 19918665202845, not reported: `en-us` present with `Draft: false`. The three `en-us` category gaps also match, independently, the three single-locale categories a read-only script found while preparing #226 |
| S3 | OK | Section 36166963504285 read `no translation` in `en-us`. After `set_section_translation { locale: "en-us", draft: true }` it read `— draft translation (not published)`. After deleting that translation (HTTP 204) it read `no translation` again and the rest of the report was identical, 12 gaps before and after |
| S4 | OK | `{ locale: "en-us", category_id: 16301205121949 }` returned exactly `1 scanned` category and `1 scanned` section, the scoped category itself classified (`no translation`) alongside its section `Replay` |
| S5 | OK | `{ locale: "es" }` emitted `⚠ "es" is not an active locale of this Help Center (active: en-us, fr)` and then reported all 58 nodes as untranslated, as the warning announces |
| S6 | OK | `{ locale: "FR" }` returned the same empty gap list as S1 and **no** inactive-locale warning |
| S7 | OK | Measured on the tenant, read-only: sideload path **3 requests / 1156 ms**, 58 of 58 nodes classified from the listings; per-node path (two listings plus one request per node, waves of five) **61 requests / 2797 ms**. The request count is the real win, wall-clock less so since the old waves ran five at a time |

**Overall: green, 7/7.**

Two notes for the reviewer:

The tenant has 58 nodes, just under the old 60-node cap, so the *rendered report* is identical before and after this change for `fr` and `en-us`. That is why S7 measures the two strategies directly instead of reading a duration off the tool: on this tenant the cost is the only externally visible difference, and the cap wording could not be exercised for real. The cap's disappearance is covered by unit tests instead (a 61-section tree now reports `## Sections (61 scanned)` with zero per-node calls).

The "listing answered without the sideload" path could not be triggered against Zendesk, which honoured `include=translations` on every path the tool uses, including the two category-scoped ones. It is covered by a unit test, and it is a defensive branch by design: an unknown `include` comes back 200 with the key absent, so a future silent removal would otherwise read as "the whole Help Center is untranslated".

Tenant state after the run: unchanged. The only write (S3) was deleted and both the section's translations and the full `en-us` report were re-read to confirm the original state.

scripts/probe-translations-sideload.ts
dlecan, line 1: Suggestion: internal script, to remove at the end of the development

dlecan:
### Addendum — review pass, and the validated SHA moved

The branch is rebased on `main` (now at `2.17.1`) and carries one more commit, so the validation report above tested `e87363b`, which is now `9fb8523`. Same tree for scenarios S1 to S7; the delta is below.

**Taken.** With part of the tree unclassified and the classified part clean, the bottom line read `No gaps: all N category/ies and 0 section(s) scanned have a published "fr" translation`. Read on its own that is an all-clear over a tree nobody fully audited. It now states that the clean verdict covers only part of the tree and how many nodes went unclassified, with a test for the partial case.

**Declined.** The pass also proposed accepting a top-level `translations` array on the scoped `GET /categories/{id}` read, on the grounds that the embedded shape had only been measured on the *listings*. It was measured on that endpoint too, while preparing this PR:

```
GET /categories/{id}?include=translations  ->  200
category_keys: [id, url, html_url, position, created_at, updated_at,
                name, description, locale, source_locale, outdated, translations]
translations:  [{locale: "en-us", draft: false}, {locale: "fr", draft: false}]
```

So the sideload is embedded there as well, and the fallback would have been a branch for a shape this API does not return, plus a mock asserting invented upstream behaviour. The measurement now sits as a comment at that call site, so the question doesn't get re-opened from the docs alone. S4 exercised the scoped path against the tenant and classified the scoped category, which is the same claim from the other end.

Neither change affects S1 to S7: the partial-unclassified path cannot be triggered against Zendesk (it honoured `include=translations` everywhere), and the scoped read is unchanged from what S4 exercised.

dlecan:
### CodeRabbit pass — five taken, one declined (`51823bb`)

**Taken.**

*Do not describe a first-page audit as a complete tree audit.* Correct, and the sharpest one: rewriting the cost sentence is what introduced the overclaim. The tool reads one page per level (`MAX_PAGE_SIZE`, 100) and flags the rest through the existing `has_more` note, so both the description and `docs/mcp-tools-reference.md` now name that limit instead of implying cursor pagination the tool does not do.

*`/users/me` through `helpCenterGet`.* It resolved to `/api/v2/help_center/users/me` and 404'd. The role it prints is precisely what tells you whether the token could see a draft at all, and the `catch` was swallowing it. `zendeskGet` now; a re-run prints `role: admin`.

*`draftsOf` key-order sensitivity.* Real latent bug: the map is compared through `JSON.stringify` but built in list order, so identical data in a different order would have reported `DRAFT_MISMATCH` and flipped the verdict to NOT VIABLE. It never fired on this tenant, which is luck, not design. Sorted before comparing.

*Verbatim translation dump.* The header promises tenant content stays a preview and that line broke the promise, on output meant for a public issue. Digested, with the field names kept, since the field names are what that dump exists to show.

*Pagination observed, not verified.* Fair. It printed cursors without ever following one. It now fetches the second page when `meta.has_more` says there is one and reports whether the sideload survived it; on a single-page tenant it says outright that the question is unverified rather than letting the cursors imply otherwise. This tenant is single-page, so requirement 4 of the issue stays unproven here, and the output now admits it.

The `.mcp.json` narrowing nitpick is in too, matching `token-persistence.ts`.

**Declined: extract `find_translation_gaps` into a standalone handler.**

All 29 handlers in `src/tools/help-center.ts` are inline arrow functions in the `ToolDefinition[]` array; there is not one standalone handler in the file. The `AGENTS.md` rule reads "Tool handlers are standalone functions in `ToolDefinition[]` arrays" alongside "no classes" — it is about handlers being plain functions rather than methods, which these are. Extracting this one alone would make it the only tool in the file shaped differently, which costs more in consistency than it buys. Worth doing across the file if we want it, not as a drive-by on the one tool this PR happens to touch.

scripts/probe-translations-sideload.ts
dlecan: Done in `9a284c8` — the script is gone and the net diff no longer carries it.

Its payload is in #226 (comment above), which is where the evidence is useful: the sideloaded translation shape, the field-by-field cross-check against the per-node endpoint, and the draft case measured on a real draft. The audit that replaced the fan-out is covered by unit tests and was validated against the tenant.

Worth noting the three older siblings — `probe-audits-shape.ts`, `probe-sla-shape.ts`, `probe-views-shape.ts` — are still on `main` under the same "one-shot instrument" rationale. Happy to remove those too if the intent is that none of them stay; that would be its own issue rather than a drive-by here.

PR #170, opened 2026-07-17
feat(help-center): expose promoted articles as pull-only MCP resources

dlecan:
## Summary
Adds pull-only Help Center **article resources** (`<scheme>://article/{id}`, default `zendesk-hc://article/{id}`) plus a companion **`list_promoted_articles` tool**. Two distinct capabilities:
- **Read-by-id** — any article id can be **read on demand** (Zendesk ACLs enforced via the caller's token), returned as **Markdown**. A cheap single fetch, no preloading; available whenever the `help_center` namespace is active.
- **Promoted pre-listing** — the resource's `list` callback enumerates the promoted ("featured") articles so a client can surface them for pinning / @-mention, and the tool returns the same set. This is the costly, fan-out part (a capped `/articles` scan; no server-side promoted filter), so it is independently toggled with **`--no-promoted-articles`** — which turns off the scan and the tool (zero preloading requests) **without** disabling read-by-id.

This is the reframe agreed on #126: discovery via the enumerable *promoted* set (works on clients that only support static resources, e.g. Claude Desktop), instead of a bare template needing client-side completion. Resource URIs follow the configurable `--hc-resource-scheme` (default `zendesk-hc`) merged from `main` (#173).

## Linked issue
Closes #126

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor (no external behavior change)
- [ ] Documentation
- [ ] Tooling / CI

## Author checklist
- [x] I checked no open or merged PR already addresses the linked issue, and the issue is not already closed as completed / `released`
- [x] If this PR resolves an issue, its description above links it with a closing keyword (`Closes #126`) so it auto-closes on merge to `main`
- [x] `pnpm test` passes locally (590 passed)
- [x] `pnpm test:coverage` meets the thresholds (functions 98.6 ≥ 98, branches 84 ≥ 77, statements 96.7, lines 97.3)
- [x] `pnpm check` is clean (lint + format)
- [x] `pnpm typecheck` passes
- [x] I have read the diff myself, line by line
- [x] I ran a Claude Code review (`/code-review` + `/simplify`) and addressed its findings; CodeRabbit's review findings were also addressed
- [x] Documentation is updated where needed (README, `docs/configuration.md`, `docs/mcp-tools-reference.md`)
- [x] If this PR adds an MCP tool: `list_promoted_articles` is documented in `docs/mcp-tools-reference.md`
[...]

PR #236, opened 2026-08-13
feat(client): retry transient failures and name network errors

### LINKED ISSUE #193: Harden the shared Zendesk HTTP client: retry transient failures + contextual network errors
## Problem

Every tool goes through `executeRequest` in `src/client/zendesk-api.ts`. Two rough edges surfaced during the #170 functional validation (run on a slow Android/Termux link with several requests in flight):

1. **Opaque network errors.** When `fetch` itself throws (connection reset, DNS, timeout — the `await fetch(...)` path, before any HTTP status), the caller (and the LLM / MCP client) sees a bare `fetch failed` with nothing naming which operation / article / page failed. HTTP *responses* are wrapped nicely as `ZendeskApiError` (401/403/404/422/429/…), but pre-response network failures are not wrapped at all.
2. **No retry.** A single transient failure fails the whole tool call. During validation, two calls failed on the first attempt and succeeded verbatim on retry — there is no backoff/retry for network blips, `429`, or `5xx`.

## Proposed solution

Harden `executeRequest` in one place (benefits every tool, read and write):

- **Wrap network failures** (the `fetch` throw path) in a typed error carrying context: the HTTP method + request path/host — **not** the token, and with query params dropped or redacted. ASCII-only message (same rule as the auth headers: non-ASCII bytes break `node:http`), e.g. `Network error on GET /api/v2/help_center/articles/123: <cause>`. Keep the existing `ZendeskApiError` messages for terminal HTTP errors.
- **Bounded retry with backoff + jitter** for transient conditions:
  - retry on network errors, `5xx`, and `429`;
  - honor `Retry-After` on `429`;
  - small attempt budget (e.g. 3) with exponential backoff + jitter;
  - **idempotency is the delicate part**: retry `GET`/`DELETE` (idempotent) freely; for non-idempotent writes (`POST` create — tickets, articles, comments) either don't retry, or retry only on a *pre-send* connection failure where the request provably never reached Zendesk, to avoid duplicate creations. Make the policy explicit and tested.

## Alternatives considered

- **Per-tool retry** — rejected: duplicative and inconsistent; the shared client is the right layer.
- **A retry library** — likely overkill; a small inline helper keeps the dependency surface minimal (per the repo's dependency stance).

## Scope

- [ ] Read-only (no data is modified in Zendesk)

Neither, exactly: this is **cross-cutting client infrastructure** that every call flows through, read and write. The write path is what needs care (retry must not duplicate creates).

## Additional context

- Touch-points: `src/client/zendesk-api.ts` — `executeRequest` (the `await fetch(...)` + `!response.ok` block) and `ZendeskApiError`.
- Surfaced by the #170 functional validation; the validator's "environment note" flagged the opaque `fetch failed` as hard to act on from a client seat.
- Tests (MSW): transient-then-success retry, no-retry on terminal 4xx, `Retry-After` honored, write-idempotency guard (no duplicate POST), and the wrapped network-error message shape (context present, token/query absent).
- Follow-up split out of #170 (do not block that PR on this).

dlecan:
## Summary

Every tool call reaches Zendesk through the shared client, and two rough edges surfaced during the #170 functional validation on a slow link: a `fetch` throw reached the LLM as a bare `fetch failed` naming nothing, and a single transient blip failed the whole tool call. This adds one method-keyed retry policy (`src/client/retry.ts`) and routes **all four** fetch call sites through a single `performFetch`, so transient failures recover silently while writes stay duplicate-safe.

## Linked issue

Closes #193

## Type of change
- [x] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tooling / CI

## What changed

`src/client/retry.ts` (new) holds the policy and nothing else:

| Method | Network failure | `5xx` | `429` |
| --- | --- | --- | --- |
| `GET` | retried (any phase) | retried | retried |
| `POST`, `PUT`, `DELETE` | retried only pre-send | **never** | retried |

- **Network failures are named.** `Network error on GET https://acme.zendesk.com/api/v2/help_center/articles/123 after 3 attempts: ECONNRESET: fetch failed` — method + origin + path, ASCII-only, original error kept as `cause`, and credential-free: the query string is dropped (`zendeskUpload` puts an upload token there) and an attachment URL's `/token/<token>/` path segment is redacted.  An Error factory + `isZendeskNetworkError` guard, matching `createAuthRequiredError` rather than adding a second error class.
- **Bounded retry:** up to 3 attempts (two retries), 250 ms base, exponential with equal jitter, capped at 4 s. `Retry-After` wins over backoff wherever it appears (a `429`, or a `503` during maintenance) up to 5 s; past that the response is surfaced instead of parking the call. A response that will be replayed is drained first so the socket is released — and a body that dies mid-drain doesn't abort the retry.
- **Per-attempt deadline:** `AbortSignal.timeout` — 30 s for JSON calls, 120 s for attachment downloads and uploads. Node's `fetch` has no deadline of its own (undici's `headersTimeout`/`bodyTimeout` default to **300 s**), so a stalled socket could otherwise hold a tool call open long past the retry budget, which cannot fire while an attempt is pending. The signal is built inside the retry thunk, so each attempt starts fresh.
- **All four call sites** now share `performFetch`: `executeRequest`, `fetchZendeskBinary`, `zendeskUpload`, `helpCenterUpload`.
- `ZendeskApiError` and every per-status message are untouched, so the 401 backstop in `server.ts` and the `403`/`404` branches in the tools behave exactly as before.

Only a `GET` is replayed freely. Everything else replays exclusively what provably did not apply:

- **`PUT` counts as non-idempotent** despite HTTP semantics: `PUT /tickets/{id}` with a `comment` *appends* one (`add_private_note`, `add_public_comment`), so a blind replay would duplicate it. The client sits below the tool layer and cannot see `idempotentHint`, so the method is all it has.
- **`DELETE` is not "retried freely"** as the issue suggested, on either a network failure or a `5xx`: replaying a delete that already applied reports a misleading `404` for work that succeeded.
- **`429` is retried on writes** — Zendesk refused the request, so it provably did not apply. A `5xx` may have applied and is therefore terminal.
- **A deadline abort needs no new rule.** It arrives as a `TimeoutError` whose `code` is numeric rather than a syscall string, so it classifies as `unknown`: a GET is retried, a write never is, since the request may have arrived before the deadline passed.

No new env var or CLI flag: the budget is hardcoded, and timing is injected at the function level (`RetryDeps`) so tests are deterministic.

## Author checklist
- [x] `pnpm test` passes locally (871 tests)
- [x] `pnpm typecheck` passes
- [x] Documentation is updated where needed (`docs/troubleshooting.md`; one durable rule in `AGENTS.md`)
- [x] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md` — n/a, no tool-surface change
- [x] If the change has **MCP-runtime-observable** behaviour: this PR carries a `## Functional validation plan` below, executable end to end by one command

`pnpm test:mutation:diff origin/main HEAD`: **175 mutants in the changed lines, 175 killed, 0 escaped.**

### Review rounds

`/code-review` found four issues, fixed in `9591752`: a 5xx DELETE was being replayed (contradicting the policy's own rationale), the body drain could throw out of the retry loop, an attachment download token leaked through the path, and two inaccurate sentences in the troubleshooting doc.

CodeRabbit then found four more, all valid:

| Finding | Outcome |
| --- | --- |
| No per-attempt deadline (Major) | Fixed in `4d556d1` — was an explicit non-goal, included after the maintainer chose to |
| `Retry-After` ignored on a retriable `5xx` | Fixed in `4e31297` |
| No client-level test for a `429` replaying a write | Added in `4e31297` (`Retry-After: 0`, so the suite doesn't wait) |
| `AGENTS.md` said "every request" | Scoped to Zendesk API requests in `4e31297`; the OAuth token exchange calls `fetch` directly |

The upload call sites also turned out to have no happy-path test at all — the mutation gate caught it once those lines were touched. `afee290` pins what both uploads put on the wire (auth header, content type, body, aggregating token).

## Functional validation plan

Two parts, because they answer different questions. Part B is the one that used to
be impossible: the retry, timeout and write-safety behaviour only shows itself when
Zendesk misbehaves, which a healthy tenant never does on request. It is now one
command.

### Part A — the live tenant (the tools still work)

Only a real tenant proves the four fetch paths still function and that a write
produces exactly one artifact. Use a throwaway ticket and a draft article.

| id | Call | Expected |
| --- | --- | --- |
| **A1** | `list_articles`, then `get_article` on a returned id | both succeed, sub-second |
| **A2** | `add_private_note` with a recognizable body, then re-read the ticket | the note appears **exactly once** — report the ticket id and the count |
| **A3** | `get_ticket_attachments` on a ticket holding an image | image content back, not a `401`/`403` (which would mean the auth header was lost) |
| **A4** | `add_private_note` with a small `attachments` entry, and `create_article_attachment` on the draft | both succeed, attached once |
| **A5** | `get_article` with id `999999999` | still `Resource not found…`, returning **immediately** — a ~1 s pause would mean 4xx is being retried |
| **A6** | the largest attachment you can find (> 1 MB) | completes; the 120 s transfer deadline must not cut real work short |

### Part B — the failure paths, deterministically

```bash
pnpm build && node scripts/validate-retry.mjs      # ~40 s
```

Each scenario boots the real server with `scripts/fault-inject.mjs` preloaded and
drives a real MCP tool call over stdio. It judges on what the caller received **and
on how many requests Zendesk actually saw** — that count is what proves a write was
not replayed, where a message could be misread. Non-zero exit if any scenario
fails; `--skip-slow` drops the 30 s deadline case.

| id | Proves |
| --- | --- |
| **S1** | a healthy read goes through untouched (1 request) |
| **S2** | a transient failure is absorbed, the caller never sees it (2 requests) |
| **S3** | a read spends its whole budget on a `5xx` (3 requests) |
| **S4** | a `5xx` write is **never replayed** (1 request) and says it may have applied |
| **S5** | a throttled write **is** replayed, exactly once (2 requests, succeeds) — the only path where the client deliberately re-sends a mutation |
| **S6** | a `Retry-After` beyond the 5 s cap is surfaced instead of parking the call (1 request) |
| **S7** | a write that failed with no response is not replayed, and the message carries no credential |
| **S8** | a read retries a response-less failure and then names it (3 requests) |
| **S9** | a transfer slower than the 30 s JSON deadline still completes on the binary path — the two budgets are genuinely distinct |
| **S10** | a stalled socket is cut by the per-attempt deadline (~30 s), not left hanging |
| **S11** | a `404` is terminal: one request, wording untouched |

### What this still cannot prove

`HttpResponse.error()` raises a failure with **no syscall code**, so the simulator
cannot tell a pre-send `ENOTFOUND` from an in-flight `ECONNRESET`. The "never
reached Zendesk, retrying is safe" branch is therefore covered by unit tests only;
reaching it live would need a fault-injecting proxy, which in turn needs a base-URL
override the client deliberately does not have.

### Reporting

Post one PR comment in English: the SHA tested, the pasted output of Part B, and
per-id verdicts for Part A with the artifact counts for A2 and A4.

- **Why the mechanics are hand-rolled rather than a library'''s** is now recorded in [`docs/decisions/client-retry.md`](docs/decisions/client-retry.md), with the spike behind it: undici'''s retry interceptor would remove the most (loop, backoff, `Retry-After`, sockets) but MSW intercepts **above** the dispatcher, so it never runs in the suite; the fetch-level alternative is testable and expresses the policy exactly, but removes only ~40 lines and last shipped in March 2024. The doc also states plainly that "a library would replay a `PUT` by default" is *not* a valid argument for owning the loop — that choice is configuration.
- **The idempotency table is the design decision to check.** Specifically: is treating `PUT` as non-idempotent the right call (it is what makes `add_public_comment` safe), and is retrying writes on `429` acceptable?
- **The two deadlines are a judgement call**: 30 s JSON / 120 s transfers. Deliberately generous, since aborting a legitimate slow upload is worse than waiting.
- The cause-chain walk is depth-bounded (5 levels) rather than cycle-detecting, and stops explicitly on `null` (`typeof null === 'object'` would otherwise crash while building an error message). Both bounds are pinned by tests.
- TLS/certificate failures are classified `unknown`, so a GET retries them 3 times (~0.5 s) before failing. Deliberate: keeping the pre-send list short and explicit beats enumerating every deterministic-but-unretriable code.

## Summary by CodeRabbit

* **New Features**
  * Added resilient handling for network failures, server errors, and rate limits across Zendesk requests.
  * Added per-attempt timeouts and safer retry behavior that avoids replaying potentially completed updates.
  * Improved error messages with sanitized request details.
  * Extended reliability safeguards to downloads and attachment uploads.

* **Documentation**
  * Added troubleshooting guidance for network failures and retry behavior.
  * Documented retry policies, design considerations, and live validation scenarios.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,6 +45,11 @@ Zod schema. Each proxy dispatches only within its own scoped handler map, so a
 `--read-only` / `--tool` are applied by `filterTools` *before* the mode switch;
 `--tool` also forces `mode: all` (`config.ts`).
 
+Every request goes through `performFetch` in `client/zendesk-api.ts`, so retry and
+network-error wrapping are decided once, in `client/retry.ts`: a new client method must
+pick its row in that policy table, where a write is never replayed once it may have
+reached Zendesk.
+
 **Auth** — per-user OAuth 2.1 PKCE only, no static API-token mode. stdio: lazy
 browser PKCE via `token-store.ts`. HTTP: per-session bearer captured from
 `Authorization:`. Dropping API-token auth is deliberate (static shared
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -49,7 +49,8 @@ Every Zendesk API request goes through `performFetch` in `client/zendesk-api.ts`
 (the OAuth token exchange in `auth/browser-oauth.ts` deliberately does not), so retry and
 network-error wrapping are decided once, in `client/retry.ts`: a new client method must
 pick its row in that policy table, where a write is never replayed once it may have
-reached Zendesk.
+reached Zendesk. Why the loop is ours rather than a library's — and what would change
+that: `docs/decisions/client-retry.md`.
 
 **Auth** — per-user OAuth 2.1 PKCE only, no static API-token mode. stdio: lazy
 browser PKCE via `token-store.ts`. HTTP: per-session bearer captured from
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,7 +45,8 @@ Zod schema. Each proxy dispatches only within its own scoped handler map, so a
 `--read-only` / `--tool` are applied by `filterTools` *before* the mode switch;
 `--tool` also forces `mode: all` (`config.ts`).
 
-Every request goes through `performFetch` in `client/zendesk-api.ts`, so retry and
+Every Zendesk API request goes through `performFetch` in `client/zendesk-api.ts`
+(the OAuth token exchange in `auth/browser-oauth.ts` deliberately does not), so retry and
 network-error wrapping are decided once, in `client/retry.ts`: a new client method must
 pick its row in that policy table, where a write is never replayed once it may have
 reached Zendesk.
--- a/docs/decisions/client-retry.md
+++ b/docs/decisions/client-retry.md
@@ -1,0 +1,127 @@
+# Client retry: why the policy is ours and the mechanics are hand-rolled
+
+> **Build documentation, not user documentation.** This records why
+> `src/client/retry.ts` implements its own retry loop instead of delegating to a
+> library. Client-facing behaviour is described in
+> [`docs/troubleshooting.md`](../troubleshooting.md).
+
+| | |
+| --- | --- |
+| **Status** | Decided and applied |
+| **Date** | 2026-08-14 |
+| **Applied in** | [#236](https://github.com/fruggr/zendesk-mcp-server/pull/236) ([#193](https://github.com/fruggr/zendesk-mcp-server/issues/193)) |
+| **Question** | Should the retry layer use an off-the-shelf library rather than its own loop, backoff and `Retry-After` handling? |
+| **Answer** | **Not yet** — for one measured reason, not on principle: the library that removes the most is invisible to the test suite, and the one that is testable removes ~40 lines while adding a dependency last published in March 2024. |
+
+## The argument that does *not* justify hand-rolling
+
+Zendesk's `PUT /tickets/{id}` **appends** a comment, so replaying it emails a
+customer twice. That is real, and it is why `POST`/`PUT`/`DELETE` only replay a
+`429` or a connection that never opened.
+
+It is tempting to conclude "a library would get this wrong by default, so we
+write our own". That conclusion is wrong, and this section exists so nobody
+reaches for it again. Every candidate below exposes that choice as
+**configuration** — `methods`, a predicate, a callback. Knowing that a `PUT` is
+not idempotent here is a *requirement*: it has to be stated either way, in a
+config value or in a policy table. Owning the requirement never implies owning
+the loop, the jitter or the header parsing. **Maintain the need, not the how.**
+
+The decision below rests only on what was measured.
+
+## What was measured
+
+Spike against `undici` 8.10.0 and the repo's `msw` 2.15.0.
+
+| | undici `RetryAgent` / `interceptors.retry` | `fetch-retry` 6.0.0 | `p-retry` |
+| --- | --- | --- | --- |
+| Policy expressible as config | yes (`methods`, `statusCodes`, `errorCodes`, `retry` callback) | yes (`retryOn`, `retryDelay`) | partly — HTTP-agnostic, predicate is ours |
+| Works with MSW | **no** | yes | yes |
+| Parses `Retry-After` for us | yes | no | no |
+| Last publish | current | **2024-03-17** | 2026-03-26 |
+
+### 1. undici's interceptor is invisible to MSW
+
+The strongest candidate — it owns the loop, the backoff, `Retry-After` *and* the
+socket handling. It is unusable here because MSW intercepts **above** the
+dispatcher layer:
+
+| Wiring | MSW intercepts | Retry runs |
+| --- | --- | --- |
+| global `fetch` + per-call `dispatcher` | yes | **never** |
+| `setGlobalDispatcher` + global `fetch` | yes | **never** |
+| `undici.fetch` + `dispatcher` | **no** | yes — against the real API |
+
+The third row is not theoretical: the probe received a genuine
+`No help desk at testsubdomain.zendesk.com` from Zendesk. So the choice would be
+an untested write-safety policy, or real API calls in the suite — and
+[`AGENTS.md`](../../AGENTS.md) requires MSW.
+
+Its defaults are also unsafe for this API (`methods` includes `PUT` and `DELETE`,
+`errorCodes` includes the post-send `ECONNRESET`), but per the section above that
+is configuration, not an argument.
+
+### 2. `fetch-retry` is testable, and expresses the policy exactly
+
+A fetch-level wrapper stays above MSW. All three cases behave as specified, with
+the policy living entirely in `retryOn` / `retryDelay`:
+
+| Call | Result |
+| --- | --- |
+| `GET` + `500` | retried, 3 requests, ends `200` |
+| `POST` + `429` (`Retry-After: 0`) | replayed once, 2 requests, ends `200` |
+| `PUT` + `500` | 1 request, ends `500` — never replayed |
+
+Two findings decide against it for now:
+
+- **It removes less than it looks.** Gone: the loop and the backoff maths, ~40
+  lines. Still ours: `parseRetryAfter` (a `retryDelay` callback must return a
+  number, so someone reads the header — including the trap that
+  `Date.parse('-1')` yields the year 2001), the pre-send/unknown classification,
+  the error wrapping and credential redaction, the per-attempt deadline, and
+  draining a discarded response body to release the socket.
+- **A fetch wrapper replays the same `init`, so it replays the same
+  `AbortSignal`.** With one `AbortSignal.timeout` in `init`, a first attempt that
+  hits the deadline ends the call: the probe stayed at 1 request and surfaced
+  `TimeoutError` with no retry. Fixable — inject a fresh signal in the fetch
+  function handed to the factory, which *is* called per attempt — but it is a
+  trap, and `performFetch` already avoids it by building the signal inside the
+  retry thunk.
+
+Weigh that against a dependency with no release in over two years, on the path
+that keeps a customer-facing comment from being sent twice.
+
+## What would reverse this
+
+- **`fetch-retry` resumes releases.** The leverage stays small, but the principle
+  holds; swap then.
+- **The client layer stops being tested through MSW** — a local `node:http`
+  server for `tests/unit/client/` would let a custom dispatcher run, unlocking
+  undici's interceptor, which is the real prize since it owns `Retry-After`
+  parsing too. That is a test-architecture decision in its own right, not a
+  drive-by swap.
+
+Until then the mechanics stay in `src/client/retry.ts`, where the diff gate holds
+them at a 100 % mutation score.
+
+## Reproducing
+
+```js
+// MSW + a custom undici dispatcher: intercepted, but the retry never runs.
+import { setupServer } from 'msw/node';
+import { http, HttpResponse } from 'msw';
+import { Agent, interceptors } from 'undici';
+
+let hits = 0;
+const server = setupServer(
+  http.get('https://testsubdomain.zendesk.com/api/v2/flaky', () => {
+    hits += 1;
+    return hits === 1 ? HttpResponse.json({}, { status: 500 }) : HttpResponse.json({ ok: true });
+  }),
+);
+server.listen();
+
+const agent = new Agent().compose(interceptors.retry({ maxRetries: 2, methods: ['GET'] }));
+const res = await fetch('https://testsubdomain.zendesk.com/api/v2/flaky', { dispatcher: agent });
+console.log(res.status, hits); // 500 1 — one attempt, no retry
+```
--- a/docs/live-testing.md
+++ b/docs/live-testing.md
@@ -91,8 +91,8 @@ runner supplies the misbehaviour:
 
 ```bash
 pnpm build
-node scripts/validate-retry.mjs             # 8 scenarios, ~40 s
-node scripts/validate-retry.mjs --skip-slow # drops the 30 s deadline scenario
+node scripts/validate-retry.mjs             # 11 scenarios, ~80 s
+node scripts/validate-retry.mjs --skip-slow # drops the two that wait out a deadline
 ```
 
 Each scenario boots the real server with `scripts/fault-inject.mjs` preloaded —
@@ -102,8 +102,8 @@ on how many requests Zendesk actually saw**: that count is what proves a write w
 not replayed, where a message could be misread. The run exits non-zero if any
 scenario fails, so a green result is a fact rather than an impression.
 
-To exercise a mode by hand, set `FAULT` (`none`, `flaky-then-ok`, `500`, `429`,
-`network`, `stall`) and preload the same file:
+To exercise a mode by hand, set `FAULT` (`none`, `flaky-then-ok`, `500`, `404`,
+`429`, `429-brief`, `network`, `stall`, `slow-transfer`) and preload the same file:
 
 ```bash
 FAULT=500 SUB=validation node --import ./scripts/fault-inject.mjs \
--- a/docs/live-testing.md
+++ b/docs/live-testing.md
@@ -82,3 +82,37 @@ It links a real MCP `Client` to the server over an in-memory transport
 `tools/call`, `resources/list` and `resources/read` cross the wire and dispatch
 through the genuine handlers. Use it for quick manual checks, or as a basis for
 scripted or CI assertions.
+
+## C. `scripts/validate-retry.mjs`: failure paths, without a flaky network
+
+The retry, timeout and write-safety behaviour of the client only shows itself
+when Zendesk misbehaves, which a healthy tenant never does on request. This
+runner supplies the misbehaviour:
+
+```bash
+pnpm build
+node scripts/validate-retry.mjs             # 8 scenarios, ~40 s
+node scripts/validate-retry.mjs --skip-slow # drops the 30 s deadline scenario
+```
+
+Each scenario boots the real server with `scripts/fault-inject.mjs` preloaded —
+MSW patching the same global `fetch` the Zendesk client uses — then drives a real
+MCP tool call over stdio. It judges the outcome on what the caller received **and
+on how many requests Zendesk actually saw**: that count is what proves a write was
+not replayed, where a message could be misread. The run exits non-zero if any
+scenario fails, so a green result is a fact rather than an impression.
+
+To exercise a mode by hand, set `FAULT` (`none`, `flaky-then-ok`, `500`, `429`,
+`network`, `stall`) and preload the same file:
+
+```bash
+FAULT=500 SUB=validation node --import ./scripts/fault-inject.mjs \
+  dist/index.js validation --mode all
+```
+
+Two things it cannot do. `HttpResponse.error()` raises a failure with **no syscall
+code**, so it cannot tell a pre-send `ENOTFOUND` from an in-flight `ECONNRESET` —
+the "never reached Zendesk, retrying is safe" branch is covered by unit tests
+only, and a real fault-injecting proxy would be needed to go further. And nothing
+here reaches production: `msw` is a devDependency, and no file under `src/`
+imports it.
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -101,22 +101,23 @@ reports; omitting `user_segment_id` on create/update keeps the default visibilit
 
 The request never got an HTTP response — DNS, a refused or reset connection, a
 proxy dropping the link. The message names the method and the path so you know
-what failed; the query string is dropped and no token appears in it.
+what failed, with credentials stripped: no bearer token, no query string, and an
+attachment URL's download token redacted from the path.
 
-Transient failures are already retried up to 3 times with backoff before you see
-this, so a message here means every attempt failed. What gets retried depends on
-the method, because a replay must never duplicate a write:
+The request is tried up to 3 times (so up to two retries) with backoff before you
+see this, meaning every attempt failed. What gets retried depends on the method,
+because a replay must never duplicate a write:
 
 | | Network failure | `5xx` | `429` |
 | --- | --- | --- | --- |
 | `GET` | retried | retried | retried |
-| `DELETE` | retried only if the connection never opened | retried | retried |
-| `POST`, `PUT` | retried only if the connection never opened | **not retried** | retried |
+| `POST`, `PUT`, `DELETE` | retried only if the connection never opened | **not retried** | retried |
 
-So a create or a comment is never sent twice: a `429` means Zendesk refused the
-request, but a `5xx` may have applied it, and it is surfaced rather than replayed.
-A `429` carrying a `Retry-After` above 5 s is surfaced too instead of parking the
-call — wait and ask again.
+So a create, a comment or a delete is never sent twice: a `429` means Zendesk
+refused the request, but a `5xx` may have applied it, so it is surfaced rather
+than replayed — including on a delete, where a replay would report a misleading
+`404` for work that succeeded. A `429` carrying a `Retry-After` above 5 s is
+surfaced too instead of parking the call — wait and ask again.
 
 Where each client writes the server's stderr:
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -97,6 +97,27 @@ those rights, read an existing article with `get_article` and reuse the IDs it
 reports; omitting `user_segment_id` on create/update keeps the default visibility
 (everyone).
 
+## A tool call fails with `Network error on GET https://…`
+
+The request never got an HTTP response — DNS, a refused or reset connection, a
+proxy dropping the link. The message names the method and the path so you know
+what failed; the query string is dropped and no token appears in it.
+
+Transient failures are already retried up to 3 times with backoff before you see
+this, so a message here means every attempt failed. What gets retried depends on
+the method, because a replay must never duplicate a write:
+
+| | Network failure | `5xx` | `429` |
+| --- | --- | --- | --- |
+| `GET` | retried | retried | retried |
+| `DELETE` | retried only if the connection never opened | retried | retried |
+| `POST`, `PUT` | retried only if the connection never opened | **not retried** | retried |
+
+So a create or a comment is never sent twice: a `429` means Zendesk refused the
+request, but a `5xx` may have applied it, and it is surfaced rather than replayed.
+A `429` carrying a `Retry-After` above 5 s is surfaced too instead of parking the
+call — wait and ask again.
+
 Where each client writes the server's stderr:
 
 | Client | Log location |
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -132,6 +132,17 @@ refused the request, but a `5xx` may have applied it, so it is surfaced rather
 than replayed — including on a delete, where a replay would report a misleading
 `404` for work that succeeded.
 
+A failed **write** says which of the two it was, so that retrying it by hand (or
+having an assistant retry it) does not create the duplicate the client just
+avoided:
+
+| The message ends with | What it means |
+| --- | --- |
+| *The write may already have been applied…* | check the ticket or article before retrying |
+| *…nothing was applied. Retrying is safe.* | the request was refused or never left; send it again |
+
+A read never carries either note: replaying a `GET` cannot duplicate anything.
+
 When a response carries `Retry-After` (Zendesk sends it on a `429`, and on a `503`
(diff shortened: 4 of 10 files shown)


dlecan:
Both findings on the validation runner are fixed in `257fc75` — noting it here because the review quota was reached before they could be re-checked, so the risk block above is still pinned to `c7db018`.

**The `429` scenario was right for the wrong reason, which was the more useful half of the finding.** It injected `Retry-After: 600`, past the client's 5 s cap, so the response was surfaced without a replay and the scenario passed with one request. Accurate for that fixture — but it proved the opposite of the rule it was named after, and a regression disabling the `429` write replay would have gone green. Split in two:

- **S5** injects a brief `Retry-After` and asserts the write is **replayed exactly once** (2 requests, succeeds). That is the only path where the client deliberately re-sends a mutation, so it is the one that most needed pinning.
- **S6** keeps the beyond-the-cap case and asserts the call is surfaced rather than parked.

The watchdog is cleared on completion too. Harmless today since the runner exits explicitly, but eight armed timers would have hung anyone who removed that exit.

All 9 scenarios pass, including the deadline firing at ~30 s on a stalled socket.

---

dlecan:
## Functional validation report — 257fc75e9985e15b4b1e17f928e4639cb309b025

Tenant values are redacted; only ids, sizes and message structure are reported.

### Part A — live tenant

| ID | Verdict | Evidence |
| -- | ------- | -------- |
| A1 | OK | `list_articles` (page_size 3) returned 3 rows + a cursor; `get_article` on a returned id returned the large-article notice, metadata block and body. No perceptible delay on either. |
| A2 | OK | Throwaway ticket **#7451** created, then `add_private_note` with marker `VALIDATION-MARKER-A2-<sha>`. Re-read with `include_comments: true`: **exactly one** internal note carrying that marker (thread = 1 public comment + 1 note). |
| A3 | OK | `get_ticket_attachments` on ticket **7427** returned the image *content* inline (`image/png`, 56495 bytes) plus its text reference — no 401/403, so the auth header survives `fetchZendeskBinary`. |
| A4 | OK | `add_private_note` with one attachment → "Private note added to ticket #7451 with 1 attachment(s)"; re-read shows **one** new note with **one** attachment id. `create_article_attachment` on draft article **38390635947805** → `list_article_attachments` returns **one** entry (`image/png`, 70 bytes). No duplicates on either path. |
| A5 | OK | `get_article` id `999999999` → `Resource not found. Please verify the ID is correct. (Not Found)`, unchanged wording, returned with no perceptible pause. Caveat: I had no instrumentation on MCP call latency, so "immediate vs ~1 s" is an observation, not a measurement — the no-retry-on-4xx rule itself is pinned by the unit suite. |
| A6 | BLOCKED | No attachment > 1 MB reachable through the tool. `get_ticket_attachments` only downloads bytes for `image/*` under the 5 MB cap; every non-image (xlsx, pdf, zip) is listed as a reference and never fetched. Sampled 8 attachment-bearing tickets (7426, 7427, 7429, 7439, 7446, 7448, plus the throwaway and empty ones); largest image found and **successfully downloaded** was 793281 bytes on ticket 7448. The 120 s transfer deadline is therefore untested against a genuinely large transfer. |

### Part B — failure paths

`pnpm build && node scripts/validate-retry.mjs`, exit code 0, all 9 scenarios passed. Request counts match the plan exactly (S1=1, S2=2, S3=3, S4=1, S5=2, S6=1, S7=1, S8=3, S9=1).

```
Retry and timeout validation — 9 scenarios

OK   S1  a healthy read goes through untouched
       fault=none tool=get_current_user requests=1 elapsed=3283ms
       ## Validation User (9999) - **Email**: validation@example.com - **Role**: admin - **Active**: undefined

OK   S2  a transient failure is absorbed, the caller never sees it
       fault=flaky-then-ok tool=get_current_user requests=2 elapsed=3109ms
       ## Validation User (9999) - **Email**: validation@example.com - **Role**: admin - **Active**: undefined

OK   S3  a read spends its whole budget on a 5xx
       fault=500 tool=get_current_user requests=3 elapsed=3242ms
       Zendesk API error 500: Internal Server Error. {"error":"injected"}

OK   S4  a 5xx write is never replayed, and says it may have applied
       fault=500 tool=add_private_note requests=1 elapsed=2623ms
       Zendesk API error 500: Internal Server Error. {"error":"injected"} The write may already have been applied. Check the current state before retrying, or you may duplicate it.

OK   S5  a throttled write IS replayed, exactly once — Zendesk refused it, so it never applied
       fault=429-brief tool=add_private_note requests=2 elapsed=2641ms
       Private note added to ticket #1.

OK   S6  a Retry-After beyond the cap is surfaced instead of parking the call
       fault=429 tool=add_private_note requests=1 elapsed=2531ms
       Rate limit exceeded. Please wait before making more requests. Zendesk refused the request, so nothing was applied. Retrying is safe.

OK   S7  a write that failed with no response is not replayed either
       fault=network tool=add_private_note requests=1 elapsed=2527ms
       Network error on PUT https://validation.zendesk.com/api/v2/tickets/1 after 1 attempt: Failed to fetch. The write may already have been applied. Check the current state before retrying, or you may duplicate it.

OK   S8  a read retries a failure with no response, then names it
       fault=network tool=get_current_user requests=3 elapsed=3226ms
       Network error on GET https://validation.zendesk.com/api/v2/users/me after 3 attempts: Failed to fetch

OK   S9  a stalled socket is cut by the per-attempt deadline, not left hanging
       fault=stall tool=add_private_note requests=1 elapsed=32710ms
       Network error on PUT https://validation.zendesk.com/api/v2/tickets/1 after 1 attempt: TimeoutError: The operation was aborted due to timeout. The write may already have been applied. Check the current state before retryi

All 9 scenarios passed.
```

S7 and S8 also confirm the credential-free message: the URL carries no query string, and the method + origin + path + attempt count are all present.

### Side observations (out of the plan's scope)

- The PNG uploaded through `add_private_note` (`content_type: image/png`) comes back from Zendesk as `application/unknown`, so `get_ticket_attachments` lists it as a text reference instead of embedding it — that is why A3 was run against a pre-existing ticket. Not a regression from this PR: `git diff origin/main HEAD` shows the `Content-Type: contentType` header on the upload only moved by indentation.
- Artifacts left on the tenant for cleanup: ticket **#7451** (tagged `validation_pr236`, `throwaway`) and draft article **38390635947805**.

## Summary

Part B is green end to end: 9/9, exit 0, with the request counts that make the write-safety claim checkable rather than message-deep. Part A is green on A1–A5, including the two things worth doubting — one note per write, one attachment per upload.

A6 is the only gap, and it is an environment gap rather than a defect: the tool layer only downloads `image/*` under 5 MB, and this tenant has no image between 1 MB and that cap among the tickets I sampled. The largest real transfer exercised was ~0.8 MB, which completed. If you want the 120 s deadline covered against real work, it needs either a seeded large image or a scenario in `validate-retry.mjs`.

dlecan:
Thanks — verified against the plan rather than taken on trust, and the two caveats you raised are now closed in `1e96a46`.

**Report checks out.** Right SHA (`257fc75`, the head at the time). Part B's counts match the plan exactly, including the two that carry the weight: S4 at one request (a `5xx` write never replayed) and S5 at two (the one path where a mutation is deliberately re-sent). A2 and A4 are the ones I would have doubted — one note per write, one attachment per upload — and both are evidenced with ids rather than assertions.

**A6 was the real finding**, and your diagnosis was right: it is an environment gap, not a defect. `get_ticket_attachments` only downloads bytes for `image/*` under the 5 MB cap, so a tenant with no image in the 1–5 MB band cannot exercise the transfer deadline at all. Rather than seed one, the runner now covers it deterministically:

- **S9** mocks the comments-plus-download path and answers the transfer after **35 s** — longer than the 30 s JSON deadline, well inside the 120 s transfer one. It completes (35 770 ms, 2 requests, attachment returned). That is the first evidence that the binary path genuinely *gets* the longer budget, rather than merely being configured with a larger constant. Until now only a unit test asserted `TRANSFER > REQUEST` as numbers.

**A5 deserved its caveat too** — "returns immediately" was a judgement without instrumentation. **S11** injects a `404` and asserts one request with the wording untouched, so the no-retry-on-4xx rule is a counter now, not an impression.

11/11 pass. One numbering shift to note: the stall scenario your report lists as S9 is now **S10**.

Two things from your side observations that I have deliberately *not* acted on:

- The `image/png` upload coming back as `application/unknown` is a genuine product wart — it means an image attached through the tool is never re-embedded on read. You already established it is not a regression here. It deserves its own issue; say the word and I will open it with your evidence.
- Ticket **#7451** and draft article **38390635947805** are still on the tenant. Deleting or archiving them is outward-facing on your data, so I would rather you confirm than assume.

---

dlecan:
Cleanup of the validation artifacts, following up on my report above:

- Draft article 38390635947805: archived via `archive_article` (soft-delete, the only removal the API offers).
- Ticket #7451: set to `closed`. The server exposes no ticket deletion, so if you want it gone for good it has to be done from the Zendesk admin UI.

Nothing else was written to the tenant.

dlecan:
Fixed in 83e6401 — the last open finding on the runner.

Reproduced first: pointing `PRELOAD` at a nonexistent file and running under `timeout 100` printed only `Terminated`, so the child was indeed dying at startup and every scenario was sitting out its full 60 s watchdog.

The fix settles on `close` rather than `exit`, so a response already in the pipe still wins the race on a server that answers and then dies; `error` covers a spawn that never starts, and `stdin` errors are swallowed since the sends are written synchronously right after `spawn`. The report now names the cause instead of the symptom — it surfaces the first stderr line matching `Error`:

```
FAIL S1  a healthy read goes through untouched
       fault=none tool=get_current_user requests=0 elapsed=52ms
       <server exited (code=1, signal=null) before responding> Error [ERR_MODULE_NOT_FOUND]: Cannot find module '…/scripts/does-not-exist.mjs'
```

~50 ms instead of 60 s, and ~0.5 s for the whole matrix instead of 9 minutes.

The 60 s watchdog stays: it still covers the case it was written for — a child that stays alive without answering, which is exactly what S10 (`stall`) relies on.

Full matrix re-run on the real preload: **11/11 pass**, with S9 at 35.9 s and S10 cut at 30.8 s as expected. `pnpm check` / `pnpm typecheck` / `pnpm test` (895) green.

---

dlecan:
## Validation addendum — three live-only items

Extends the `## Functional validation plan` in the description. Nothing in the code changed: head is still `83e6401`, CI green. These are the checks that **only a live tenant or a real network** can answer, and they exist because A6 came back `BLOCKED` and a side observation turned out to matter more than a side observation should.

### Context

Two upload paths reach Zendesk, and the last report suggests they disagree:

| Path | Tool | Transport | Reported back as |
| --- | --- | --- | --- |
| Uploads API | `add_private_note` (`attachments`) | raw body + `Content-Type` header | `application/unknown` |
| Help Center | `create_article_attachment` | `multipart/form-data` (`Blob` type) | `image/png` |

That difference decides where a fix goes, and right now it is unattributed — the report states `image/png` was passed on the ticket call, but **both** tools declare `content_type: z.string().default('application/octet-stream')`, so an omitted field is indistinguishable from an explicit one after the fact.

Ruled out already, so please don't re-derive it: our side sends what it is told. A local probe confirms `fetch` with a `Buffer` body transmits `Content-Type: image/png` verbatim, and the `filename` query parameter carries the extension. Zendesk's own docs say the header is what counts — *"The Content-Type header must contain a recognized MIME type that correctly describes the type of the uploaded file"*, with *"Failing to send a recognized, correct type may cause undesired behavior."*

### Setup

A throwaway ticket and a draft article, as before. **C5 needs a real PNG larger than 1 MB** — a full-resolution screenshot is fine. Keep it under ~2 MB: the bytes travel as base64 inside the tool call, so a 2 MB file is ~2.7 MB of JSON.

### Scenarios

`content_type` is either passed explicitly or **omitted entirely** — omitted means the field is absent from the call, not empty.

| id | Call | `content_type` | Observe | Expected |
| --- | --- | --- | --- | --- |
| **C1** | `add_private_note` + 1 small PNG | `"image/png"` | re-read the ticket, report the attachment's stored `content_type` | record verbatim |
| **C2** | `add_private_note` + 1 small PNG | *omitted* | same | record verbatim |
| **C3** | `create_article_attachment` on the draft | `"image/png"` | `list_article_attachments`, report `content_type` | record verbatim |
| **C4** | `create_article_attachment` on the draft | *omitted* | same | record verbatim |
| **C5** | `add_private_note` + the **> 1 MB** PNG | `"image/png"` | then `get_ticket_attachments` on that ticket | image content returned inline, well under 120 s — report the byte size and the wall-clock |

There is no pass/fail on C1–C4: they are a **measurement**. The four cells are the finding. What each outcome means:

| Outcome | Reading |
| --- | --- |
| C1 `image/png`, C2 `application/unknown` | our `application/octet-stream` default is the cause. Fix: derive the type from the `file_name` extension instead — an agent omitting an optional field currently guarantees a mislabelled attachment |
| C1 also `application/unknown` | Zendesk's Uploads API overrides the header. The fix is not on the upload side, and the read path needs to tolerate it |
| C3/C4 mirror C1/C2 | the two paths agree; the reported asymmetry was about what was passed, not about Zendesk |
| C3/C4 differ from C1/C2 | the asymmetry is real and Uploads-API-specific |

**C5 depends on C1.** `get_ticket_attachments` only downloads bytes for `content_type` starting with `image/`, so if the stored type is not `image/*` the download never happens and A6 stays open — report it as `BLOCKED` on C1's result rather than as a C5 failure. That dependency is exactly why C5 passes `content_type` explicitly.

### N1 — the pre-send branch (manual, no tenant)

The one thing the plan in the description admits it cannot prove: *"never reached Zendesk, so nothing was applied. Retrying is safe."* MSW raises a failure with no syscall code, so this branch is unit-tested only. A real DNS failure produces a real `ENOTFOUND`.

This does not fit the rest of the brief — it needs a second server on a nonexistent subdomain, which no MCP tool can reach, and it touches no tenant and needs no credentials. From the repo root:

```bash
printf '{"accessToken":"probe"}' > /tmp/probe-token.json && chmod 600 /tmp/probe-token.json
ZENDESK_TOKEN_FILE=/tmp/probe-token.json ZENDESK_OAUTH_CLIENT_ID=probe \
  node dist/index.js no-such-subdomain-zz9x7q --mode all
```

Then call `get_current_user` over stdio. Expected:

```
Network error on GET https://no-such-subdomain-zz9x7q.zendesk.com/api/v2/users/me after 3 attempts: ENOTFOUND …
```

with no note appended — a `GET` never carries a write-safety note. Report the message verbatim and the wall-clock (three attempts with backoff should land well under a second).

**Known environment dependency, and the reason this isn't a committed scenario:** behind an egress proxy the DNS failure never reaches the client. I tried it in a sandboxed environment and got `Resource not found. Please verify the ID is correct. (Not Found)` in 1.5 s — the proxy answered with an HTTP 404. If that is what you see, report `N1: BLOCKED (proxied egress)` and move on; it is not a defect. A scenario that only passes on some networks would be worse than none in `validate-retry.mjs`.

### Deliberately not asked for

- **Real `429`s.** Reaching them means hammering a production tenant. S5 and S6 cover the replay and the cap deterministically.
- **Upload-token redaction in a real error.** The token only exists on the *second* aggregated upload, and a network failure kills the first, so the case cannot be staged. It stays unit-tested.

### Priority

C1 and C2 first — they are the whole diagnosis, and everything downstream depends on which reading they produce. Then C3/C4, then C5. N1 is independent and can go first or last.

### Reporting

One PR comment in English: the SHA tested, the four measured `content_type` values verbatim in a table, your reading of them against the matrix above, C5's size and wall-clock, and N1's message or its `BLOCKED` reason. Note the ids of anything left on the tenant so it can be cleaned up.

---

dlecan:
## Validation addendum report — 83e640199c37d07fa0c38c357eae846f0a735347

Ran the addendum on the current head. Part B was re-run on this SHA first: **11/11 pass, exit 0** (S9 at 37.7 s on the binary path, S10 cut at 32.6 s, S11 one request).

### C1–C4 — the measurement

`content_type` values exactly as Zendesk stored them, read back through the tools:

| id | Path | File | `content_type` passed | Stored as |
| --- | --- | --- | --- | --- |
| C1 | Uploads API (`add_private_note`) | 1×1 PNG, 70 bytes | `"image/png"` | `application/unknown` |
| C2 | Uploads API (`add_private_note`) | 1×1 PNG, 70 bytes | *omitted* | `application/unknown` |
| C3 | Help Center (`create_article_attachment`) | 1×1 PNG, 70 bytes | `"image/png"` | `image/png` |
| C4 | Help Center (`create_article_attachment`) | 1×1 PNG, 70 bytes | *omitted* | `image/png` |

C1 = C2 puts the reading in the second row of your matrix, but it does not stop there. C5 uploaded a **real** 1.27 MB PNG through the same Uploads API with `"image/png"` and got back `image/png`, which contradicts "the Uploads API overrides the header" as a general rule. So I added the cell that isolates the variable:

| id | Path | File | `content_type` passed | Stored as |
| --- | --- | --- | --- | --- |
| C5 | Uploads API (`add_private_note`) | 650×650 PNG, 1268603 bytes | `"image/png"` | `image/png` |
| **C2b** | Uploads API (`add_private_note`) | 650×650 PNG, 1268603 bytes | *omitted* | `image/png` |

### Reading

Within each path, the parameter changes nothing: C1 = C2, and C2b = C5. What changes the stored type is the **file**. A 70-byte 1×1 PNG lands as `application/unknown` on the Uploads API; the same format at 1.27 MB lands as `image/png`, whether or not the header says so — including when the call omits the field and our schema default sends `application/octet-stream`.

Three consequences for the fix you were scoping:

1. **The `application/octet-stream` default is not the cause.** C2b sends it and still stores `image/png`. Deriving the type from the file extension would not have changed any of the six cells measured here.
2. **Nothing on our side is misbehaving.** The header goes out as instructed — you had already ruled that out locally, and nothing here contradicts it. Zendesk's Uploads API decides the stored type from the content, and disagrees with itself on a degenerate file.
3. **The asymmetry in the earlier report was an artifact of the test fixture**, not of the transport. The 1×1 PNG is what produced `application/unknown`; the two paths do differ on that file (multipart says `image/png`, Uploads API says `application/unknown`), but on a real image they agree. I would not ship a fix off the 70-byte case.

### C5 — the large transfer, and A6 closed

The >1 MB payload cannot travel inside a tool call from a validator session (~1.7 M base64 chars), so the upload was driven by a script against the **real MCP server over stdio**, same mechanism as `validate-retry.mjs`; auth came from the server's own token store. The measured call, `get_ticket_attachments`, then ran both through that script (for a clean wall-clock) and through the real MCP tool (for the functional proof).

- uploaded: **1268603 bytes**, upload call **3746 ms**
- `get_ticket_attachments` returning that attachment inline: **1008 ms**, against a 120 s transfer deadline
- the real tool call returned the image content inline, `image/png`, `1268603 bytes`

A6 from the previous report is therefore no longer blocked: a real >1 MB binary download completes with about two orders of magnitude of headroom.

### N1 — BLOCKED, but not for the reason the addendum predicts

Message received, verbatim, in **2715 ms** (server boot included):

```
Resource not found. Please verify the ID is correct. (Not Found)
```

That is the outcome you described, so `N1: BLOCKED`. The cause is not proxied egress here:

```
zz9x7q-nonexistent-probe.invalid        -> FAILED ENOTFOUND
another-random-zz9x7q-test.zendesk.com  -> 216.198.53.2
```

A genuinely nonexistent domain does fail with `ENOTFOUND` on this network, so nothing is intercepting DNS. `*.zendesk.com` has a **wildcard DNS record**: every subdomain resolves, and the server answers 404. Since the client builds its URL from the subdomain, this probe cannot produce a pre-send `ENOTFOUND` on any network — it is a property of the domain, not of the environment. Worth correcting in the addendum note, and it reinforces the decision to keep it out of `validate-retry.mjs`.

### Tenant artifacts

Created and already cleaned up: draft article **38394851485469** (archived), ticket **#7452** (closed — 4 attachments: 38394905678237, 38394924147613, 38394969625501, 38394963324189). From the previous run, ticket **#7451** is closed and article **38390635947805** archived. Permanent deletion is not available through the API for either, so both need the admin UI if you want them gone.

One process note: creating throwaway tickets on this tenant is now restricted, so #7452 was a one-off. If a future addendum needs ticket writes, say so explicitly and expect a single reusable ticket rather than one per run.

dlecan:
Report verified against the addendum rather than taken on trust. It refutes the fix I was scoping, corrects a false statement in my own addendum, and closes A6. Outcome: **no code change** — on this PR or a follow-up.

**C2b is the cell that settled it, and it wasn't in my brief.** My 2×2 varied the parameter across the two paths but held the file constant, so it could not have distinguished "the parameter matters" from "the file matters" — every cell that returned `application/unknown` used the same 70-byte fixture. Adding *omitted × large file* is what isolates the variable, and it lands on the one answer my matrix had no row for.

The result kills my hypothesis outright: **C2b sends `application/octet-stream` and Zendesk still stores `image/png`.** Deriving the MIME type from the `file_name` extension would not have changed a single one of the six measured cells. That was the whole content of the follow-up PR I was scoping, so it is dead, and the "product wart" reduces to *Zendesk mislabels a 70-byte 1×1 PNG* — which also means the observation that started this was an artifact of the test fixture, not something a user hits. Good outcome: the cheapest fix is the one you don't ship.

I'd add one bound to the reading, since it rests on two files: what is established is that **our header is not the deciding input**, which is what the fix hinged on. Whether the trigger is byte size, dimensions, or something else about a degenerate PNG is unresolved and not worth resolving — no real content reproduces it.

**Correcting my addendum: the N1 note gave the wrong cause, and yours is right.** I wrote that proxied egress converts the DNS failure into a 404. I re-checked from here and that was a misattribution:

```
zz9x7q-nonexistent-probe.invalid           -> FAILED ENOTFOUND
no-such-subdomain-zz9x7q.zendesk.com       -> 216.198.53.2
another-random-zz9x7q-test.zendesk.com     -> 216.198.54.2
totally-made-up-9f3a1c.zendesk.com         -> 216.198.53.2
totally-made-up-9f3a1c.example.com         -> FAILED ENOTFOUND
```

Nonexistent domains do fail `ENOTFOUND` on this network; only `*.zendesk.com` resolves, to two different IPs. So the 404 came from a real Zendesk edge server, exactly as you diagnosed. The consequence is stronger than "BLOCKED here": **a bogus subdomain cannot reach the pre-send branch on any network**, because the client derives its URL from the subdomain and the wildcard guarantees a resolvable host. That is a property of the domain, and the addendum's note should be read as corrected by this comment. It also means the PR description's existing wording — that reaching this branch live needs a fault-injecting endpoint plus a base-URL override the client deliberately lacks — was right, and my addendum briefly made it look like a network quirk instead of a structural limit.

**A6 is closed.** 1268603 bytes down in 1008 ms against a 120 s deadline is about two orders of magnitude of headroom, and driving it over stdio for a clean wall-clock while also confirming it through the real tool call is the right way around the base64 ceiling — the payload genuinely cannot fit in a tool call.

**On the process note:** understood, and the restriction is a good default. Nothing further needs ticket writes — this closes the live-only questions. If a future addendum does, I'll name it explicitly and design it for one reusable ticket rather than one per run.

Thanks for extending the matrix instead of just filling it in. That is the second time in this PR that a validator's own scenario, not mine, produced the finding that mattered.

---

PR #249, opened 2026-08-22
feat: exit when the client disconnects instead of leaking a process

### LINKED ISSUE #246: Exit on stdin close in stdio mode
Running in stdio mode behind `npx`, the server never receives `SIGTERM`. The `npm exec` and `sh -c` links in the chain do not relay signals, so the only sign that the client is gone is EOF on stdin.

Today the process still exits, but only because nothing keeps the event loop alive once stdin closes. That is circumstantial, not by design: adding an HTTP keep-alive agent, a refresh timer or a persistent cache would turn it into a process that outlives every session.

Suggestion: listen for stdin close next to the existing signal handlers.

```js
process.on("SIGINT", shutdown);
process.on("SIGTERM", shutdown);
process.stdin.on("end", shutdown);   // client is gone, signals never arrive
```

`@bytebase/dbhub` does this and exits cleanly; `mongodb-lens`, which handles signals only and holds a database connection, accumulates orphans instead. Checked against 2.18.0.

dlecan:
## Summary

Behind `npx`, a stdio server never sees `SIGTERM` when its client goes away — the only notice is EOF on stdin, and nobody read it. This gives the server an explicit shutdown path and triggers it from that EOF, so an abandoned session releases its process instead of being left behind.

**The failure mode is observed, not theoretical.** `@bytebase/dbhub` exits cleanly on EOF. `mongodb-lens`, which handles signals only and holds a MongoDB connection with no timeout, accumulates orphans instead: **79 live instances** were counted on one workstation after 7 days of uptime. That machine had 64 GB of RAM and absorbed it; a 16 GB laptop at 180 days of uptime would not. The Mongo connection is exactly the "something long-lived holds the event loop" case.

**Where this server stands — and why the gap matters more than the number.** The same audit found **2** live instances of this server, not 79. That is the honest figure, and it is small for a specific reason: nothing here holds the loop open, so an abandoned process usually falls over on its own. The one retention path that does exist — the OAuth callback server in `auth/browser-oauth.ts`, a listening socket nobody `unref()`'d — is bounded by the 5-minute auth timeout. At ~127 MB resident per process (measured on this branch, default `namespace` mode, 52 tools), 2 instances is a few hundred megabytes, not gigabytes.

So the distance between 2 and 79 is not a property of this codebase's discipline. It is the absence of a long-lived handle, and nothing states that absence as a rule. Add one — an HTTP keep-alive agent, a refresh timer, a persistent cache, a database connection like the one above — and this server's number starts moving toward the other one, while nothing in the diff that introduced it would have looked like it touched process lifecycle. That is what this closes off, at the source rather than at the symptom.

## Linked issue

Closes #246

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tooling / CI

## Author checklist
- [x] `pnpm typecheck` passes
- [x] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md` — n/a, no tool change
- [x] If the change has **MCP-runtime-observable** behaviour: this PR carries a `## Functional validation plan` below — **executed by an independent validator against `5ecb20f`, all nine scenarios OK ([report](https://github.com/fruggr/zendesk-mcp-server/pull/249#issuecomment-5381671127), [rerun addendum](https://github.com/fruggr/zendesk-mcp-server/pull/249#issuecomment-5381806744)).**

## What the issue got right, and what it missed

The diagnosis is exact: `StdioServerTransport` attaches `'data'` and `'error'` listeners on stdin and nothing else, so `transport.onclose` fires only when we close it ourselves. EOF was read by nobody.

Two things widened the fix.

**There were no existing signal handlers to sit beside.** `grep -rn "SIGINT\|SIGTERM\|process.on(" src/` returned nothing on `main`. `tokenStore.dispose()`, `server.close()` and the HTTP transport's `close()` were all unreachable from the entrypoint.

**The orphan already existed.** The `unref()` discipline elsewhere is deliberate, but the OAuth callback server is a listening `http.Server` that is not `unref()`'d — the five-minute window described above, and the likeliest explanation for the 2 instances that were found. That is also why shutdown ends in an explicit `exit(0)` rather than a drained loop: draining would wait out that same five minutes.

## Design notes for the reviewer

**The watchdog is load-bearing, not defensive.** Registering `process.on('SIGTERM')` *removes* Node's default terminate and makes the exit our responsibility. A cleanup stalled on an in-flight Zendesk request would otherwise produce a process `SIGTERM` cannot kill — a worse version of the bug. It is armed before cleanup and `unref()`'d, bounding shutdown at 3s, inside `docker stop`'s 10s and Kubernetes' 30s.

`unref()` does not defeat it, though that is easy to assume: an unref'd timer still fires while the loop runs and only declines to hold the loop open by itself. So a hung cleanup that holds something (a socket, an in-flight request) is forced out at the grace, and one that holds nothing lets the loop empty and Node exit before the watchdog was ever needed. Both terminate.

**Handlers are installed after the transport connects**, because the SDK's stdin `'data'` listener is what puts stdin in flowing mode and `'end'` only fires there. That leaves a startup window — ~0.3s on normal hardware, up to 27s on a loaded slow one — where a signal takes Node's default action instead. That is deliberate and harmless: in that window there is no connected server to close and the token store's only handle is an `unref()`'d timer, so there is nothing a graceful path would do differently. Validation surfaced it as `exit=130`; the plan now polls for readiness rather than assuming a delay.

**A correction worth reading, because it reverses a claim made earlier in this branch.** I initially justified `watchStdin: false` for HTTP by asserting that `/dev/null` stdin would shut a daemonised server down at boot. That is false, and the third commit corrects it. `'end'` fires only once a stream is actually *read*; attaching an `'end'` listener does not resume a paused stream. Measured:

| stdin | `'data'` listener | `'end'` |
| --- | --- | --- |
| `/dev/null` | yes | fires immediately |
| `/dev/null` | no | never fires |
| closed pipe | no | never fires |

In HTTP mode nothing reads stdin, so EOF never arrives regardless. `watchStdin: false` is therefore not averting a live failure — it refuses to depend on an invariant nothing states, since the day anything in the process reads stdin a server carrying an EOF handler would start exiting as soon as its supervisor handed it `/dev/null`. The stdio half is unaffected: the SDK *does* attach `'data'`, so EOF there is genuinely live, which is why `scripts/smoke-test.mjs` had to stop spawning with `stdio: ['ignore', …]` and the Node 20 tarball job had to hold stdin open.

**Every new smoke check was verified against an injected regression rather than assumed.** This took three passes, and the failures are worth knowing about because they were all the same shape — a check reporting success on a broken server:

1. The EOF check initially passed even with the feature removed, because the old accidental exit produced the same observable — precisely the distinction the issue draws. Fixed by requiring `shutdown_started reason=stdin_eof`.
2. That still accepted a *forced* shutdown: a cleanup that hangs is cut short by the watchdog, which also exits 0, at 3s, inside the check's 5s deadline. The fourth commit adds `shutdown_complete reason=stdin_eof` and rejects any run logging `shutdown_forced`. Both rejections were confirmed reachable by hanging a cleanup with and without something holding the event loop.
3. The HTTP check sent SIGTERM with no SIGKILL escalation, so a server that stopped honouring SIGTERM — the regression it exists to catch — would have hung the run until the CI job timeout instead of failing. It now escalates and fails on `hadToKill`.

The HTTP check is documented as *unable* to tell `watchStdin: false` apart from the paused stream; the unit suite is what asserts the listener's absence.

**Placement.** `src/index.ts` and `src/transports/stdio.ts` are excluded from coverage as thin bootstraps, so logic there is invisible to both the coverage ratchet and Stryker. The module lives in `src/utils/` (inside Stryker's `mutate` scope) and takes its `process` access through an injected runtime so the adapter is tested rather than `v8 ignore`d.

Rationale in full: `docs/decisions/lifecycle-shutdown.md`.

## Gates

`pnpm typecheck`, `pnpm check`, `pnpm test` (932), `pnpm test:coverage` (97.85 / 85.80 / 98.91 / 98.36 — up on all four axes), `pnpm test:smoke` (4 checks), `pnpm test:mutation:diff origin/main HEAD` → **53/53 mutants killed, 0 escaped**. CI green on `5ecb20f`.

---

## Functional validation plan

✅ **Executed against `5ecb20f` — all nine scenarios OK** ([report](https://github.com/fruggr/zendesk-mcp-server/pull/249#issuecomment-5381671127), [rerun addendum](https://github.com/fruggr/zendesk-mcp-server/pull/249#issuecomment-5381806744)). No code change came out of it: every finding was against this plan's own timing assumptions, and they are fixed below.

**For an independent validator.** Please report as a PR comment in English.

### ⚠️ Read this first — this plan deviates from the usual shape

This change is observable at the **process lifecycle** level, not through the MCP tool surface. The tool surface is unchanged, so calling `mcp__zendesk__*` tools cannot exercise it. Two consequences:

1. **Do not close the stdin of your own running server** — in stdio mode that is now a shutdown trigger and you would end your own session. All lifecycle scenarios spawn **separate, short-lived** server processes.
2. You will need a runnable build. `dist/` is normally already built at session start; if it is stale, run `pnpm build` once. This is the one setup step the plan asks for, and it is unavoidable here.

Capture the SHA you tested: `git rev-parse HEAD` (expected `5ecb20f` or later).

All commands below run from the repo root. `LOG_LEVEL=info` is required — the markers are logger events. The positional `smoke-test` is consumed as the Zendesk subdomain, so no credentials are needed: every scenario asserts lifecycle only and never reaches Zendesk.

**No scenario below uses a fixed delay to wait for the server.** Boot was measured at ~0.3s on a workstation, 1.9–2.2s warm on a slow one, and 21–27s on that same machine under load. Every wait polls for the readiness marker instead, so the plan adapts rather than needing constants retuned per machine.

### Scenarios

Every command below was run on this branch before being written here, so the expected output is observed, not predicted. Use `LOG_LEVEL=info` and **redirect to a file** rather than piping to `grep`.

Set `SP=$(mktemp -d)` once for the log files.

**V1 — normal operation is not broken.** Call any read-only tool in your live session, e.g. `mcp__zendesk__get_current_user`. It must succeed as before, and your session must stay up: the shutdown wiring must never fire during normal use.

**V2 — the core of #246: stdio shuts down *deliberately* on EOF, no signal sent.**
```bash
LOG_LEVEL=info node dist/index.js smoke-test < /dev/null > $SP/v2.log 2>&1; echo "exit=$?"
grep -E 'shutdown_started|shutdown_complete' $SP/v2.log
```
Expect `exit=0` with `shutdown_started reason=stdin_eof` **and** `shutdown_complete reason=stdin_eof`. The markers are the point — a bare `exit=0` also happened before this change, which is exactly the "circumstantial, not by design" distinction the issue draws.

**V3 — EOF does not truncate in-flight work.**
```bash
printf '%s\n%s\n' \
  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"v","version":"0"}}}' \
  '{"jsonrpc":"2.0","id":2,"method":"tools/list"}' \
  | LOG_LEVEL=info node dist/index.js smoke-test 2>/dev/null | grep -o '"id":[0-9]*' | sort -u
```
Expect both `"id":1` and `"id":2` — the server answers every queued request before EOF stops it.

**V4 / V5 — HTTP ignores stdin, and drains on SIGTERM.** Polls for readiness rather than racing a fixed `timeout`, and asserts the two things separately: that the server survived an EOF it received at startup, and that SIGTERM then drained it.
```bash
LOG_LEVEL=info node dist/index.js smoke-test --transport http --port 0 \
  < /dev/null > $SP/v4.log 2>&1 & PID=$!
for i in $(seq 1 120); do grep -q 'http_transport_ready' $SP/v4.log && break; sleep 0.5; done
sleep 2
kill -0 $PID 2>/dev/null && echo "alive: yes"
kill -TERM $PID; wait $PID; echo "exit=$?"
grep -E 'http_transport_ready|shutdown_|stdin_eof' $SP/v4.log
```
Expect `alive: yes` — its stdin was at EOF from the very first read, and it is still running. Then `exit=0` with `shutdown_started reason=SIGTERM` and `shutdown_complete reason=SIGTERM`, and **no `stdin_eof` anywhere**. A process already gone before the `kill -0`, or any `stdin_eof`, is a FAIL.

**V6 — SIGINT works on stdio.** stdin is held open by a fifo the shell keeps on a spare descriptor. Wait for the readiness marker rather than a fixed delay: handlers are installed only once the transport is connected, so a signal sent during boot takes Node's default action and you get `exit=130` with an empty log — a timing artefact, not a failure of the change.

A fifo rather than `sleep N |` on purpose: a holding `sleep` keeps the shell busy for its full duration after the server is already gone, which cost one validator up to 90s per scenario. Closing the descriptor releases immediately, so the poll can be generous for free — measured at 1s total here.
```bash
mkfifo $SP/hold6; exec 9<>$SP/hold6
LOG_LEVEL=info node dist/index.js smoke-test < $SP/hold6 > $SP/v6.log 2>&1 & PID=$!
for i in $(seq 1 120); do grep -q 'stdio_transport_ready' $SP/v6.log && break; sleep 0.5; done
kill -INT $PID 2>/dev/null; wait $PID; echo "exit=$?"
exec 9>&-
grep -E 'shutdown_started|shutdown_complete' $SP/v6.log
```
Expect `exit=0` with `reason=SIGINT` on both markers.

**V7 — dev mode behaves like stdio.**
```bash
LOG_LEVEL=info node dist/index.js smoke-test --dev < /dev/null > $SP/v7.log 2>&1; echo "exit=$?"
grep -E 'dev_mode_enabled|shutdown_started' $SP/v7.log
```
Expect `dev_mode_enabled`, then `shutdown_started reason=stdin_eof`, `exit=0`.

**V8 — idempotence: one shutdown, not two.** Same fifo and readiness poll as V6, for the same reasons.
```bash
mkfifo $SP/hold8; exec 8<>$SP/hold8
LOG_LEVEL=info node dist/index.js smoke-test < $SP/hold8 > $SP/v8.log 2>&1 & PID=$!
for i in $(seq 1 120); do grep -q 'stdio_transport_ready' $SP/v8.log && break; sleep 0.5; done
kill -TERM $PID 2>/dev/null; kill -TERM $PID 2>/dev/null; wait $PID; echo "exit=$?"
exec 8>&-
grep -c 'info. shutdown_started' $SP/v8.log
```
Expect `exit=0` and a count of exactly **1**.

**V9 — the committed harness agrees.** `pnpm test:smoke` — all four checks print `ok`, including `shut down deliberately on stdin EOF` and `survived stdin EOF in http mode`.

### Long-running behaviour — do not wait it out

The 3s watchdog that bounds a stalled cleanup is **already covered by unit tests** with an injected runtime and fake timer (`tests/unit/utils/shutdown.test.ts` → "forces the exit when cleanup never settles", "arms an unref-ed timer at the grace period"). Please do not try to reproduce a hung cleanup from outside the process. The only thing worth confirming externally is the negative: **no scenario above should ever print `shutdown_forced`** — if one does, a normal shutdown is hitting the watchdog and that is a FAIL.

### Priorities

Do **V2**, **V4**, **V3** first — they are the issue's actual claim, the regression it must not cause, and the correctness risk of shutting down too eagerly. **V1** is the cheap "nothing is broken" check. **V5–V9** round out the surface.

### Reporting

Post one PR comment with a per-id verdict (`OK` / `FAIL` / `BLOCKED`) plus the evidence you observed — the exit code and the relevant log lines, not just an assertion that it passed. Finish with an overall green/failing summary and the SHA you tested. If anything is `BLOCKED` on environment rather than code, say so explicitly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,11 @@ jobs:
       - name: Verify tarball runs on Node 20
         run: |
           set -euo pipefail
-          OUT=$(env LOG_LEVEL=info timeout 5 zendesk-mcp-server smoke-test 2>&1 || true)
+          # `sleep` holds stdin open for the whole run. A step's stdin is
+          # /dev/null, which reaches EOF immediately, and EOF now shuts the
+          # stdio server down — without this the job would silently stop
+          # testing that the tarball *runs* and start testing that it exits.
+          # The EOF path itself is asserted in scripts/smoke-test.mjs.
+          OUT=$(sleep 6 | env LOG_LEVEL=info timeout 5 zendesk-mcp-server smoke-test 2>&1 || true)
           echo "$OUT"
           echo "$OUT" | grep -q "stdio_transport_ready"
--- a/docs/decisions/lifecycle-shutdown.md
+++ b/docs/decisions/lifecycle-shutdown.md
@@ -55,32 +55,48 @@ bug being fixed.
 
 The watchdog is what keeps that from happening: it is armed *before* cleanup
 runs, so a cleanup that never settles is still bounded. It is `unref()`'d, so
-having a grace period does not mean waiting one out. `SHUTDOWN_GRACE_MS` is 3s —
-under the 10s that systemd and Docker allow between their own `SIGTERM` and
-`SIGKILL`, because being killed by the supervisor is the unclean exit this is
-meant to avoid.
+having a grace period does not mean waiting one out. `SHUTDOWN_GRACE_MS` is 3s,
+inside the tightest common supervisor grace — `docker stop` allows 10s and
+Kubernetes 30s before their own `SIGKILL` (systemd is far laxer, 90s by default)
+— because being killed by the supervisor is the unclean exit this is meant to
+avoid.
 
-### `/dev/null` stdin reaches EOF immediately
+### EOF only exists in flowing mode
 
-A daemonised HTTP server is routinely handed `/dev/null` on stdin, which is at
-EOF from the first read. A stdin listener registered in HTTP mode would shut the
-server down at boot — in production, and in CI, where a GitHub Actions step's
-stdin is likewise `/dev/null`.
+`'end'` fires when a readable stream is *read* to exhaustion. Attaching an
+`'end'` listener does not by itself resume a paused stream, and Node leaves stdin
+paused until something attaches `'data'`. Measured:
 
-So the trigger set is per transport, and `watchStdin` is a required option rather
-than a default: stdio watches stdin, HTTP does not. The unit suite asserts the
-*absence* of the listener in HTTP mode, which is the assertion that would catch a
-regression here before it reaches an operator.
+| stdin | `'data'` listener | `'end'` |
+| --- | --- | --- |
+| `/dev/null` | yes | fires immediately |
+| `/dev/null` | no | never fires |
+| closed pipe | no | never fires |
 
-Two consequences worth naming:
+Two consequences, and they pull in opposite directions.
 
-- In stdio mode, `zendesk-mcp-server < /dev/null` now exits immediately. This is
-  intended — there is no client on the other end — and is documented as a
-  symptom in `troubleshooting.md`.
-- `scripts/smoke-test.mjs` had to stop spawning with `stdio: ['ignore', …]`, and
-  the Node 20 tarball job in `ci.yml` had to hold stdin open, or both would have
-  quietly stopped testing that the server *runs* and started testing that it
-  exits.
+**In stdio mode the SDK attaches `'data'`, so EOF is genuinely live** — including
+at startup. `zendesk-mcp-server < /dev/null` now exits immediately, which is
+intended (there is no client on the other end) and is documented as a symptom in
+`troubleshooting.md`. It is also why `scripts/smoke-test.mjs` had to stop
+spawning with `stdio: ['ignore', …]` and the Node 20 tarball job in `ci.yml` had
+to hold stdin open: otherwise both would have quietly stopped testing that the
+server *runs* and started testing that it exits.
+
+**In HTTP mode nothing reads stdin**, so the stream stays paused and would never
+report EOF even on `/dev/null`. `watchStdin: false` there is therefore not
+averting a live failure — it is refusing to depend on that.  Nothing states that
+stdin must stay unread in the HTTP process; the day a library, a debugger or a
+future feature attaches a `'data'` listener, a server carrying an EOF handler
+would start exiting as soon as its supervisor handed it `/dev/null`. Making
+`watchStdin` a required option rather than a default keeps the choice at each
+call site instead of resting on that invariant.
+
+The unit suite asserts the *absence* of the stdin listener in HTTP mode, because
+that is the only place the distinction is observable. The smoke check that closes
+stdin against a running HTTP server pins the user-visible contract, but cannot
+tell the flag apart from the paused stream and would keep passing if only one of
+the two survived.
 
 ## Why the runtime is injected
 
--- a/docs/decisions/lifecycle-shutdown.md
+++ b/docs/decisions/lifecycle-shutdown.md
@@ -55,7 +55,18 @@ bug being fixed.
 
 The watchdog is what keeps that from happening: it is armed *before* cleanup
 runs, so a cleanup that never settles is still bounded. It is `unref()`'d, so
-having a grace period does not mean waiting one out. `SHUTDOWN_GRACE_MS` is 3s,
+having a grace period does not mean waiting one out.
+
+`unref()` does not defeat it, though that is easy to assume. An unref'd timer
+still fires while the loop is running; it only declines to keep the loop alive by
+itself. The two cases therefore both terminate:
+
+- something still holds the loop (an in-flight request, a socket) — the loop is
+  running, so the watchdog fires and forces the exit;
+- nothing holds it — the loop empties and Node exits on its own, before the
+  watchdog was ever needed.
+
+`SHUTDOWN_GRACE_MS` is 3s,
 inside the tightest common supervisor grace — `docker stop` allows 10s and
 Kubernetes 30s before their own `SIGKILL` (systemd is far laxer, 90s by default)
 — because being killed by the supervisor is the unclean exit this is meant to
--- a/docs/decisions/lifecycle-shutdown.md
+++ b/docs/decisions/lifecycle-shutdown.md
@@ -1,0 +1,106 @@
+# Process lifecycle: why the server owns its exit
+
+> **Build documentation, not user documentation.** This records why
+> `src/utils/shutdown.ts` exists and why its triggers differ per transport.
+> Client-facing symptoms are in [`docs/troubleshooting.md`](../troubleshooting.md);
+> operator concerns in [`docs/http-deployment.md`](../http-deployment.md).
+
+| | |
+| --- | --- |
+| **Status** | Decided and applied |
+| **Date** | 2026-08-22 |
+| **Applied in** | [#249](https://github.com/fruggr/zendesk-mcp-server/pull/249) ([#246](https://github.com/fruggr/zendesk-mcp-server/issues/246)) |
+| **Question** | How should the server learn that it is done, and what must it guarantee once it does? |
+| **Answer** | One idempotent shutdown path, triggered by `SIGINT`/`SIGTERM` on both transports and additionally by stdin EOF on stdio only, always ending in an explicit `exit(0)` bounded by a watchdog. |
+
+## Why EOF at all
+
+Behind `npx`, the `npm exec` and `sh -c` links in the chain do not relay signals,
+so a stdio server never receives `SIGTERM` when its client goes away. The only
+remaining evidence is EOF on stdin — and the SDK does not read it.
+`StdioServerTransport` attaches exactly two listeners:
+
+```js
+this._stdin.on('data',  this._ondata);
+this._stdin.on('error', this._onerror);
+```
+
+Never `'end'`, never `'close'`. `transport.onclose` fires only when *we* call
+`close()`. So EOF was observed by nobody.
+
+## Why it was not already broken
+
+Before this change the process still exited, because nothing kept the event loop
+alive. That was mostly discipline rather than luck — every timer is deliberately
+`unref()`'d, and says so at its declaration (`auth/token-store.ts`,
+`transports/http.ts`).
+
+But the discipline had a hole, and it was not hypothetical. The OAuth callback
+server in `auth/browser-oauth.ts` is a listening `http.Server` that is **not**
+`unref()`'d; only its `authTimeout` is. A client that disappeared mid-auth left
+the process alive for up to `AUTH_TIMEOUT_MS` — five minutes — on 2.18.0.
+
+That hole is also why the shutdown ends in an explicit `exit(0)` rather than
+letting the loop drain: draining would wait out that same five minutes.
+
+## Two constraints that shaped the design
+
+### Registering a signal handler removes a guarantee
+
+Without a handler, `SIGTERM` terminates the process by Node's default. Adding
+`process.on('SIGTERM', …)` **replaces** that default and makes the exit our
+responsibility. A cleanup that stalls on an in-flight Zendesk request would
+therefore produce a process that `SIGTERM` cannot kill — a worse version of the
+bug being fixed.
+
+The watchdog is what keeps that from happening: it is armed *before* cleanup
+runs, so a cleanup that never settles is still bounded. It is `unref()`'d, so
+having a grace period does not mean waiting one out. `SHUTDOWN_GRACE_MS` is 3s —
+under the 10s that systemd and Docker allow between their own `SIGTERM` and
+`SIGKILL`, because being killed by the supervisor is the unclean exit this is
+meant to avoid.
+
+### `/dev/null` stdin reaches EOF immediately
+
+A daemonised HTTP server is routinely handed `/dev/null` on stdin, which is at
+EOF from the first read. A stdin listener registered in HTTP mode would shut the
+server down at boot — in production, and in CI, where a GitHub Actions step's
+stdin is likewise `/dev/null`.
+
+So the trigger set is per transport, and `watchStdin` is a required option rather
+than a default: stdio watches stdin, HTTP does not. The unit suite asserts the
+*absence* of the listener in HTTP mode, which is the assertion that would catch a
+regression here before it reaches an operator.
+
+Two consequences worth naming:
+
+- In stdio mode, `zendesk-mcp-server < /dev/null` now exits immediately. This is
+  intended — there is no client on the other end — and is documented as a
+  symptom in `troubleshooting.md`.
+- `scripts/smoke-test.mjs` had to stop spawning with `stdio: ['ignore', …]`, and
+  the Node 20 tarball job in `ci.yml` had to hold stdin open, or both would have
+  quietly stopped testing that the server *runs* and started testing that it
+  exits.
+
+## Why the runtime is injected
+
+`installShutdown` takes its `process` access through a `ShutdownRuntime`, and
+`createRuntime(process)` is what adapts the real one. Not ceremony: `index.ts`
+and `transports/stdio.ts` are excluded from coverage as thin bootstraps
+(`vitest.config.ts`), so logic placed there is invisible to both the coverage
+ratchet and Stryker. Putting the module in `src/utils/` instead — where
+`logger.ts` already sets the precedent for cross-cutting infrastructure — brings
+it inside Stryker's `mutate` scope, and injecting the process is what lets the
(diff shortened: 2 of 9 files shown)


dlecan:
@coderabbitai review

dlecan:
## Functional validation report (5ecb20f8c941ba2a0a25cf3bfff41e575a203c85)

Validated on the PR branch as checked out, `dist/index.js` already newer than every `src/**` file (no `pnpm build` needed before the run; `pnpm test:smoke` rebuilt it anyway in V9). `LOG_LEVEL=info` throughout, output redirected to files as the plan asks.

| ID | Verdict | Evidence |
| -- | ------- | -------- |
| V1 | OK | `get_current_user` returned its usual card: heading `## <name> (<id>)` followed by Email, Role, Role type, Active and Organization fields, all populated (values redacted). Repeated at the very end of the run, after every lifecycle scenario, with the same result, so the session never went down. |
| V2 | OK | `exit=0`. Log order: `stdio_transport_ready`, then `shutdown_started reason=stdin_eof`, then `shutdown_complete reason=stdin_eof`. Both markers present, so this is the deliberate path and not the old accidental exit. |
| V3 | OK | `exit=0`, stdout carries both `"id":1` and `"id":2` (4 lines total, responses plus notifications). Nothing truncated by the EOF. |
| V4 | OK | `exit=124`, so `timeout` had to signal it and the server was still alive 4s after its stdin hit EOF. Log has `http_transport_ready host=0.0.0.0 port=<int>` and no `stdin_eof` anywhere. |
| V5 | OK | Same run: `shutdown_started reason=SIGTERM` then `shutdown_complete reason=SIGTERM`. |
| V6 | OK, after a timing fix (see note below) | `exit=0` with `shutdown_started reason=SIGINT` and `shutdown_complete reason=SIGINT`. The first attempt, run exactly as written, gave `exit=130` and a zero-byte log. |
| V7 | OK | `dev_mode_enabled`, then `shutdown_started reason=stdin_eof` and `shutdown_complete reason=stdin_eof`, `exit=0`. |
| V8 | OK, same timing fix | `exit=0`, `grep -c 'info. shutdown_started'` returns exactly 1 for two SIGTERMs, and a single `shutdown_complete reason=SIGTERM`. |
| V9 | OK | Four `[smoke] ok` lines, including `shut down deliberately on stdin EOF` and `survived stdin EOF in http mode`. Script exit 0. |

The negative the plan flags also holds: `grep -l shutdown_forced` over all eight scenario logs matches nothing.

### Note on the V6 and V8 timing (environment, not code)

Run verbatim, V6 produced `exit=130` and an empty log, which is Node's default SIGINT behaviour, meaning the signal landed before the handler existed. This machine is Android/Termux under PRoot with a slow filesystem: a timed V2 took 5.3s wall for the whole spawn to EOF to exit cycle, and the poll below saw `stdio_transport_ready` at roughly 4.5 to 6s after spawn. So `sleep 1.5` signals a server that has not finished booting, and `sleep 5` on the holding end of the pipe can expire before it is ready either way.

Replacing the fixed delay with a poll on the readiness marker, stdin held open for 30s, made both scenarios pass:

```bash
sleep 30 | LOG_LEVEL=info node dist/index.js smoke-test > $SP/v6b.log 2>&1 & PID=$!
for i in $(seq 1 60); do grep -q 'stdio_transport_ready' $SP/v6b.log && break; sleep 0.5; done
kill -INT $PID; wait $PID; echo "exit=$?"
```

The behaviour under test is confirmed; only the plan's hardcoded delays are fragile on slow hardware. Worth loosening them if this plan may be run somewhere other than a fast workstation.

## Summary

Green. All nine ids OK on 5ecb20f8c941ba2a0a25cf3bfff41e575a203c85. V6 and V8 needed the readiness poll described above before they could exercise what they were written to exercise, which is an environment constraint rather than a finding against the change. Nothing BLOCKED: every scenario asserts lifecycle only and none of them reached Zendesk.

dlecan:
Report verified against the plan's expected observables, not just its verdicts. It tested `5ecb20f`, which is the current head, and every piece of evidence matches what the scenario specified — including the negative (`shutdown_forced` absent from all eight logs).

**On the V6/V8 timing.** Your diagnosis is right, and the code confirms the mechanism: `installShutdown` is called after `connectStdio` resolves, because the SDK's stdin `'data'` listener is what puts stdin in flowing mode and `'end'` only fires there. A signal arriving before that takes Node's default action — hence `exit=130` (128+SIGINT) and an empty log, since `stdio_transport_ready` had not been written yet.

I looked at whether to close that window by registering the signal handlers earlier, and decided against it. In that window there is no connected server to close, and the token store's only handle is an `unref()`'d timer, so a graceful path would do nothing a default terminate does not already do. Moving registration earlier would split it in two for a cosmetically nicer exit code during a boot-time abort — not worth widening this PR. Noted in the description instead, so the window is documented rather than incidental.

**The plan was the fragile part, so that is what changed.** V6 and V8 now poll for `stdio_transport_ready` instead of assuming a fixed `sleep 1.5`, using the loop you demonstrated, with stdin held for 30s. Both carry a note explaining that `exit=130` there is a timing artefact rather than a failure of the change, with the ~0.3s vs 5s+ boot spread called out. Thanks for catching it — a plan that only works on fast hardware is a plan that silently stops validating.

No code change: the behaviour under test was confirmed on every id.

---

dlecan:
## Addendum: plan delays relaxed, affected ids rerun (5ecb20f8c941ba2a0a25cf3bfff41e575a203c85)

Same commit as the report above, no code touched. I widened the timing windows in the plan itself and then reran the three scenarios that depend on them, verbatim from the updated text.

What changed in the plan:

- V4 / V5: `timeout 4` becomes `timeout 20`, plus a line saying the window exists so `http_transport_ready` lands before the signal, and that a missing marker means raising it rather than filing a FAIL.
- V6 and V8: stdin is held for 90s instead of 30s and the readiness poll waits up to 60s instead of 30s.

Why those numbers. Boot to `http_transport_ready` measured 1.9 to 2.2s across four warm sequential runs on this machine, but 21 to 27s when anything else was running alongside, and one V6 rerun saw `stdio_transport_ready` at 17s. A 4s window with roughly 2s of headroom is the same trap V6 already had, so it was worth the same fix.

| ID | Verdict | Evidence |
| -- | ------- | -------- |
| V4 | OK | `exit=124`, `http_transport_ready host=0.0.0.0 port=<int>`, no `stdin_eof` in the log |
| V5 | OK | same run: `shutdown_started reason=SIGTERM` then `shutdown_complete reason=SIGTERM` |
| V6 | OK, no fix needed this time | `exit=0`, `shutdown_started reason=SIGINT` and `shutdown_complete reason=SIGINT`, readiness reached at 17s |
| V8 | OK | `exit=0`, `grep -c 'info. shutdown_started'` returns 1 for two SIGTERMs, readiness at 5s |

No `shutdown_forced` in any of the rerun logs, and `ps` showed no leftover `dist/index.js` process afterwards.

One cost worth knowing: the holding `sleep` keeps the shell busy after the server is gone, so V6 and V8 now take up to 90s of wall clock each even though the assertion resolves in seconds. I tried releasing it with `kill %1` and could not show it helped, so the plan does not carry that line.

dlecan:
Your relaxations are in the plan, and the V4 point was the one I should have caught myself: I hardened V6/V8 against exactly this and then left V4 racing a fixed 4s window with ~2s of headroom. Same trap, one scenario over.

Two things I changed rather than take verbatim, both verified here before writing them down.

**The 90s cost is fixable.** `sleep N |` keeps the shell busy for the full N after the server is already gone, which is why `kill %1` did not help — the wait is on the pipeline, not the server. Holding stdin with a fifo on a spare descriptor removes the sleeping child entirely, and closing the descriptor releases immediately:

```bash
mkfifo $SP/hold6; exec 9<>$SP/hold6
LOG_LEVEL=info node dist/index.js smoke-test < $SP/hold6 > $SP/v6.log 2>&1 & PID=$!
for i in $(seq 1 120); do grep -q 'stdio_transport_ready' $SP/v6.log && break; sleep 0.5; done
kill -INT $PID 2>/dev/null; wait $PID; echo "exit=$?"
exec 9>&-
```

Measured here: V6 **1s** total wall clock, V8 **0s**, against the up-to-90s you were paying. The poll bound is now 60s and costs nothing when readiness is fast, so the generous ceiling is free rather than a tax.

**V4 polls instead of using a bigger `timeout`.** `timeout 20` works, but it still encodes a guess, and 27s under load would break it again. Polling for `http_transport_ready`, then holding 2s, then signalling ourselves adapts on its own — 3s wall here, and it would sit through your 27s boot with no constant to retune. It also splits an assertion the old form conflated: `exit=124` only ever proved "still running", and said nothing about the drain. Now `kill -0` proves survival and `exit=0` plus the SIGTERM markers prove the graceful path, separately.

Verified locally before landing: V4 `alive: yes` then `exit=0` with both SIGTERM markers and zero `stdin_eof`; V6 `exit=0` with both SIGINT markers; V8 `exit=0` with `shutdown_started` counted exactly once.

The plan now says up front that no scenario uses a fixed delay, and carries your measurements (0.3s workstation, 1.9–2.2s warm, 21–27s loaded) as the reason. Thanks for the rerun — the second pass is what turned "V6 was fragile" into "fixed delays are the wrong mechanism here".

---

PR #258, opened 2026-08-31
fix: cap attachment payloads so an oversized message cannot drop the session

### LINKED ISSUE #205: Cap attachment payloads: an oversized JSON-RPC message now kills the whole stdio session
## Summary

`@modelcontextprotocol/sdk` 1.30.0 introduces a **10 MB cap on the stdio `ReadBuffer`** (`STDIO_DEFAULT_MAX_BUFFER_SIZE = 10485760`). A message exceeding it does not merely fail the call: the transport emits an error and **closes**, tearing down the MCP session.

Our tool surface can produce such messages in **both directions**, because neither the inbound attachment upload nor the outbound embedded-image response has a total-size budget.

## Evidence

Measured on `StdioServerTransport` / `ReadBuffer` with a 12 MB inbound message and a 20 MB outbound one, on the same code, only the SDK version differing:

```
SDK 1.29.0 : inbound 12 MB  -> onerror (none),                             onclose false
SDK 1.30.0 : inbound 12 MB  -> "ReadBuffer exceeded maximum size of 10485760 bytes", onclose TRUE
SDK 1.29.0 : outbound 20 MB -> client ReadBuffer accepted (parsed)
SDK 1.30.0 : outbound 20 MB -> client ReadBuffer THREW
```

Note `onclose = true`: the failure mode is a dead session, not a failed tool call.

## The two exposed paths

**Inbound — `create_article_attachment`.** Its `file_base64` parameter (`src/tools/help-center.ts`) is a `z.string().min(1)` with **no maximum**. Base64 inflates by ~33%, so a source file larger than ~7.5 MB produces a request over the cap and kills the stdio session.

**Outbound — `get_ticket_attachments` / `get_ticket(include_comments=true)`.** `MAX_ATTACHMENT_BYTES` (5 MB, `ZENDESK_MAX_ATTACHMENT_BYTES`) is enforced **per image** (`src/tools/tickets.ts:156`), and up to `MAX_EMBEDDED_IMAGE_COUNT` (10) images are embedded in a single response. There is no aggregate budget, so the real ceiling for one message is ~7.5 MB of raw bytes across *all* images: **two screenshots of ~3.8 MB already exceed it**, which is reachable on a real ticket.

## Scope note

The outbound direction depends on the **client's** SDK version, not ours — it is not a regression introduced by the SDK bump on our side, and it is not a reason to hold #201 (that upgrade is backward compatible; see the analysis there). But it is a new ecosystem failure mode our tool surface can trigger, and it fails in the worst possible way: silently, by dropping the session.

The HTTP transport is unaffected inbound (`MAX_BODY_BYTES` already caps request bodies at 4 MB and returns 413); stdio had no equivalent guard. The response direction is uncapped on both transports.

## Proposed changes

- [ ] Add a documented maximum to `file_base64` on `create_article_attachment`, sized so the encoded payload stays under the stdio cap, with an error message that names the limit and the actual size so the model can react (downscale, split, or link instead of upload).
- [ ] Introduce an **aggregate** byte budget for embedded images, enforced alongside the existing per-image cap and the image count, with the same explicit skip-reason mechanism already used at `src/tools/tickets.ts:156-165` (`skipped: total embedded image budget reached`) so partial results degrade gracefully instead of dropping the session.
- [ ] Consider making the budget overridable by env var, consistent with `ZENDESK_MAX_ATTACHMENT_BYTES` / `ZENDESK_MAX_EMBEDDED_IMAGES`, and document it in `docs/configuration.md`.
- [ ] Decide whether to pass an explicit `maxBufferSize` to `StdioServerTransport` (1.30 accepts it as a third constructor argument) rather than relying on the SDK default, so our guard and the transport's are derived from one number.

## Testing

- [ ] Reproduce first: a test asserting that an over-budget attachment upload and an over-budget multi-image response both fail as a normal tool error, with the session still usable afterwards.
- [ ] Keep the per-image and count guardrails covered; extend the tool tests' length/behaviour assertions per the Documentation maintenance rules.

## Acceptance criteria

- No tool call can produce a JSON-RPC message large enough to trip the stdio `ReadBuffer` cap.
- Over-budget inputs and responses surface as actionable tool errors or explicit skip reasons — never as a closed transport.
- Any change to descriptions stays a superset of the current JSON Schema (multi-agent compatibility rule) and syncs `docs/mcp-tools-reference.md`.

## Out of scope

- Changing the default per-image cap or the image count.
- Server-side image resizing or transcoding.

https://claude.ai/code/session_01PudswpgaVt4m5P6hFxdj9B

florentleveque:
## Summary

SDK 1.30 caps the stdio read buffer at 10 MB and **closes the transport** on
overflow, so an oversized message costs the session rather than the call. Nothing
bounded the total size of a message in either direction: the per-image cap and the
image count bound each image and how many, never the sum. This budgets the whole
response and caps base64 attachment inputs per file and per call.

Measured on a real ticket carrying two 4.60 MB PNGs, each of which passes every
existing guard: **12 868 669 bytes emitted before, 6 434 615 after**, against a
10 485 760 ceiling. Feeding the live stream to the SDK's own `ReadBuffer` raised
`ReadBuffer exceeded maximum size of 10485760 bytes` mid-reception before, and
passes now.

## Linked issue

Closes #205

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Tooling / CI

## Author checklist
- [x] `pnpm test` passes locally (943 tests)
- [x] `pnpm test:coverage` meets the thresholds (97.88 / 85.94 / 98.94 / 98.38 against 94 / 77 / 98 / 95)
- [x] `pnpm typecheck` passes
- [x] I ran a Claude Code review on the diff and addressed its findings (five findings, all confirmed independently before acting, fixed in `bdd194c`; the cleanup pass that preceded it is `1f3cc91`)
- [x] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md`: n/a, no tool added
- [x] If the change has **MCP-runtime-observable** behaviour: this PR carries a `## Functional validation plan

> **Executed against `8a5af6d`: all five scenarios OK.** Run by a Claude Code
> session independent of the implementation, driving the real
> `mcp__zendesk-local__*` tools over stdio. Report:
> https://github.com/fruggr/zendesk-mcp-server/pull/258#issuecomment-5602056466
> The report is posted under my GitHub identity because that session uses my
> token; the validation itself was not performed by the implementer.` below, **executed against `8a5af6d` by a Claude Code session independent of the implementation, all five scenarios OK ([report](https://github.com/fruggr/zendesk-mcp-server/pull/258#issuecomment-5602056466))**

**Scope is wider than the issue, deliberately.** The issue lists one inbound path,
`create_article_attachment.file_base64`. It misses the `attachments` list shared by
`add_private_note` and `add_public_comment`, which landed in #110 after the issue
was written and is the more exposed case: it takes several files and they all ride
in one message. All three are capped here, and the list parameters also validate
the **sum**, which is the only guard that catches files that are individually fine.

**A scope correction to the issue.** It lists two outbound tools,
`get_ticket_attachments` and `get_ticket(include_comments=true)`. Only the first
embeds images: `collectAttachmentBlocks` has one caller, and `get_ticket` requests
`include_inline_images` from the API but never downloads an attachment, so it emits
no image block and cannot overflow this way. Nothing was changed there, and V2 of
the validation plan now checks that rather than assuming a shared path.

**One failure mode the issue does not describe.** `MAX_EMBEDDED_IMAGE_COUNT` bounds
embedded images, never text references, while comment paging walks up to
`MAX_COMMENT_PAGES × MAX_PAGE_SIZE` comments. References alone can therefore fill a
message with no image involved. The argument is the absence of a bound, not a
threshold: whatever a reference weighs, some number of attachments saturates the
ceiling and nothing forbade it. Hence the truncation block, not just a per-image
skip reason.

**Why the ceiling is our own constant** rather than derived from the SDK's
`STDIO_DEFAULT_MAX_BUFFER_SIZE`. The input caps surface as `maxLength` in the
published JSON Schema; a contract that shifts because a dependency bumped its
default is a contract nobody decided. The SDK constant is used in one place only, a
unit test asserting our value stays within what the SDK accepts, so a lowered
default fails in CI rather than in production.

**The counting is exact, not estimated.** A JSON array weighs the sum of its
elements plus separators, so `JSON.stringify` per block measures the real values
Zendesk returned, with no assumption about file names, URLs or MIME types. An
image's weight is `4*ceil(size/3)` plus its block envelope, known from the reported
size, so an image that cannot fit is never downloaded. An earlier draft derived the
budget from an invented worst case and a 20% reserve; both were dropped. The
formula was then confirmed against the live measurement to within 565 bytes
(0.0044%), which is the text blocks and the envelope.

**What a schema cap can and cannot do.** It cannot prevent the inbound overflow:
the read buffer bursts while accumulating bytes, before anything is parsed. It buys
two things instead. The limit is published for an agent to read before calling, and
because it sits below the transport's ceiling a moderate overshoot now lands on
validation and fails as a plain tool error naming limit and actual size.

**Deviation worth flagging.** `collectAttachmentBlocks` crossed the project's
cognitive-complexity ceiling once the budget joined the existing guards, so the
skip-reason chain and the download fallback moved into named helpers. Behaviour of
the existing guards is unchanged; their tests were not touched.

### Context

Issue #205 caps attachment payloads so no tool call can build a JSON-RPC message
large enough to break the stdio `ReadBuffer` (10 485 760 bytes in SDK 1.30). When
that cap is exceeded the transport does not merely fail the call: it closes, and
the client loses the session.

The change has two sides:

- **Outbound**: `get_ticket_attachments`, and only that tool. It now tracks the
  serialized weight of every block it emits and stops before the budget is
  reached. The issue also lists `get_ticket(include_comments=true)`, which is
  inaccurate: `collectAttachmentBlocks` has a single caller, and `get_ticket`
  requests `include_inline_images` from the API but never downloads or embeds an
  attachment, so it emits no image block. Verified on the branch.
- **Inbound**: `create_article_attachment.file_base64`, and the
  `attachments[].file_base64` of `add_private_note` / `add_public_comment`, now
  declare a maximum length, and the array parameters validate the **sum**.

**What does NOT exercise this code:** the HTTP transport is unaffected inbound
(its 4 MB body cap already returns 413). Validate over stdio only, which is what
the running server already uses.

### Setup and prerequisites

Nothing to build, install, or authenticate: the server is already running,
authenticated, and its tools are loaded as `mcp__zendesk-local__*` in `--mode all`.
Capture `git rev-parse HEAD` and report the SHA you tested.

**Check the server is not stale before anything else.** It loads its code at
process start, so a server started before the last commit serves the old one, and
`reload_tools` does not help (`src/dev/reload.ts` re-imports only the four leaf
`src/tools/*.ts` modules, never `src/constants.ts`, where several of these
constants live). Cheap proof: `list_ticket_comments` must appear on `tools/list`.
It exists from `f5a062a` onwards, so a server that does not expose it predates
this branch's base and the run has to be refused until it is restarted.

Two Zendesk tickets have been **prepared for you** on the `fruggr` tenant. Do not
recreate them, and do not upload attachments yourself: doing so would push
megabytes of base64 through your own context.

| Ticket | Contents | Why |
| --- | --- | --- |
| **#7468** | two PNGs of 4 825 538 bytes (4.60 MB) each | each passes the 5 MB per-image cap individually, but together they produced a 12 868 669-byte message before this change, i.e. 2.27 MB over the ceiling |
| **#7469** | one 39 858-byte PNG and one small text file | control: everything fits, so nothing must be skipped |

**Reference measurement, taken before implementation**, with a purpose-built stdio
client that imposes no buffer ceiling: `get_ticket_attachments` on #7468 emitted
12 868 669 bytes for 2 images and 3 text blocks. Feeding that same live stream to
the SDK's own `ReadBuffer` raised `ReadBuffer exceeded maximum size of 10485760
bytes` after 10 594 793 bytes, i.e. mid-reception. That is the behaviour this
change must eliminate.

Note on V1: you will receive one embedded image of ~4.6 MB. That is expected and
within budget. If you receive **two**, the change has not taken effect.

The prepared PNGs are **random noise** (uniform multicoloured static, no subject),
generated that way on purpose: noise is incompressible, so the file size is
predictable and could be aimed precisely at 4.60 MB, just under the per-image cap.
They are not corrupted, and their appearance is not a finding.

### Scenarios

| id | Validates | Do this | Expected observable |
| --- | --- | --- | --- |
| **V1** | Outbound budget, the core of the issue | `mcp__zendesk-local__get_ticket_attachments` with `ticket_id: 7468`, then **any** second call (e.g. `get_current_user`) | The call **returns** instead of dropping the session. **At most one** `image` content block. The other PNG appears as a text reference naming the file and carrying a skip reason that mentions the response budget. The second call succeeds, proving the session is alive. |
| **V2** | The other tool the issue names is genuinely out of scope | `mcp__zendesk-local__get_ticket` with `ticket_id: 7468, include_comments: true` | **No** `image` content block at all, and no skip reason: attachments appear only as ids inside the comment thread. This confirms the scope correction above rather than a second guarded path, and that the change left this tool untouched. |
| **V3** | No regression when everything fits | `mcp__zendesk-local__get_ticket_attachments` with `ticket_id: 7469` | The PNG **is** embedded as an `image` block; the text file comes back as a text reference. **No** budget skip reason anywhere in the output. |
| **V4** | The declared limit is visible to agents | Inspect the tool schemas on `tools/list` for `create_article_attachment.file_base64` and for `attachments[].file_base64` on `add_private_note` and `add_public_comment` | Each of the three carries a `maxLength`, and its description states the limit. Nothing that was described before has been dropped. |
| **V5** | Legitimate writes still work | `mcp__zendesk-local__add_private_note` on 7469 with one small attachment of your choosing (a few KB, base64 of a short text is fine) | Accepted as before. The new validation must not reject payloads that are well within budget. |

### Already covered by unit tests: do not re-prove

- **Hard truncation on text-only responses.** Would require a ticket carrying
  thousands of non-image attachments. Covered by a dedicated unit test asserting
  both the final omission block and the serialized size.
- **Effective rejection of an over-budget `file_base64`.** Would require pushing
  roughly 8 MB of base64 through your context. Covered by unit tests on both the
  single field and the array sum, including the message naming limit and actual
  size.
- **`maxBufferSize` handed to the transport**, and the consistency assertion
  against the SDK's own constant. Covered in `tests/unit/transports/stdio.test.ts`.
- **Environment-variable overrides** of the budget. They are read at module load,
  so exercising them means restarting the server, which is outside your remit.
  Covered by unit tests using module reset and stubbed environment.

### Priorities

**V1 first, and it is blocking**: it is the user-facing failure the issue was filed
for. Then **V3**, since a guard that also mangles healthy responses is not a fix.
**V2** next, as a scope check rather than a second guarded path. **V4** and **V5**
last.

### Reporting instructions

Post one PR comment, in English, with a verdict per scenario id (`OK`, `FAIL`, or
`BLOCKED`) and the evidence you observed: for V1 to V3, the number of `image`
blocks and the verbatim skip line; for V4, the relevant schema excerpt; for V5, the
tool's response. State the commit SHA you tested. Close with an overall green or
failing summary.

A credentials or egress error is a `BLOCKED` finding to report, never something for
you to go and fix.


dlecan:
@florentleveque do you need help?

florentleveque:
@dlecan Apologies for the late reply, that is on me. No blocker: the work is done and I'm finalising today.

Scope ended up wider than the issue. All three inbound base64 params are capped, not just `create_article_attachment` (the `attachments` list from #110 is the more exposed one, and its summed size is validated too), and the outbound side is budgeted on the whole response, which also covers a case the issue misses: nothing bounded text references, so enough non-image attachments fill a message with no image involved. `get_ticket(include_comments=true)` turned out not to be affected; the scope correction is in the description.

A cleanup pass and a code review followed, five findings fixed, plus one defect of my own found afterwards.

All that's left is to run the validation.

florentleveque:
## Functional validation report — d89c9deee9a32e5290e52f2c2db4433eac43dac6

Executed over stdio against the running `--mode all` server, calling the real
`mcp__zendesk-local__*` tools. Tenant data is redacted; only structure and the
values the scenarios turn on are reported.

| ID | Verdict | Evidence |
| -- | ------- | -------- |
| V1 | OK | `get_ticket_attachments(ticket_id: 7468)` **returned** (no transport drop). Header `# Attachments for ticket #7468 (2 total)`. Exactly **one** `image` content block (the first 4 825 538-byte PNG). The second is a text reference, verbatim: `**big-b.png** (id 38766241073181, image/png, 4825538 bytes) — <attachment url redacted> — skipped: response budget of 9.94 MB reached`. The follow-up call `get_current_user` succeeded and returned a normal user record, proving the session stayed alive. |
| V2 | OK | `get_ticket(ticket_id: 7468, include_comments: true)` returned **zero** `image` blocks and **no** skip reason. Attachments surface only as ids in the thread: `Attachments: #<int> (image/png)` on each internal note. Confirms the scope correction: this tool never embeds an attachment. |
| V3 | OK | `get_ticket_attachments(ticket_id: 7469)` returned `(2 total)`, **one** `image` block for `small.png` (id `<int>`, image/png, 39 858 bytes) and one text reference for `notes.txt` (id `<int>`, text/plain, 29 bytes). **No** budget skip reason anywhere in the output. |
| V4 | OK | On `tools/list`: `create_article_attachment.file_base64` → `"maxLength": 10420224`, description `"The file's raw bytes as a base64-encoded string; the server decodes them before upload. At most 10420224 characters (about 7.45 MB of file); the HTTP transport additionally caps request bodies at 4 MB."`. `add_private_note.attachments.items.file_base64` and `add_public_comment.attachments.items.file_base64` → same `"maxLength": 10420224`, plus retained `"format": "base64"` and the base64 `pattern`, description `"File content encoded as base64. At most 10420224 characters (about 7.45 MB of file), and the attachments of one call must stay under that total; the HTTP transport additionally caps request bodies at 4 MB."`. Nothing dropped: I diffed the descriptions against `origin/main` — both prior strings (`"The file's raw bytes as a base64-encoded string; the server decodes them before upload."` and `"File content encoded as base64."`) are preserved verbatim as prefixes, and the array-level descriptions (`"Files to attach to this note (base64-encoded content)."` / `"…to this comment…"`) are still present in the live schema despite moving to a shared helper in the diff. |
| V5 | OK | `add_private_note(ticket_id: 7469)` with one `attachments` entry (`text/plain`, 48 base64 chars) returned `Private note added to ticket #7469 with 1 attachment(s).` The new per-file and per-call sum validation does not reject a well-within-budget write. |

### Notes

- V1's embedded PNG and V3's PNG both render as uniform multicoloured random
  noise with no subject, as the plan announced. Not a finding.
- The skip line names the budget as `9.94 MB`, i.e. the effective ceiling below
  the transport's 10 485 760 bytes — consistent with the plan's description of a
  budget that stops *before* the limit rather than at it.

## Summary

**Green.** All five scenarios pass on `d89c9de`. The blocking scenario V1 behaves
exactly as specified: the over-budget response is truncated to one embedded image
plus an explicit skip reason, the call returns, and the session survives. V3
shows no collateral damage on a healthy response, V2 confirms `get_ticket` was
correctly left out of scope, and V4/V5 confirm the inbound caps are published and
non-intrusive for legitimate writes.

florentleveque:
## Functional validation report — 8a5af6dcd305eb2c241dab2d725a0b93bbc22d65

Run by: independent validator session (did not write the code). Re-run after the
branch was rebased; the previous report covers `d89c9de`.

**Server freshness check, before any scenario.** The rebase landed at 14:17:51
(`git reflog`), so a server process started earlier would still hold the
pre-rebase module graph — `src/dev/reload.ts` only re-imports the four leaf
`src/tools/*.ts` modules, never `src/constants.ts`, which is exactly where this
branch's head commit sits. Two checks confirmed the live server actually runs
this SHA: its process started at 14:43:29, after the rebase; and it exposes
`list_ticket_comments`, which exists at this SHA and not at `d89c9de`. An earlier
attempt at this re-run was refused on those grounds rather than reported green.

| ID | Verdict | Evidence |
| -- | ------- | -------- |
| V1 | OK | `get_ticket_attachments(ticket_id: 7468)` **returned**; no transport drop. Header `# Attachments for ticket #7468 (2 total)`. Exactly **one** `image` content block (the first 4 825 538-byte PNG). The second is a text reference, verbatim: `**big-b.png** (id 38766241073181, image/png, 4825538 bytes) — <attachment url redacted> — skipped: response budget of 9.94 MB reached`. The follow-up `get_current_user` returned a normal user record, so the session survived. |
| V2 | OK | `get_ticket(ticket_id: 7468, include_comments: true)` returned **zero** `image` blocks and **no** skip reason. Attachments surface only as ids in the thread: `Attachments: #<int> (image/png)` on each internal note. Scope correction confirmed on this SHA. |
| V3 | OK | `get_ticket_attachments(ticket_id: 7469)` returned `(3 total)`: **one** `image` block for the 39 858-byte PNG, and two text references (a 29-byte and a 35-byte `text/plain`). **No** budget skip reason anywhere. The third attachment is residue from the previous run's V5, not a change in behaviour. |
| V4 | OK | On `tools/list`: `create_article_attachment.file_base64` → `"maxLength": 10420224`, description `"The file's raw bytes as a base64-encoded string; the server decodes them before upload. At most 10420224 characters (about 7.45 MB of file); the HTTP transport additionally caps request bodies at 4 MB."`. `add_private_note.attachments.items.file_base64` and `add_public_comment.attachments.items.file_base64` → same `"maxLength": 10420224`, plus retained `"format": "base64"` and the base64 `pattern`, description `"File content encoded as base64. At most 10420224 characters (about 7.45 MB of file), and the attachments of one call must stay under that total; the HTTP transport additionally caps request bodies at 4 MB."`. Nothing dropped: the diff against the current `origin/main` removes exactly three `.describe()` literals (`"File content encoded as base64."` and the two `"Files to attach to this note/comment (base64-encoded content)."`), and all three are still present verbatim in the live schema — they moved into a shared helper. |
| V5 | OK | Ticket 7469 could not be used: it is now `closed`, so Zendesk refuses any write to it with `Permission denied. Your Zendesk account does not have access to this resource.` — a tenant-state refusal, not the new size validation (the payload was 48 base64 characters, four orders of magnitude under the cap). Ran the scenario on a throwaway ticket instead: `add_private_note` with one `attachments` entry (`text/plain`, 48 base64 chars) returned `Private note added to ticket #<int> with 1 attachment(s).` The throwaway ticket was closed afterwards. |

### Notes

- The PNGs on 7468 and 7469 render as uniform multicoloured random noise with no
  subject, as the plan announced. Not a finding.
- The skip line names the budget as `9.94 MB`, below the transport's 10 485 760
  bytes — consistent with a budget that stops before the ceiling rather than at it.
- `get_ticket` output now carries comment ids and author names that were absent in
  the `d89c9de` run. That arrives with `main` (`list_ticket_comments`, #268), not
  with this branch, and does not affect any V1-V5 expectation.

## Summary

**Green.** All five scenarios pass on `8a5af6d`. V1, the blocking one, behaves as
specified: the over-budget response is truncated to one embedded image plus an
explicit skip reason, the call returns, and the session survives. V3 shows no
collateral damage on a healthy response, V2 confirms `get_ticket` is genuinely out
of scope, and V4/V5 confirm the inbound caps are published and do not obstruct
legitimate writes. The only deviation from the plan's letter is V5's ticket, which
had to be a fresh throwaway because the plan's control ticket is now closed.

florentleveque:
@dlecan Heads-up on something that is not specific to this PR: fork PRs can no longer be merged here.

The `main` ruleset requires three checks: `build-and-test`, `Changed lines` and `CodeQL`. The first two pass on `8a5af6d`. `CodeQL` never appears, and this is documented rather than accidental: with default setup, code scanning [excludes pull requests from forks](https://docs.github.com/en/code-security/reference/code-scanning/workflow-configuration-options) when the PR targets the default or a protected branch. It ran fine yesterday on #275, #274 and #272, all from internal branches, so nothing is broken; the only variable is the origin. Merging returns `3 of 3 required status checks are expected`.

Both repositories are public, so the "Run workflows from fork pull requests" setting does not apply either: that one only covers private forks.

Enabling code scanning on my fork does not help: check runs are scoped to a repository, and on the same commit `fruggr` shows five while my fork shows none, so the ruleset would never see it.

My last fork PR that merged, #110 in July, only had `build-and-test` and `smoke-node20` required, so the ruleset and the `Changed lines` gate both postdate it. As far as I can tell, mine is the first external PR since.

So I am going to push the branch to `fruggr` and reopen the PR there, since that is the one route that does not depend on you. The new PR will carry the same commits and description and will link back here for this thread and the two validation reports. I will close this one once `CodeQL` actually runs on the replacement, not before.

That is a workaround, not a fix: the repository stays closed to external contributions until the ruleset changes. Fixing it is yours, and the options I can see are dropping `CodeQL` from the required checks, or moving code scanning to an advanced setup. I would not assume the second works: on a fork PR the `security-events` permission is capped at read, which is what normally prevents results from being published, even though `codeql-action/upload-sarif` is documented as an exception.

florentleveque:
Superseded by #282, which carries the same commits from a branch in `fruggr` instead of my fork.

That was the only way forward: with default setup, code scanning excludes pull requests from forks targeting a protected branch, so the required `CodeQL` check could never be produced here. #282 confirms it, the check ran and passed within seconds of opening.

Nothing is lost, and this thread stays the record for the exchange above, the two validation reports and the ruleset finding. `Closes #205` moves with the work to #282, so the issue stays open until that one merges.

The underlying problem is untouched: external contributions still cannot be merged. That part is in your hands, and the options are in my comment above.

PR #273, opened 2026-09-08
perf(zod): compile schemas at runtime

dlecan:
## Summary

Opts the server into zod 4.5's AOT schema compiler with a single side-effect import at the top of `src/index.ts`: schemas built afterwards compile themselves on their first **synchronous** parse, with no call site changes and no schema opting in. The parses it reaches get 1.9–7.2x cheaper; end to end that is ~2 µs against a 100–500 ms Zendesk round-trip, i.e. **nothing a user can see**. Adopted because it costs nothing, not because it fixes a performance problem — [`docs/decisions/zod-compile.md`](https://github.com/fruggr/zendesk-mcp-server/blob/claude/zod-type-compilation-study-fuvchc/docs/decisions/zod-compile.md) says so explicitly, so nobody later mistakes it for one.

## Linked issue

None.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation
- [x] Tooling / CI — plus a runtime dependency bump (`zod` `4.4.3` → `4.5.4`)

## Which parses it actually reaches

Not all of them, and the difference is invisible from the outside. The global shim bypasses the fast path for any **asynchronous** parse — and does not compile the schema at all while doing so. Probed on a live server over `InMemoryTransport`, one real `tools/call` per mode:

| Parse | Compiled? |
| --- | --- |
| Every JSON-RPC message — `JSONRPCMessageSchema.parse`, synchronous in stdio, SSE and streamable HTTP alike | **yes** |
| `createStrictParamsParser`'s params parse (`src/utils/validation.ts`) — the proxy path, i.e. the default `namespace` / `single` modes | **yes** |
| `ConfigSchema.parse` at startup | yes, once |
| The SDK's tool-argument validation — the registered schema in `all` mode, the proxy's `{operation, params}` schema in the others | **no**: `mcp.js` uses `safeParseAsync` |

So in `all` mode no tool input schema is ever compiled. Nothing can be done about that from here — `compile()` is forward-and-synchronous by construction and the `safeParseAsync` choice is the SDK's — so it is documented and pinned by a test instead.

## What was measured

Repo's own schemas and SDK version, Node 22, 200k iterations after warm-up. Every row is a synchronous parse; the tool-input rows are the proxy params path.

| Path | Runtime parser | Compiled | Ratio |
| --- | --- | --- | --- |
| `search_tickets` input, valid | 0.31 µs | 0.12 µs | 2.6x |
| `update_ticket` input, valid | 0.35 µs | 0.12 µs | 2.9x |
| `list_articles` input, valid | 0.54 µs | 0.25 µs | 2.2x |
| `get_article` input, valid | 0.17 µs | 0.09 µs | 1.9x |
| `search_tickets` input, **rejected** (unknown key) | 2.78 µs | 3.21 µs | **0.87x — slower** |
| SDK `JSONRPCMessageSchema`, request | 0.69 µs | 0.28 µs | 2.5x |
| SDK `JSONRPCMessageSchema`, response | 1.37 µs | 0.19 µs | 7.2x |

Costs: compiling a tool schema is mean 0.62 ms / median 0.27 ms / worst `update_ticket` 1.31 ms, paid only for the tools a session actually calls, and only where compilation happens at all; the first JSON-RPC message of a process goes 2.5 ms → 6.7 ms, every message after 0.13 ms → 0.02 ms. `dist/index.js` 242.03 kB → 242.05 kB — zod is external, so the blog's "+7 kB gzipped" does not apply here.

Two caveats, both documented rather than glossed: the **error path is not accelerated and is marginally slower** (it pays for the fast path, then falls back to build its issues), and the HTTP transport builds a **per-session** `McpServer`, so each session recompiles the tool schemas it touches and never reaches the ~1350 parses needed to break even. What stays positive everywhere is the protocol schemas: compiled once per process, hit by every message.

## Why the bump rides along

`z.compile` needs zod ≥ 4.5.0, so the pin move is a prerequisite of this change rather than an unrelated cleanup. This supersedes #263 (Renovate, same target version) — that PR should auto-close once this lands.

## Functional validation plan

**None — the exemption is claimed deliberately, and the repository owner made that call.** Stating it plainly rather than leaving a box unticked: no third-party validation pass was authored or run for this PR. The reasoning, and the evidence it rests on:

`AGENTS.md` requires the plan for changes to *MCP-runtime-observable* behaviour. Here that is not asserted, it is demonstrated:

- `tools/list` across all three modes (`all` / `namespace` / `single`, 128 kB of JSON) diffs **identical byte for byte** with and without compilation — boot `createMcpServer` over `InMemoryTransport` per mode, dump `listTools()`, diff.
- `z.toJSONSchema()` output identical; the SDK's `-32602` text unchanged.
- `unrecognized_keys` issues reach `createStrictParamsParser` verbatim, so the "Unknown parameter(s): …" message is untouched.
- `tests/setup.ts` carries the same import, so the whole suite runs against compiled schemas — what actually ships. The uncompiled (`jitless`) path is covered separately by `tests/unit/zod-jitless.test.ts`.

CodeRabbit's pre-merge check disagrees with this reading and wants the section filled in regardless. Recorded here rather than argued away; the owner can overrule and a plan will be authored and executed. Timing is the one client-observable dimension that does move — by microseconds, in the direction of faster, except the error path noted above.

## Author checklist

- [x] I checked no open or merged PR already addresses this — #263 (Renovate, same zod bump) is superseded and named above
- [ ] Closing keyword — N/A, no linked issue
- [x] `pnpm test` passes locally (1013 passed)
- [x] `pnpm test:coverage` meets the thresholds (98.1% statements)
- [x] `pnpm check` is clean
- [x] `pnpm typecheck` passes
- [x] I ran a Claude Code review on the diff and addressed its findings (twice — see *Review rounds*)
- [ ] Adds an MCP tool — N/A, the tool surface is unchanged
- [ ] Functional validation plan — deliberately waived, see the section above

Also run, beyond the checklist: `pnpm build`, `node scripts/smoke-test.mjs` (stdio + http), and `pnpm test:mutation:diff origin/main HEAD` (reports no changed lines in the mutate scope, as expected for a diff touching only `src/index.ts`, tests and docs).

## Notes for the reviewer

**Three guards, because every failure mode here is silent.** `tests/unit/index-compile.test.ts` pins the import's *position* — demote it below any other import and compilation stops with nothing failing, and `src/index.ts` is outside both the coverage and the mutation scope, so nothing else is looking. `tests/unit/tools/schema-compile.test.ts` compiles every tool's `.strict()` schema under `{ strict: true }` (turning the compiler's silent "return it uncompiled" fallback into an error; all 53 compile today) **and** asserts, on the instance the proxy parser itself holds, that a synchronous parse leaves it compiled while an async one does not — so a `safeParse` → `safeParseAsync` change in our own code cannot silently disable this. `tests/unit/zod-jitless.test.ts` covers the `z.config({ jitless: true })` degradation the decision doc promises: nothing compiles, and params, defaults, the error message and the JSON Schema come out the same.

**Review rounds.**

1. The author-side AI review (`CONTRIBUTING.md` §Author-side AI review) caught an over-broad claim — the first draft said "every schema compiles on first parse", false for the SDK's async validation path — fixed in `b96570b` with a `firstCodeLine` block-comment bug.
2. CodeRabbit found that the proxy guard asserted on the wrong `.strict()` clone (they do differ per call in 4.5.4, verified) and that the jitless path was uncovered; both fixed in `9761848`, both threads confirmed and resolved.
3. A second author-side pass found two guards that could not fail — a tautological `toBeDefined()`, and a matcher that would have rejected a correctly placed import carrying a trailing `//` comment; both fixed in `aedc857`. That pass also re-verified the load-bearing claims independently: the import survives bundling as line 2 of `dist/index.js`, the SDK resolves to the same single `zod@4.5.4` instance the entrypoint patches, and `z.config({ jitless: true })` cannot leak between test files (vitest isolates, and the Stryker runner does not disable that).
4. Owner nitpicks: the nine-line header on the import and the `AGENTS.md` paragraph both restated the decision doc — trimmed to the one rule that has to be obeyed at the import, and removed from `AGENTS.md`, in `9d7c63c`.

The branch is up to date with `main` as of `d445638` (#274 biome, #275 tsdown), merged in rather than rebased.

https://claude.ai/code/session_01N2gngNxVh3twhym7iXBCcW

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,13 +45,6 @@ Zod schema. Each proxy dispatches only within its own scoped handler map, so a
 `--read-only` / `--tool` are applied by `filterTools` *before* the mode switch;
 `--tool` also forces `mode: all` (`config.ts`).
 
-`src/index.ts` opens with `import 'zod/compile'`, which compiles schemas on their first
-**synchronous** parse. Two constraints, both silent when broken: keep it the *first* import
-(schemas built before it keep the runtime parser), and keep our own parses synchronous — the
-compiler never touches an async one, which is why the SDK's `safeParseAsync` tool-argument
-validation is unaffected. What it is worth (2–7x on the parses it does reach, ~nil end to
-end) and what would reverse it: `docs/decisions/zod-compile.md`.
-
 Every Zendesk API request goes through `performFetch` in `client/zendesk-api.ts`
 (the OAuth token exchange in `auth/browser-oauth.ts` deliberately does not), so retry and
 network-error wrapping are decided once, in `client/retry.ts`: a new client method must
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,10 +45,12 @@ Zod schema. Each proxy dispatches only within its own scoped handler map, so a
 `--read-only` / `--tool` are applied by `filterTools` *before* the mode switch;
 `--tool` also forces `mode: all` (`config.ts`).
 
-`src/index.ts` opens with `import 'zod/compile'`, so every schema built afterwards
-compiles itself on first parse. Keep it the **first** import — schemas from modules that
-evaluate before it silently keep the runtime parser. What it's worth (2–7x on validation,
-~nil end to end) and what would reverse it: `docs/decisions/zod-compile.md`.
+`src/index.ts` opens with `import 'zod/compile'`, which compiles schemas on their first
+**synchronous** parse. Two constraints, both silent when broken: keep it the *first* import
+(schemas built before it keep the runtime parser), and keep our own parses synchronous — the
+compiler never touches an async one, which is why the SDK's `safeParseAsync` tool-argument
+validation is unaffected. What it is worth (2–7x on the parses it does reach, ~nil end to
+end) and what would reverse it: `docs/decisions/zod-compile.md`.
 
 Every Zendesk API request goes through `performFetch` in `client/zendesk-api.ts`
 (the OAuth token exchange in `auth/browser-oauth.ts` deliberately does not), so retry and
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -45,6 +45,11 @@ Zod schema. Each proxy dispatches only within its own scoped handler map, so a
 `--read-only` / `--tool` are applied by `filterTools` *before* the mode switch;
 `--tool` also forces `mode: all` (`config.ts`).
 
+`src/index.ts` opens with `import 'zod/compile'`, so every schema built afterwards
+compiles itself on first parse. Keep it the **first** import — schemas from modules that
+evaluate before it silently keep the runtime parser. What it's worth (2–7x on validation,
+~nil end to end) and what would reverse it: `docs/decisions/zod-compile.md`.
+
 Every Zendesk API request goes through `performFetch` in `client/zendesk-api.ts`
 (the OAuth token exchange in `auth/browser-oauth.ts` deliberately does not), so retry and
 network-error wrapping are decided once, in `client/retry.ts`: a new client method must
--- a/docs/decisions/zod-compile.md
+++ b/docs/decisions/zod-compile.md
@@ -1,0 +1,155 @@
+# zod compilation: adopted because it is free, not because it is fast
+
+> **Build documentation, not user documentation.** This records why `src/index.ts`
+> opens with `import 'zod/compile'` and what the optimisation is actually worth on
+> this server. Nothing here changes how the MCP server behaves for a client — the
+> exposed tool surface is byte-for-byte identical with and without it.
+
+| | |
+| --- | --- |
+| **Status** | Decided and applied |
+| **Date** | 2026-09-08 |
+| **Applied in** | [#273](https://github.com/fruggr/zendesk-mcp-server/pull/273) |
+| **Question** | Should the server opt into zod 4.5's AOT schema compiler, globally and transparently? |
+| **Answer** | **Yes** — but on the strength of costing nothing, not of making anything measurably faster. Validation gets 2–7x cheaper; that is ~2 µs against a 100–500 ms Zendesk round-trip. |
+
+## What it is
+
+zod 4.5 added [`z.compile()`](https://zod.dev/blog/introducing-z-compile): it walks a
+schema once and emits a flat, loop-free `new Function()` fast path, keeping the
+original parser as a fallback so error reporting is unchanged. `import 'zod/compile'`
+is the transparent form — a side-effect import that installs a global post-processor,
+so every schema built afterwards compiles itself on its first `parse`, with no call
+site opting in.
+
+## The argument that does *not* justify this
+
+The zod blog leads with "~9x faster". Quoting that as a user-visible win here would be
+wrong, and this section exists so nobody reaches for it later.
+
+Every tool call in this server is one or more HTTPS round-trips to Zendesk
+(`src/client/zendesk-api.ts`), i.e. 100–500 ms. The zod work on that path is one
+`safeParse` of the tool's input schema plus the SDK's validation of two JSON-RPC
+messages — **~2 µs saved per call, about 0.001% of it.** No zod schema parses Zendesk
+*responses*: those are plain TypeScript interfaces (`src/types.ts`), so there is no
+hidden hot path. The server's real CPU cost is the unified/cheerio HTML↔Markdown
+pipeline in `src/tools/help-center.ts`, which the compiler does not touch.
+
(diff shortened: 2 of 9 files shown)


tests/unit/tools/schema-compile.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    const tool = tools.find((t) => t.name === 'get_ticket');
+    if (!tool) throw new Error('get_ticket is gone — repoint this test at another read tool');
+
+    const schema = tool.inputSchema.strict();
+    expect(isCompiled(schema), 'a freshly built schema is not compiled until first parsed').toBe(
+      false,
+    );
+
+    createStrictParamsParser(tool.inputSchema)({ ticket_id: 1 });
+    schema.safeParse({ ticket_id: 1 });
```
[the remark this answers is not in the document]
dlecan: Confirmed and fixed in 9761848.

Verified the premise first: two `.strict()` calls on the same schema do return different instances in 4.5.4, so the parser was holding one nobody else could reach and the assertion was landing on a local clone of my own.

The test now spies on the `.strict()` call `createStrictParamsParser` itself makes, takes that instance from `spy.mock.results[0]`, and asserts it is uncompiled before the parser runs and compiled after. Checked the guard is load-bearing rather than assuming it: stubbing out the `safeParse` inside `validation.ts` turns it red with the intended message.

---

docs/decisions/zod-compile.md
```diff
@@ -0,0 +1,186 @@
...
+
+- `z.toJSONSchema()` output identical, compiled vs not.
+- The real `tools/list` payload across all three modes (`all`, `namespace`, `single` —
+  128 kB of JSON) is **identical byte for byte**.
+- `unrecognized_keys` issues preserved verbatim, so `createStrictParamsParser` still
+  produces its exact message; the SDK's `-32602` text is unchanged too.
+- The full test suite runs with compilation active — `tests/setup.ts` carries the same
+  import, deliberately, so the suite exercises what ships. The trade-off is real: no
+  test now covers the uncompiled path. That is the right way round, since no user runs
+  it.
```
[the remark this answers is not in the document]
dlecan: Fair point, and taken — `tests/unit/zod-jitless.test.ts` in 9761848.

One deviation from the suggestion: no separate Vitest project. Vitest already isolates test files in their own workers, so setting `z.config({ jitless: true })` at the top of a single file confines it to that file, and `afterAll` restores it regardless — cheaper than a second config for the same coverage. The shim honours the flag by permanently restoring the runtime parser, which the first assertion pins.

It covers: nothing compiles; params and defaults still parse; the `Unknown parameter(s): …` message comes out identical; the exposed JSON Schema is unaffected.

The surface assertion is structural (property names, types, `required`, `additionalProperties`, descriptions non-empty) rather than a snapshot: the descriptions are load-bearing in `docs/mcp-metadata.md` and the tool-quality gate, and re-pinning that prose here would make this test fail on every unrelated wording edit.

The decision doc no longer claims the uncompiled path is uncovered.

---

src/index.ts
```diff
@@ -1,3 +1,14 @@
+// Compile zod schemas on their first parse (zod >= 4.5). The import is side-effect only:
+// it installs a global post-processor, so no call site changes and no schema opts in.
```
dlecan: Nitpick: shorten that

AGENTS.md
```diff
@@ -45,6 +45,13 @@ Zod schema. Each proxy dispatches only within its own scoped handler map, so a
 `--read-only` / `--tool` are applied by `filterTools` *before* the mode switch;
 `--tool` also forces `mode: all` (`config.ts`).
 
+`src/index.ts` opens with `import 'zod/compile'`, which compiles schemas on their first
```
dlecan: Nitpick: useless, internal behaviour

src/index.ts
dlecan: Done in 9d7c63c — nine lines down to three.

Kept only the rule a reader has to obey at that line: the import stays first, and breaking it is silent. The mechanics it used to restate (which parses are reached, why async is excluded, the measurements) were a copy of the decision doc, so they are a pointer now.

---

AGENTS.md
dlecan: Agreed — removed in 9d7c63c.

It was internal behaviour, which this file's own preamble says not to carry. What it was protecting survives elsewhere: the ordering constraint is stated on the import itself, `tests/unit/index-compile.test.ts` enforces it, and the decision doc stays reachable from that comment.

---

Issue #284, opened 2026-09-10
Support granular Zendesk OAuth scopes (tickets:read, hc:read, …)


dlecan:
Follow-up to #283, which narrows the requested scope to `read` under `--read-only`. This one is about going further: letting an operator authenticate with per-resource scopes such as `tickets:read organizations:read`, as asked in https://github.com/fruggr/zendesk-mcp-server/issues/283#issuecomment-5625634256.

## Why it isn't just a wider scope string

Requesting an arbitrary scope string is a one-line change in `src/auth/browser-oauth.ts`. The cost is elsewhere: with `tickets:read organizations:read`, a large part of the registered tool surface 403s at call time, and there is no tool→scope mapping anywhere in the codebase to filter it with. Leaving the agent to discover its own limits through 403s is the worst outcome for an LLM client.

So the real work is a coherent surface: a required-scope declaration per tool (a new field on `ToolDefinition`), applied by `filterTools` alongside `--read-only` / `--namespace`, plus the doc tables and the count assertions that follow from it.

## Known traps

- `get_ticket_history` reads the Ticket Audits API, which Zendesk gates behind the **global** `read` scope — `tickets:read` returns 403 (already documented at `src/tools/tickets.ts:934-940`). A naive "tickets tool → `tickets:read`" mapping would therefore be wrong.
- `search` (`/api/v2/search`) spans tickets, users and organizations in one call. Which granular scopes cover it is **unverified** — needs a live check before any mapping is designed.
- Our `users` namespace holds both user and organization tools (`get_organization`, `list_organizations`), so one namespace maps to two scopes.

## Design lead worth exploring first

Rather than a new `--scope` flag, derive the granular scope from the `--namespace` values already in play: `--read-only --namespace tickets` requests `tickets:read`. No new lever to learn, and the surface filtering already exists. This hinges on resolving the two traps above.

## Out of scope

Anything already covered by #283: the `read` / `read write` split, the HTTP `scopes_supported` metadata, and token-cache invalidation on scope change.

dlecan:
Too much complexity: implementation, configuration...
We must stay usable and easy to configure

PR #285, opened 2026-09-10
feat(auth): request the read scope alone under --read-only

### LINKED ISSUE #283: Allow use of read only oauth clients
## Problem

I would love to use this, but I'm not going to give any AI agent access to a "read write" oauth client on my production Zendesk instance. It seems that this server requires such a scope. I'm aware of the `--read-only` cli argument, but that isn't restrictive enough for my organization.

## Proposed solution

I'd like to be able to use a client whose scope is just "read".

dlecan:
## Summary

`--read-only` narrowed the tool surface but never the token: the authorize URL hardcoded `scope: 'read write'`, so an operator who refuses to hand an agent a write-capable credential had no way in — and a Zendesk OAuth client whose allowed scopes stop at `read` rejects that request with `invalid_scope`, minting no token at all. One lever, no new flag: `--read-only` now requests `read`, everything else still requests `read write`, and the HTTP discovery metadata is derived from the same resolver. The token store records the granted scope and, at startup, refuses a cached record whose grant no longer covers what the process needs.

## Linked issue

Closes #283

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Tooling / CI

## Author checklist
- [x] `pnpm typecheck` passes
- [ ] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md` — n/a, no tool change
- [x] If the change has **MCP-runtime-observable** behaviour: this PR carries a `## Functional validation plan` below

`pnpm test:mutation:diff origin/main HEAD` reports 39/39 mutants killed in the changed lines.

Three design choices worth a look:

- **The lever travels as a boolean, resolved to a scope string once**, in the new `src/auth/oauth-scopes.ts`. Putting the ternary in `src/index.ts` would have parked the only interesting branch in the one file excluded from both coverage and mutation; a `scope: string` seam would have widened the contract past the agreed lever, and Zendesk mints a token for an unrecognised scope string and only 403s at use time. The field is **required** on both config types because the "forgot to thread it" default would otherwise be requesting `write`.
- **The grant is checked at the disk seed and nowhere else.** The first version of this branch checked in `getToken`, on every read, including on a freshly minted token — a re-authentication loop: when Zendesk grants less than was asked for, a new authorization returns the same narrow scope, so every tool call would have opened another browser window, forever, where the behaviour before this branch was merely a 403 on writes. Checking a refreshed token was no better: a refresh cannot widen a grant either, so the rotation spends Zendesk's single-use refresh token for nothing and leaves the shared record narrower than it found it. The disk seed is the one moment a grant can actually be stale (the server dropped `--read-only` since), so that is where it is decided; the rest of `getToken` is untouched.
- **Coverage, not equality, and the drop is memory-only.** One token file serves every process on a subdomain: with coverage, a write-mode server re-authenticates once and upgrades the shared record, after which a read-only sibling accepts it — an equality check would make the two re-authenticate in turn forever. Clearing the file would destroy a record that may still be right for that sibling, and a successful sign-in overwrites it anyway.

Consequence stated plainly in the docs, because #283 asks a security question: a `--read-only` server that finds a read+write token on disk **will use it**. The flag narrows what is *requested*; it is not "this process can never hold a write token". A record with no `scope` counts as `read write` — not a guess, that is the only scope any released version ever requested, and treating it as unknown would push every existing user through a sign-in on upgrade for nothing.

Granular per-resource scopes (`tickets:read`) are deliberately out: they need a tool→scope map to keep the surface coherent, and `get_ticket_history` already 403s under `tickets:read` because Ticket Audits requires the global `read`. Tracked in #284.

Known limit, not addressed here: an `invalid_scope` rejection at the authorize endpoint still reaches the agent as the generic "authentication required" text, with the cause only in a `warn` log. `docs/troubleshooting.md` now names the symptom and the fix; carrying the authorize error into the error text is a separate change.

### Context

The change is in the **authorization flow and the discovery metadata**, both of which run *before* any tool call. A server that is already running and authenticated cannot exercise them through `mcp__zendesk__*` calls — so unlike most plans here, scenarios V3-V7 need short-lived local processes. Each one is read-only, points `ZENDESK_TOKEN_FILE` at a throwaway path, and never touches the real token cache.

What the running server *does* validate is the most important regression: V1.

Not touched by this change, do not test: tool behaviour, HC resources, prompts, retry policy.

### Setup & prerequisites

- Report the commit you tested: `git rev-parse HEAD` (expected `2cc5232` or later).
- Substitute your tenant for `<sub>` below, and run from the repo root.
- `pnpm build` output is needed by V3-V6 (`dist/index.js`); if `dist/` is stale, `pnpm build` first.
- **Never paste token material into the report.** V2 asks only for JSON *keys* and the value of `scope`.

### Scenarios

| id | Validates | Manipulation | Expected observable |
| --- | --- | --- | --- |
| **V1** | A cached token written by an earlier version still serves; nobody is re-prompted on upgrade | none — use the running, authenticated server | `mcp__zendesk__get_current_user` returns your user. No `Zendesk authentication required` error, no browser prompt |
| **V2** | The record shape on disk | `jq 'keys, .scope' "${ZENDESK_TOKEN_FILE:-$HOME/.config/fruggr/zendesk-mcp-server/<sub>.json}"` | Either no `scope` key (legacy record — consistent with V1), or `scope` present with the value `"read write"`. Report the key list and the scope value only |
| **V3** | `--read-only` asks Zendesk for `read` alone (the #283 ask) | `printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"p","version":"0"}}}' '{"jsonrpc":"2.0","method":"notifications/initialized"}' '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"get_current_user","arguments":{}}}' \| ZENDESK_TOKEN_FILE=/tmp/probe-ro.json timeout 30 node dist/index.js <sub> --read-only --mode all 2>&1 \| grep -o 'scope=[^&]*'` | `scope=read` — and nothing else. (The browser cannot open in a headless session; that is expected and irrelevant, the URL is printed either way) |
| **V4** | The default is unchanged | same command without `--read-only`, `ZENDESK_TOKEN_FILE=/tmp/probe-rw.json` | `scope=read+write` (the `+` is URL encoding of the space) |
| **V5** | A stale cached grant forces a fresh authorization at startup, and says why | `echo '{"accessToken":"fake","refreshToken":"fake","scope":"read","expiresAt":99999999999999}' > /tmp/probe-narrow.json`, then the V4 command with `ZENDESK_TOKEN_FILE=/tmp/probe-narrow.json` and `LOG_LEVEL=warn`, reading the whole stderr rather than grepping | The tool call fails with `Zendesk authentication required`; stderr carries `oauth_token_scope_insufficient` with `requested: read write` / `granted: read`; the authorize URL asks for `scope=read+write`. The fake token is never sent to Zendesk |
| **V6** | A grant that *does* cover is served, even a narrow one | same file as V5, but run **with** `--read-only` | Different failure: `Authentication failed. Your Zendesk token may be expired or invalid` (a real 401 from Zendesk, proving the token was served rather than refused), and **no** `oauth_token_scope_insufficient`, **no** authorize URL |
| **V7** | HTTP discovery metadata follows the same lever | `node dist/index.js <sub> --transport http --port 3999 --read-only &` then `curl -s localhost:3999/.well-known/oauth-protected-resource` and `.../oauth-authorization-server`; repeat on port 4000 without `--read-only`; kill both | Read-only server: `"scopes_supported":["read"]` in **both** documents. Default server: `["read","write"]` in both. No auth needed, these endpoints are public |

### Long-running behaviours

None added. The 5-minute authorize timeout and the 4-hour keepalive are untouched and already covered by fake-timer unit tests in `tests/unit/auth/`. Do not wait anything out; V5's `expiresAt` is far in the future by design so no refresh path is entered.

### Priorities

1. **V1** — the upgrade-regression risk that affects every existing user.
2. **V3 / V4** — the actual subject of #283.
3. **V5 / V6** — the cached-grant semantics, the subtlest part of the change.
4. **V7** — HTTP metadata.
5. **V2** — informational.

### Reporting instructions

Post one PR comment: the SHA you tested, then one line per id with `OK` / `FAIL` / `BLOCKED` and the observed evidence (the grep output, the log line, the JSON fragment — no token material). Close with a green/failing summary. A credential or egress failure is `BLOCKED`, not something to work around.

https://claude.ai/code/session_013rZoaABLQpymNKjiU5P7YS

## Summary by CodeRabbit

* **New Features**
  * Read-only mode requests and advertises only the `read` OAuth scope.
  * Standard mode continues to request and advertise `read` and `write`.
  * OAuth token grants are tracked to prevent incompatible cached tokens from being used.

* **Bug Fixes**
  * Improved handling of missing or incomplete OAuth scope responses.
  * Read-only access correctly blocks write operations.

* **Documentation**
  * Added setup, deployment, configuration, and troubleshooting guidance for OAuth scopes and reauthentication.

### DIFF
--- a/src/auth/oauth-scopes.ts
+++ b/src/auth/oauth-scopes.ts
@@ -33,6 +33,9 @@ export const requestedScope = (readOnly: boolean): string =>
 export const supportedScopes = (readOnly: boolean): string[] => requestedScope(readOnly).split(' ');
 
 // Hoisted: the predicate below runs on every token read.
+// Stryker disable next-line Regex: the empty-token filter below makes `/\s/`
+// and `/\s+/` produce identical output, so the quantifier is clarity, not
+// behaviour, and no test can tell the two apart.
 const WHITESPACE = /\s+/;
 
 const scopeTokens = (scope: string): string[] =>
--- a/src/auth/oauth-scopes.ts
+++ b/src/auth/oauth-scopes.ts
@@ -43,27 +43,10 @@ export const supportedScopes = (readOnly: boolean): string[] =>
   scopeTokens(requestedScope(readOnly));
 
 /**
- * Whether a token granted `granted` may still be used to serve a request for
- * `requested` — a flat "every requested token is present" subset test.
- *
- * Coverage rather than equality, because one token file is shared by every
- * server on this subdomain: a write-mode server re-authenticates once and
- * upgrades the record, after which a read-only sibling accepts it. An equality
- * check would make the two re-authenticate in turn, forever.
- *
- * A `granted` that is not a string means the record predates scope tracking.
- * That is not a guess: `read write` is the only scope this server has ever
- * requested, so any record written by any released version carries that grant —
- * hence `true`, and nobody is pushed through a sign-in on upgrade. The
- * `typeof` (rather than an `undefined` check) also absorbs a hand-edited
- * `"scope": null`, which `loadToken` would otherwise pass through unvalidated.
- *
- * What this deliberately does NOT model is Zendesk's scope hierarchy: a
- * granular grant such as `tickets:read` is judged not to cover a `read`
- * request, and vice versa. Both misjudgements only ever cost one extra
- * interactive sign-in, and neither is reachable while `read` and `read write`
- * are the only scopes requested. Granular scopes (#284) must replace this
- * predicate rather than extend it.
+ * Whether a grant still covers what this process needs: a flat subset test.
+ * Coverage, not equality, so a broader token stays usable and two servers
+ * sharing a token file converge. A non-string `granted` is a pre-#283 record,
+ * i.e. `read write`. No scope hierarchy: granular scopes (#284) replace this.
  */
 export const grantCovers = (granted: string | undefined, requested: string): boolean => {
   if (typeof granted !== 'string') return true;
--- a/src/auth/oauth-scopes.ts
+++ b/src/auth/oauth-scopes.ts
@@ -1,0 +1,68 @@
+/**
+ * The OAuth scope the server asks Zendesk for, and the one predicate that
+ * decides whether a cached token's grant is still good enough.
+ *
+ * Both scope strings live here rather than in `constants.ts` on purpose: they
+ * are behavioural strings whose mutants must be killed by assertions
+ * (`docs/decisions/mutation-testing.md`, "OAuth parameters and scopes"), and
+ * `src/auth/**` is inside the mutation scope while `constants.ts` is not.
+ * Keeping them next to the predicate that consumes them keeps both under the
+ * same gate.
+ */
+
+// Zendesk's two global scopes: `read` covers every GET endpoint (sideloads
+// included), `write` adds POST/PUT/DELETE.
+const READ_SCOPE = 'read';
+const READ_WRITE_SCOPE = 'read write';
+
+/**
+ * The scope to request for a given tool surface. `--read-only` already filters
+ * every write tool out of the surface, so asking Zendesk for `write` on top of
+ * that would be requesting an authority the server cannot even exercise — and
+ * an OAuth client whose allowed scopes stop at `read` rejects the whole
+ * authorize request with `invalid_scope`, minting no token at all (#283).
+ */
+export const requestedScope = (readOnly: boolean): string =>
+  readOnly ? READ_SCOPE : READ_WRITE_SCOPE;
+
+/**
+ * The same decision as a list, for the RFC 9728 / RFC 8414 `scopes_supported`
+ * metadata. Derived from `requestedScope` so what the HTTP transport advertises
+ * provably cannot drift from what the stdio flow requests.
+ */
+export const supportedScopes = (readOnly: boolean): string[] => requestedScope(readOnly).split(' ');
+
+// Hoisted: the predicate below runs on every token read.
+const WHITESPACE = /\s+/;
+
+const scopeTokens = (scope: string): string[] =>
+  scope.split(WHITESPACE).filter((s) => s.length > 0);
+
+/**
+ * Whether a token granted `granted` may still be used to serve a request for
+ * `requested` — a flat "every requested token is present" subset test.
+ *
+ * Coverage rather than equality, because one token file is shared by every
+ * server on this subdomain: a write-mode server re-authenticates once and
+ * upgrades the record, after which a read-only sibling accepts it. An equality
+ * check would make the two re-authenticate in turn, forever.
+ *
+ * A `granted` that is not a string means the record predates scope tracking.
+ * That is not a guess: `read write` is the only scope this server has ever
+ * requested, so any record written by any released version carries that grant —
+ * hence `true`, and nobody is pushed through a sign-in on upgrade. The
+ * `typeof` (rather than an `undefined` check) also absorbs a hand-edited
+ * `"scope": null`, which `loadToken` would otherwise pass through unvalidated.
+ *
+ * What this deliberately does NOT model is Zendesk's scope hierarchy: a
+ * granular grant such as `tickets:read` is judged not to cover a `read`
+ * request, and vice versa. Both misjudgements only ever cost one extra
+ * interactive sign-in, and neither is reachable while `read` and `read write`
+ * are the only scopes requested. Granular scopes (#284) must replace this
+ * predicate rather than extend it.
+ */
+export const grantCovers = (granted: string | undefined, requested: string): boolean => {
+  if (typeof granted !== 'string') return true;
+  const held = new Set(scopeTokens(granted));
+  return scopeTokens(requested).every((token) => held.has(token));
+};
--- a/src/auth/oauth-scopes.ts
+++ b/src/auth/oauth-scopes.ts
@@ -15,6 +15,15 @@
 const READ_SCOPE = 'read';
 const READ_WRITE_SCOPE = 'read write';
 
+// Hoisted: the predicate below runs on every token read.
+// Stryker disable next-line Regex: the empty-token filter below makes `/\s/`
+// and `/\s+/` produce identical output, so the quantifier is clarity, not
+// behaviour, and no test can tell the two apart.
+const WHITESPACE = /\s+/;
+
+const scopeTokens = (scope: string): string[] =>
+  scope.split(WHITESPACE).filter((s) => s.length > 0);
+
 /**
  * The scope to request for a given tool surface. `--read-only` already filters
  * every write tool out of the surface, so asking Zendesk for `write` on top of
@@ -30,16 +39,8 @@ export const requestedScope = (readOnly: boolean): string =>
  * metadata. Derived from `requestedScope` so what the HTTP transport advertises
  * provably cannot drift from what the stdio flow requests.
  */
-export const supportedScopes = (readOnly: boolean): string[] => requestedScope(readOnly).split(' ');
-
-// Hoisted: the predicate below runs on every token read.
-// Stryker disable next-line Regex: the empty-token filter below makes `/\s/`
-// and `/\s+/` produce identical output, so the quantifier is clarity, not
-// behaviour, and no test can tell the two apart.
-const WHITESPACE = /\s+/;
-
-const scopeTokens = (scope: string): string[] =>
-  scope.split(WHITESPACE).filter((s) => s.length > 0);
+export const supportedScopes = (readOnly: boolean): string[] =>
+  scopeTokens(requestedScope(readOnly));
 
 /**
  * Whether a token granted `granted` may still be used to serve a request for
--- a/src/auth/token-store.ts
+++ b/src/auth/token-store.ts
@@ -63,27 +63,29 @@ export const createTokenStore = (
   // Seed the in-memory cache from disk so a restart (notably the Cowork-on-Windows
   // process churn) reuses the existing token instead of re-prompting.
   let token: StoredToken | undefined = loadToken(tokenPath);
-  if (token && !grantCovers(token.scope, requested)) {
-    // The disk record was minted for a narrower surface (typically: this server
-    // dropped `--read-only` since). A refresh cannot widen a grant, so the only
-    // way out is a fresh authorization -- which the empty cache below triggers
-    // on the first tool call.
-    //
-    // Checked here and nowhere else, on purpose. A freshly minted token is the
-    // best this client can obtain: refusing it would open a browser window on
-    // every single tool call, forever, where the pre-#283 behaviour merely got
-    // a 403 on writes. A refreshed one cannot have widened either, and dropping
-    // it would rotate Zendesk's single-use refresh token for nothing. Both are
-    // served, and the next start re-decides from the record on disk.
-    //
-    // The drop is memory-only, never `clearPersistedToken`: one token file
-    // serves every process on this subdomain, so a record this server refuses
-    // may be exactly right for a read-only sibling -- and a successful sign-in
-    // overwrites the file regardless.
-    logger.warn('oauth_token_scope_insufficient', { requested, granted: token.scope });
-    token = undefined;
-  } else if (token) {
-    logger.debug('oauth_token_loaded_from_disk');
+  if (token) {
+    if (grantCovers(token.scope, requested)) {
+      logger.debug('oauth_token_loaded_from_disk');
+    } else {
+      // The disk record was minted for a narrower surface (typically: this server
+      // dropped `--read-only` since). A refresh cannot widen a grant, so the only
+      // way out is a fresh authorization -- which the empty cache below triggers
+      // on the first tool call.
+      //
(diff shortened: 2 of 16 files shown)


src/auth/oauth-scopes.ts
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * The same decision as a list, for the RFC 9728 / RFC 8414 `scopes_supported`
+ * metadata. Derived from `requestedScope` so what the HTTP transport advertises
+ * provably cannot drift from what the stdio flow requests.
+ */
+export const supportedScopes = (readOnly: boolean): string[] =>
+  scopeTokens(requestedScope(readOnly));
+
+/**
+ * Whether a token granted `granted` may still be used to serve a request for
```
dlecan: Nitpick: reduce to 4 lines max

src/auth/token-store.ts
```diff
@@ -48,14 +49,44 @@ const expiryFrom = (expiresIn: number | undefined): number | undefined =>
...
+  const requested = requestedScope(config.readOnly);
   // Seed the in-memory cache from disk so a restart (notably the Cowork-on-Windows
   // process churn) reuses the existing token instead of re-prompting.
   let token: StoredToken | undefined = loadToken(tokenPath);
-  if (token) logger.debug('oauth_token_loaded_from_disk');
+  if (token) {
+    if (grantCovers(token.scope, requested)) {
+      logger.debug('oauth_token_loaded_from_disk');
+    } else {
+      // The disk record was minted for a narrower surface (typically: this server
```
dlecan: Nitpick: reduce to 4 lines max

src/transports/http.ts
```diff
@@ -188,12 +189,17 @@ export const buildOAuthMetadata = (
   const { authorizeUrl, tokenUrl } = getOAuthUrls(config.subdomain);
   const issuer = `https://${config.subdomain}.zendesk.com`;
   const resource = resolveResourceUrl(config, logger);
```
dlecan, line 191: Nitpick: remove

dlecan:
## Functional validation report — d3a88bd36db1bbf39ce75b0d1d21cd97f77db40b

Run by: independent validator session (did not write the code)

`dist/index.js` was already newer than every file under `src/`, so V3-V7 ran against the current build without a rebuild. Values from the live tenant are replaced with placeholders below; only structure is reported.

| ID | Verdict | Evidence |
| -- | ------- | -------- |
| V1 | OK | Running authenticated server, no manipulation. `get_current_user` returned the user record with fields name, id, email, role, role type, active, organization (values redacted). No `Zendesk authentication required`, no browser prompt. Caveat below. |
| V2 | OK | `jq keys` returns `["accessToken","expiresAt","refreshToken","scope"]`; `.scope` is `"read write"`. No token material read or reported. |
| V3 | OK | Two `scope=` occurrences in the whole run, both `scope=read`: the ungated stderr line `If the browser doesn't open, visit: <authorize url>` and the `tools/call` error text. `grep -c scope` returns 2, so no other scope string appears anywhere in the run. |
| V4 | OK | Same probe without `--read-only`: both occurrences are `scope=read+write`. |
| V5 | OK | First stderr line: `[zendesk-mcp] [warn] oauth_token_scope_insufficient requested=read write granted=read`. `tools/call` returns `isError: true` with `Zendesk authentication required. A browser window should have opened...`, and the authorize URL in both the stderr fallback and the error text carries `scope=read+write`. |
| V6 | OK | Same narrow record with `--read-only`: `tools/call` returns `isError: true` with `Authentication failed. Your Zendesk token may be expired or invalid. Re-authenticate to get a new token.`, so the fake token was served and the tenant answered 401. `grep -c oauth_token_scope_insufficient` returns 0 and `grep -c oauth/authorizations/new` returns 0. The probe token file still has the same four keys after the run. |
| V7 | OK | Read-only server: `"scopes_supported":["read"]` in both `/.well-known/oauth-protected-resource` and `/.well-known/oauth-authorization-server`. Default server on the other port: `["read","write"]` in both. No credentials sent. Both processes killed afterwards, no listener left. |

Two notes on how the runs were set up, neither of which changes an observable.

The session had a live X display, so `open(authUrl)` would have opened real browser windows on the operator's desktop. V3-V7 therefore ran with `DISPLAY` unset and `BROWSER=/bin/true`. The authorize URL reaches stderr through the ungated `console.error` at `src/auth/browser-oauth.ts:293-294`, before `open()` is called, so it is printed either way. The probe files live in a session scratchpad rather than `/tmp`, and the real token cache kept its original mtime through all seven scenarios.

V6 first came back with no response to the `tools/call` at all: the initialize result, then nothing. That was the probe, not the code. Unlike V5, V6 reaches the network, and stdin closed at the end of the `printf` before Zendesk's answer arrived. Holding stdin open for 25 seconds produced the 401 text quoted above. The V5 and V6 pair is what shows where the grant is checked: the record that does not cover is refused at the seed and never sent, while the one that covers goes out and comes back 401.

V1 has one limit worth stating. The on-disk record already carries a `scope` key (V2), so this run cannot prove a pre-branch record without that key was read back. What it does show is that a `read write` grant on disk is accepted by a default-mode server with no re-prompt, which is the upgrade path for anyone whose record gets a `scope` key on first sign-in after the change. Reading a record with no `scope` key at all is covered by unit tests, not by this run.

## Summary

Green. All seven scenarios pass on d3a88bd. Nothing blocked, nothing failed, and the only gap is the V1 caveat above: the legacy record shape was not exercised end to end because the cache on this machine already carries a scope.

dlecan:
Thanks — the evidence is exactly what the plan asked for, and the V5/V6 pair reads as the sharpest part of it: the record that does not cover is refused at the seed and never leaves the process, the one that covers goes out and comes back 401. The stdin note on V6 is a real gap in my plan's probe, not in the code; anyone re-running it should keep stdin open past the round-trip.

Two things before I call this closed.

**The SHA does not resolve here.** `d3a88bd36db1bbf39ce75b0d1d21cd97f77db40b` is not in the repository — not on `origin/claude/issue-283-analysis-mbrob7` (head is `08faeb2`), not on `main`, and not reachable locally after a fetch. The report is very probably against the right tree: it cites `src/auth/browser-oauth.ts:293-294` for the ungated `console.error`, and those are exactly the two lines at that offset in `08faeb2` (the offset only became 293-294 after the review nits shortened three comments). So this looks like a local commit carrying the same tree rather than a stale run. Could you confirm with `git diff d3a88bd 08faeb2 --stat` — empty output settles it, and nothing needs re-running.

**The V1 caveat is worth one more scenario, not a re-run.** You are right that a record already carrying `scope` cannot prove the legacy shape is read back. That is testable without waiting for anyone's cache to age:

```
jq 'del(.scope)' "$ZENDESK_TOKEN_FILE_OR_DEFAULT" > /tmp/probe-legacy.json
# then a tools/call against a default-mode server with ZENDESK_TOKEN_FILE=/tmp/probe-legacy.json
```

Expected: the call succeeds, no `oauth_token_scope_insufficient`, no authorize URL — the absent `scope` is read as the `read write` grant it was minted under. That is the upgrade path for every existing user, and it is the one line of the change I would least like to be wrong about. Unit tests cover the predicate; this would cover the real record.

---

dlecan:
## V1 addendum, legacy record shape — d3a88bd36db1bbf39ce75b0d1d21cd97f77db40b

Run by: independent validator session (did not write the code)

The first report left V1 with a caveat: the cache on this machine already carried a `scope` key, so a record written by a released version, with no `scope` key at all, was never actually read back. That gap is now closed, on the same commit.

`loadToken` runs once, at `createTokenStore` (`src/auth/token-store.ts:65`), and `getToken` never reads the disk again, so the already-running server could show nothing about the seed. Instead, a copy of the real record with `del(.scope)` applied was installed at the default token path, and two fresh stdio processes were started against it, one in default mode and one with `--read-only`. No `ZENDESK_TOKEN_FILE` override, so this also exercises the normal path resolution. The access token had 16 minutes left against a 1 minute refresh skew, so `needsRefresh` was false and no refresh could spend the shared refresh token.

| Observation | Default mode | `--read-only` |
| --- | --- | --- |
| Seed | `oauth_token_loaded_from_disk` at debug, once | same |
| `oauth_token_scope_insufficient` | absent | absent |
| Authorize URL emitted | none | none |
| Refresh events | none | none |
| `tools/call get_current_user` | no `isError`, user record returned with fields Email, Role, Role type, Active, Organization (values redacted) | same |
| Tools registered | `count=53 mode=all` | `count=36 mode=all` |
| Token file after the run | byte identical | byte identical |

A record with no `scope` key is therefore served in default mode with no prompt, and it also covers the narrower `read` request under `--read-only`. The two runs differ only in the registered tool count, which confirms they were genuinely separate processes in different modes.

Two hygiene points. Debug-level logs carried no token material: the real access and refresh token strings were matched against the captured output and returned zero hits. The original record was moved aside and moved back, and its sha256, its size, its mtime and its `0600` mode are what they were before the run.

V1 is OK without the caveat. Everything else in the first report stands.

dlecan:
That closes it. The addendum is the run I wanted: a record with no `scope` key, at the default path, read back by two fresh processes — served in both modes, no prompt, no refresh, file byte identical afterwards. The tool counts differing (53 / 36) is a nice touch: it proves the two runs were genuinely separate processes rather than one repeated.

On the SHA I raised: I'll treat it as settled by the evidence rather than hold the PR for the `git diff`. The addendum cites `src/auth/token-store.ts:65` for `loadToken`, and that is exactly where it sits in `08faeb2` — a second independent line-offset match, in a different file from the first, and both offsets only became those numbers in the last commit. A tree that differed would have to coincide twice.

Validation is green end to end: V1 (now including the legacy record shape), V2, V3 `scope=read`, V4 `scope=read+write`, V5 refused at the seed with `oauth_token_scope_insufficient` and never sent, V6 served and answered 401 by the tenant, V7 `scopes_supported` narrowed in both discovery documents. CI is green on `08faeb2`, the branch is up to date with `main`, and every review thread is resolved.

Nothing further from me — the merge call is the maintainer's.

---

dlecan:
Correction to my previous comment: the SHA does resolve. `d3a88bd` is `chore(build): improve WT config`, pushed on top of `08faeb2`, and my fetch simply predated it. So there was never any ambiguity about which tree was validated — the report ran against the branch head, and the line-offset reasoning I used was unnecessary.

It changes nothing about the verdict, and nothing needs re-running: `d3a88bd` touches `.config/wt.toml` only, with an empty diff against `08faeb2` across `src/`, `tests/`, `docs/` and `README.md`. CI is re-running on the new head.

---

dlecan:
On the outside-diff finding at `src/auth/token-store.ts:147` / `:188` — the observation is correct, the remedy is not, and this is deliberate rather than an oversight.

The observation: yes, a token minted or refreshed with a grant narrower than requested is cached and served, and the seed check does not cover those paths.

The remedy proposed — validate with `grantCovers` and restart authorization — is the first version of this branch, and it was removed in `7af01e5` after review because it is an unbounded loop. When Zendesk grants less than was asked for, a **new authorization returns exactly the same narrow grant**: nothing about re-running the flow changes the client's allowed scopes. So every tool call would open another browser window, forever, where the behaviour before this PR was merely a 403 on writes. Refusing a *refreshed* token is the same trap with an added cost: a refresh cannot widen a grant either (RFC 6749 §6, and our refresh request sends no `scope` at all), so the rejection spends Zendesk's single-use refresh token and leaves the shared record narrower than it found it.

The disk seed is the one moment a grant can actually be stale — the server dropped `--read-only` since the record was written — and there a fresh authorization genuinely fixes it. That asymmetry is why the check sits there and nowhere else; it is stated in the code comment at the seed and in the PR description.

Worth adding: on Zendesk this input does not arise. A scope outside the OAuth client's allowed scopes fails the authorization with `400 invalid_scope` and mints no token at all — that is the #283 symptom itself. A downgraded grant is an RFC-permitted behaviour that this provider does not exhibit.

What the finding does point at fairly is diagnosis: if a narrow grant ever were served in write mode, the operator would see 403s on writes with nothing naming the cause. A one-line `warn` at mint time, serving the token but recording that the grant fell short, would cover that without reintroducing the loop. Happy to add it if you want it, but I am not widening the PR for a case Zendesk does not produce.

---

dlecan:
Added in `dcc7a1d`: the diagnosis half of that finding, without the loop.

Both installation paths now go through one `noteGrant`, which warns `oauth_token_grant_narrowed` with the requested and granted scopes when a minted or refreshed grant falls short. The token is still served, for the reason above — a new authorization returns the same narrow grant — so the seed stays the only place a shortfall sends anyone through a sign-in. What changes is that the operator gets a line naming the cause instead of silent 403s on writes.

No behaviour change, so the functional validation from `d3a88bd` still stands: none of V1-V7 reaches this path (none of them mints a token). Gates on the new head: lint, typecheck, 1064 tests, 46/46 mutants in the changed lines, smoke.

---

PR #298, opened 2026-09-18
fix(mutation): patch the vitest 5 incompatibility and add a canary that notices when the gate stops measuring

### LINKED ISSUE #297: vitest 5 silently disables the mutation gate — and nothing in CI noticed
Found while merging `main` into #295. Two problems, and the second is the bigger one.

## 1. The gate no longer works

Since #292 bumped vitest to v5, `@stryker-mutator/vitest-runner` stops observing test results. Mutants the test suite genuinely kills are reported as **survived**, so `pnpm test:mutation:diff` fails the author for assertions that are in fact doing their job.

Same tree, same 44 mutants, nothing changed but the runner version:

| vitest | result |
|--------|--------|
| 4.1.11 | `Killed: 44. Escaped: 0.` |
| 5.0.0  | `Killed: 7. Escaped: 37.` |

Direct evidence that the survivors are phantom — one Stryker reports as survived, `src/utils/formatting.ts:167`:

```diff
-  Math.round((t - now) / 60_000)
+  Math.round((t - now) * 60_000)
```

Applied by hand, `pnpm vitest run tests/unit/utils/formatting.test.ts` fails **11 snapshots**. Stryker calls it survived on the same tree.

`@stryker-mutator/vitest-runner@10.0.0` is the latest release and declares `vitest: ">=2.0.0"`, so vitest 5 satisfies the peer range and installs cleanly. Nothing warns. No upstream release supports vitest 5 yet.

Consequence today: every PR touching `stryker.config.mjs` scope on top of `main` gets a false red, and the only way to read it is to know this issue exists.

## 2. The part that should worry us more: nothing detected it

This is a regression **of the tests**, and the test suite has nothing to say about it. `pnpm test` is green on vitest 5 — 1199 passing. Coverage is above thresholds. #292's own CI was green, because the gate is diff-scoped and a dependency-only change touches no mutated source line, so there was nothing for it to get wrong. A tool whose entire job is to prove the assertions are load-bearing was switched off, and every signal we have kept saying "fine".

The gate has no way to distinguish "no mutant escaped because the tests are strong" from "no mutant was ever really tested". Those two states produce the same green, and the first is the one we believe.

The missing piece is a **canary**: a known-good mutant that the suite provably kills, asserted to be reported as killed. If the runner stops observing results, the canary flips and names the cause instead of blaming the next author's diff. `tests/unit/mutation-scope.test.ts` already pins the scope matcher against `stryker.config.mjs`, so the habit of testing this tooling exists — it just stops short of checking that the tool still runs.

Worth deciding alongside it: whether a dependency bump to a **test-infrastructure** package should be allowed to merge on a diff-scoped gate that, by construction, cannot exercise it. A deps-only PR touching `vitest`, `@stryker-mutator/*` or `@vitest/coverage-v8` is exactly the case where the full-scope run (or the canary above) is the only meaningful signal.

## Options for the immediate breakage

1. **Pin back** — `vitest` and `@vitest/coverage-v8` to `4.1.11`, held in Renovate until `@stryker-mutator/vitest-runner` supports 5. Restores the gate now; costs the v5 upgrade.
2. **Stay on 5** and accept a non-functioning gate until upstream catches up, with the failure documented so authors do not chase phantom survivors.

Option 1 restores the guarantee the repo actually relies on; option 2 keeps an upgrade whose benefit is smaller than what it silently disabled.

## Notes for whoever picks this up

- Reproduce with any branch touching mutated scope: run `pnpm test:mutation:diff origin/main HEAD` on vitest 5, then re-run it with `vitest@4.1.11` installed and compare the killed/escaped counts. The A/B above took two minutes.
- #295 carries the diagnosis in a comment and is red on `Changed lines` for this reason alone; its other gates (`typecheck`, `check`, `test`, `coverage`) are green.
- Whatever the outcome, the canary in §2 is the part worth keeping — pinning fixes today's break, but nothing yet prevents the next silent one.

dlecan:
## Summary

Carries [stryker-js#6214](https://github.com/stryker-mutator/stryker-js/pull/6214) as a pnpm patch so the mutation gate works **on vitest 5**, and adds `pnpm test:mutation:canary` — a ~3 s check, wired into both `mutation.yml` jobs, that fails when StrykerJS stops observing test results instead of leaving the next author to chase phantom survivors.

## Linked issue

Closes #297

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Documentation
- [x] Tooling / CI

## Root cause

A **test-filter** failure, not a result-reporting one — which is why it produced survivors rather than errors.

`VitestTestRunner.run()` narrows a mutant's run to its covering tests by writing a regex into `project.config.testNamePattern`, built from the runner's own `collectTestName()`: suite names and the test name joined with a **single space**. Vitest 4 matched that against `getTaskFullName(task)`, built the same way. Vitest 5 matches it against the new precomputed `task.fullTestName`, which `createTaskName()` joins with **`" > "`** (`interpretTaskModes`, `vitest/dist/task-utils.js`). Nothing matches, every test in the file is set to `skip`, Stryker observes zero results, and a mutant no test ran is reported `Survived`.

Confirmed against the sandbox of a real run (`src/utils/formatting.ts:167`, `ArithmeticOperator`), driving the vitest 5 node API exactly as the runner does:

| `testNamePattern` | tests observed | file state | mutant |
| --- | ---: | --- | --- |
| none | 12 → 1 failed | `fail` | detected |
| `formatSlaBlock states the exact number of minutes remaining` (what the runner builds) | **0** | `skip` | **undetected** |
| `states the exact number of minutes remaining` | 1 → failed | `fail` | detected |
| `formatSlaBlock > states the exact number of minutes remaining` | 1 → failed | `fail` | detected |

That accounts for the shape of the numbers in the issue, not just their direction: a mutant planned with **no** test filter (static, or one with no coverage information) still runs the whole suite and is killed normally — the `Killed: 7` — while every mutant that gets a per-test filter escapes. Mutant activation, `provide`/`inject`, the setup-file injection and the `suite.meta` coverage round-trip are all intact on vitest 5; only the filter is broken.

## The fix: a patch, not a rollback

`patches/@stryker-mutator__vitest-runner@10.0.0.patch` is #6214 transposed onto the runner's published `dist/src/` — the same four files as its `src/` (`test-helpers`, `stryker-setup`, `vitest-helpers`, `vitest-test-runner`), same shape, so re-reading it against the PR is a diff rather than an exercise. The separator becomes a value chosen once from `vitestWrapper.version` (`' > '` for ≥ 5, `' '` below), and it is `provide`d to the setup file because `ns.currentTestId` — the key coverage is recorded under — must be built with the same separator the filter regex is built from. Fixing only the filter would move the mismatch rather than remove it.

Three things make it safe to carry:

- **It cannot go stale silently.** `patchedDependencies` pins the exact version (`@stryker-mutator/vitest-runner@10.0.0`), so a runner bump leaves the patch unapplied rather than applying it to code it was not written for — and the canary is red on that PR before anyone reads it.
- **It cannot go stale invisibly in the baseline either.** `pnpm-lock.yaml` records the patch's content hash, and that file is already part of the mutation baseline cache key, so editing the patch discards the baseline exactly as a dependency bump does. By construction, not by anyone remembering.
- **It is verified, not assumed.** On vitest 5 with the patch: `src/utils/pagination.ts` is **43 killed / 0 escaped** (the 100 % the ADR already records for that file), and the four lines of `formatting.ts` from the reproduction are **3 / 0** where unpatched vitest 5 gives **0 / 3**.

Drop the patch when a runner release carries #6214 — `renovate.json` keeps `@stryker-mutator/*` off automerge so that decision reaches a person, and the canary says whether the release really fixed it.

## The canary

Four design points, each load-bearing:

- **Two-sided.** `killedByItsTest` must come back `Killed`, `reachedByNoTest` must come back `NoCoverage`. A one-sided "it was killed" check would pass against a runner that reported everything killed, which says just as little.
- **The fixture's test is nested two `describe`s deep.** The filter regex joins suite names to the test name, so a top-level `it` exercises a filter with no separator in it — the one shape that kept working through this incident. A flat fixture would have stayed green throughout.
- **Exhaustive in both directions.** A missing mutant, an unexpected one and a wrong verdict are each a finding: a report with *no* mutants is precisely what a collapsed run produces, and "nothing escaped" would wave it through. `judgeCanary` is pinned in `tests/unit/mutation-canary.test.ts`.
- **Its own vitest project and report path.** No setup file, no MSW, no coverage thresholds — nothing that gives a canary failure a second possible cause — and nothing that writes over the gate's report or its incremental baseline. The report is deleted before each run and its path read from the canary config, so the canary can never pass on a file it did not write.

It runs in **both** `mutation.yml` jobs, before the gate and before the baseline, and unconditionally. That is the answer to the second half of the issue: a dependency-only PR changes no mutated line, so the gate has nothing to judge and by construction cannot exercise the runner. The canary is the only signal such a PR gets, and it costs 2.4 s.

Verdicts measured: passes on 4.1.11, passes on patched 5.0.0, and on **unpatched** 5.0.0 with `pnpm test` green it fails in 2 seconds with ```a - b`: expected Killed, reported Survived.``

## Upstream

**No new issue was filed: [stryker-js#6210](https://github.com/stryker-mutator/stryker-js/issues/6210) already names the same cause** (opened 2026-09-04), with [#6214](https://github.com/stryker-mutator/stryker-js/pull/6214) as the fix — open and unmerged, which is why it is carried here as a patch.

`docs/decisions/stryker-vitest5-upstream-report.md` keeps our reproduction: it is the evidence the reading is independent (upstream's numbers are the same failure at larger scale — 47.36 → 2.96, 1398 of 1651 survivors with `testsCompleted: 0` against a populated `coveredBy`), and it is what the next person needs when judging whether a release makes the patch droppable. One point it raises that #6214 does not cover: tightening the runner's peer range from `vitest: ">=2.0.0"`, which is what turned a compatibility gap into a silent one. Say the word if that is worth posting on #6210.

## Author checklist

- [x] `pnpm test` passes locally — 45 files, 1202 tests, on vitest 5.0.0
- [x] `pnpm test:coverage` meets the thresholds — 98.22 % statements, 89.57 % branches, 99.47 % functions, 98.62 % lines
- [x] `pnpm typecheck` passes
- [x] I ran a Claude Code review on the diff and addressed its findings — both are in `f5f8b5e`: the canary read a hardcoded report path it never cleared (a stale report could print "passed"), and `replacement in expected` matched inherited property names, leaving such a mutant unjudged. Fixed and pinned by tests; the stale-report case verified by planting a passing report and breaking the run, which now exits 1.
- [x] Documentation is updated where needed — ADR section 9, the upstream report, `AGENTS.md`, `CONTRIBUTING.md`
- [x] If this PR adds an MCP tool: it is documented in `docs/mcp-tools-reference.md` — n/a
- [x] **MCP-runtime-observable behaviour**: none. Dev tooling, CI, a devDependency patch; a running server behaves identically for a client, so the standard gates are the validation and no third-party pass is needed (`AGENTS.md`, Planning).

- The branch keeps the rollback commit that preceded the patch (`bf1e580`, pin to 4.x) rather than rewriting it away: it is where the A/B measurement lives, and the patch commit that follows is what supersedes it. Net effect on `main` is vitest 5 plus the patch.
- `scripts/mutation-canary/` holds the fixture rather than `tests/`: its test must **not** be discovered by the repo's vitest project (`tests/**/*.test.ts`), or the canary would nest a Stryker run inside the suite Stryker is running. Keeping it out of `tests/` also leaves the baseline cache key's "every non-`*.test.ts` file under `tests/`" invariant (`tests/unit/mutation-scope.test.ts`) untouched, with no second exemption to reason about.
- The canary config extends the repo's `stryker.config.mjs` rather than restating it, so the TypeScript 7 `tsconfigFile` sentinel and the explicit `plugins` entry cannot drift. It overrides `incremental: false` explicitly although that is the default — the inherited `incrementalFile` is the PR gate's baseline, and a canary verdict must never be written into it.
- `ArrowFunction` is excluded in the canary config only: it empties both one-line arrows to the same `() => undefined` text, so the two mutants become indistinguishable by replacement — and one of them is static, which Stryker runs without a test filter, i.e. on the one path this failure left working.
- To see the failure again with the patch in place: drop the `patchedDependencies` entry in `pnpm-workspace.yaml`, re-install, run the canary.

https://claude.ai/code/session_0187QMhPuwpWXgZfKq6R7ANF

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed mutation testing compatibility with Vitest 5, ensuring nested test filters produce accurate results.
  * Added safeguards to detect missing, unexpected, or incorrect mutation results.

* **Testing**
  * Added a mutation-testing canary that runs in pull-request and baseline checks.
  * Added coverage for mutation verdict validation and Vitest integration scenarios.

* **Documentation**
  * Documented the Vitest 5 compatibility issue, canary usage, and mutation gate behavior.


scripts/mutation-canary.mjs
```diff
@@ -0,0 +1,167 @@
...
+// doing their job. The canary turns that into a failure that names its own
+// cause, on the PR that introduces it.
+//
+// Two mutants, two different expected verdicts. A one-sided "it was killed"
+// check would pass against a runner that reported everything killed, which
+// tells us just as little — so the uncovered half must come back `NoCoverage`.
+// Keying on the replacement text rather than a line number keeps the fixture
+// editable.
+
+import { execFileSync } from 'node:child_process';
```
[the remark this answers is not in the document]
dlecan: Correct, and fixed in `2dc16e2` — all four sites plus the canary's Stryker config, which had the same problem and was not flagged.

Each header now keeps only the non-obvious constraint and points at `docs/decisions/mutation-testing.md` (§9), which is where the #297 narrative belongs and now lives in full: why the fixture's test is nested two `describe`s deep, why `ArrowFunction` is excluded, why the config extends the repo's rather than restating it.

---

AGENTS.md
```diff
@@ -123,6 +123,14 @@ quality bar still hold — the freedom is from Zendesk's shape, not from craft.)
...
   `// Stryker disable next-line <mutator>: <reason>`. Rationale, scope and costs:
   `docs/decisions/mutation-testing.md`.
+- That gate cannot tell a strong suite from a mutation runner that stopped
+  observing results, so `pnpm test:mutation:canary` (~3 s, runs in both
+  `mutation.yml` jobs) asserts a fixed verdict on a fixture. A canary failure is
+  never about the code under review — read it as the toolchain, starting with
+  the vitest and StrykerJS versions. `@stryker-mutator/vitest-runner` carries
+  stryker-js#6214 as a pnpm patch for the same reason (`patches/`, pinned to an
+  exact version in `pnpm-workspace.yaml`): drop it when a runner release ships
+  that fix, with the canary — not the release notes — as the check. §9 of the ADR.
```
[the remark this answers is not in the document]
dlecan: Right again — fixed in `2c0926a`. The bullet is down to what changes an agent's next step: the command, how to read its failure (toolchain, not the code under review), and that a patch exists with the condition for dropping it. Mechanics, the patch's location and the upstream history stay in §9.

---

AGENTS.md
```diff
@@ -123,6 +123,11 @@ quality bar still hold — the freedom is from Zendesk's shape, not from craft.)
   for a genuinely equivalent mutant say why with
   `// Stryker disable next-line <mutator>: <reason>`. Rationale, scope and costs:
   `docs/decisions/mutation-testing.md`.
+- `pnpm test:mutation:canary` (~3 s) asserts that gate still observes results at
```
dlecan: Suggestion: let the CI does its job, so remove this doc

CONTRIBUTING.md
```diff
@@ -215,7 +215,13 @@ assumption that everything below has already been done.
    `pnpm build`. A green CI on a non-green local run means a flaky check, not a
    free pass. If your PR changes anything under `src/`, also run the mutation
    gate on your diff, because CI does and it fails on a surviving mutant in a
-   line you changed: `pnpm test:mutation:diff origin/main HEAD`. Background:
+   line you changed: `pnpm test:mutation:diff origin/main HEAD`. If your PR
```
dlecan: Suggestion: same as agents.md

AGENTS.md
dlecan: Done in `52c75f9` — the bullet is gone, `AGENTS.md` is back to its previous text. The canary runs unconditionally in both `mutation.yml` jobs, so nothing was telling an agent anything CI would not have told it.

The one instruction that is not CI's to give — delete the pnpm patch when a runner release carries stryker-js#6214 — stays where it acts: the Renovate rule keeping `@stryker-mutator/*` off automerge, so the update lands in front of a person, with §9 of the ADR behind it.

---

CONTRIBUTING.md
dlecan: Same, in `52c75f9` — step 5 is back to its previous text.

One line of mine is left nearby, in step 1 of "What happens after you open the PR": that the canary step does not skip itself when the diff touches nothing the gate mutates. It describes what CI does rather than asking anyone to run anything, so I kept it — say the word and it goes too.

---

